PR #95, opened 2018-05-01
Build docker images in Circle CI

sargun:
* Use the build docker images from Circle CI builds
* Add code coverage metrics
* Try to collect all of the image builds in one place


sargun:
This again requires the "inside" and "outside" GO PATHs to match, but things break under go 1.10 caching (if we share it), and we have a different "inside" and "outside" set of paths.

hack/test-images/ubuntu-systemd-bionic/Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM ubuntu:bionic
```
sargun: This was being built by hand, and uploaded by hand before.

hack/test-images/ubuntu-systemd-xenial/Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM ubuntu:xenial
```
sargun: This was being built by hand, and uploaded by hand before.

hack/test-images/build.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
```
sargun: This is a combined way to build all the build images.

hack/ci-builder/Dockerfile
```diff
@@ -44,4 +44,5 @@ ENV GOPATH /go
 ENV PATH $GOPATH/bin:/usr/local/go/bin:$PATH
```
sargun: goveralls is to push the reports

executor/mock/standalone/standalone_test.go
```diff
@@ -43,14 +35,16 @@ type testImage struct {
 }
 
 var (
+	// TODO: Determine how this got built, and add it to the auto image builders?
```
sargun: These are all bumped to automated build versions.

executor/mock/standalone/standalone.go
```diff
@@ -1 +1,26 @@
 package standalone
+
```
sargun: :(

.buildkite/pipeline.yml
```diff
@@ -1,27 +1,6 @@
...
-              pushd $$GOPATH/src/github.com/Netflix/titus-executor
-
-              go get -u github.com/alecthomas/gometalinter
-              gometalinter --install
-              go get -u github.com/kardianos/govendor
-
-              make clean
-              make --output-sync -j16 builder all 2>&1 | tee build.log
-    artifact_paths: "*.log;build/distributions/*.deb;build/distributions/*.tar;build/distributions/*.tar.gz"
+  - command: ./.buildkite/build.sh
```
sargun: This got moved into a little script

fabiokung:
LGTM

@sargun I'm still unsure and :-1: on bind mounting `GOPATHs`, and why we need to share the same cache inside and outside the build containers. I'd rather they have separate caches, they may even be different OSes and arch inside and outside. The build containers can use volumes from data-only container to cache things across builds if it needs to be sped up.

sargun:
Getting rid of the data only containers, and sharing caches with the host builds sped things up significantly. Initially doing the test-local, which built the build cache, and then mounting the GOPATH made build times cut down by almost 50%.

sargun:
If the GOPATH wasn't the same inside and outside of the container, then it wouldn't use the GOPATH. Go seems to freak out whenever it sees bind mounts, or symlinks in gopath. Not sure why.

PR #96, opened 2018-05-01
Remove all mentions of junit

sargun:


sargun:
The existence / "need" of the junit test format is from before my time. Not sure why we have it?

PR #97, opened 2018-05-02
Update build / make instructions

sargun:


README.md
```diff
@@ -39,18 +68,38 @@ Tests will run inside a Docker container and run a dedicated docker daemon as do
...
 
 ```sh-session
-make clean-proto-defs && make protogen
+make -C vpc/bpf/
 ```
 
+### api/netflix/titus
+If you update the dependency `github.com/Netflix/titus-api-definitions/src/main/proto/netflix/titus`, you will need to regenerate the Go proto definitions. 
+
+You must first install the protobuf toolchain, and the Go protobuf compiler. More documentation on that can be found [in the protobuf repo](https://github.com/golang/protobuf/blob/master/README.md#installation).
```
fabiokung: different versions of the `protoc` compiler will generate different code, and some can potentially fail. This was the main reason the `protoc` stuff was containerized and pinned to a particular version.

sargun: Yeah, since the generated code is checked in (by us). We should be able to verify it.

PR #98, opened 2018-05-02
Remove dependency on titus-bootstrap

sargun:


sargun:
For example, the contents of our /etc/titus-executor/config.sh:
would be:
export TITUS_REGISTRY=OBJUSCATED.main.${EC2_REGION}.dyn${NETFLIX_ENVIRONMENT}.netflix.net:1234
export TITUS_LOG_BUCKET=${EC2_REGION}.netflix.s3.OBJUSCATED.${NETFLIX_ENVIRONMENT}
export USE_NEW_NETWORK_DRIVER=true

PR #91, opened 2018-04-21
Nested containers [WIP]

sargun:
This bumps the builders to artful, so they have access to the
right headers in order to get access to cgroup unshare.

It also bumps tini to pull in this version:
Netflix-Skunkworks/tini#2

There are a few things that have changed:
 - A new seccomp profile for nested containers
 - A new apparmor profile for nested containers

This PR still has a few TODOs:
 Lock down the following syscalls:
  * Clone: No making new namespaces
  * Mount: Only allow tmpfs
  * Unshare: Only allow cgroups
 Figure out how to better secure systemd
  * Custom LSM that doesn't allow it to give
    away the machine

PR #99, opened 2018-05-07
Dynamically generate JOB IDs for standalone tests

sargun:


sargun:
Adding @fabiokung for visibility. This is breaking out a change from #91

PR #102, opened 2018-05-07
darion: Use http.ServeContent for serving log data

cHYzZQo:
This allows us to support range serving ranges.

I'm not sure I fully understood the way the virtual file mapping is to supposed to work, so while I believe the behavior will be the same as the old code, that part might need some extra scrutiny.

### DIFF
--- a/darion/api/handlers_test.go
+++ b/darion/api/handlers_test.go
@@ -12,6 +12,7 @@ import (
 	"github.com/Netflix/titus-executor/filesystems"
 	"github.com/Netflix/titus-executor/filesystems/xattr"
 	log "github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/assert"
 )
 
 func init() {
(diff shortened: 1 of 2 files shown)


andrew-leung:
FYI, this is for TITUS-1922 (which I realize doesn't help anyone external to Netflix).

andrew-leung:
@cHYzZQo, looks like the CI tests are now passing. Ready for review?

sargun:
@cHYzZQo Just curious -- do you need this both on stdio and normal log files, or just on normal log files?

cHYzZQo:
@sargun yes ready for review. Both stdio and normal log files would be good.

sargun:
Okay, let me mull this over and I'll look at it EOD.

darion/api/handlers_test.go
```diff
@@ -184,22 +231,36 @@ func verifyStdout(resp *http.Response, t *testing.T) {
...
 	if err != nil {
 		t.Fatal("Unexpected error")
 	}
+
+	if rangeHeader != "" {
+		req.Header.Add("Range", rangeHeader)
+	}
+
+	resp, e := client.Do(req)
+	if e != nil {
```
sargun: Please use https://godoc.org/github.com/stretchr/testify/require#NoError

darion/api/handlers_test.go
```diff
@@ -168,7 +215,7 @@ func verifyHelper(resp *http.Response, t *testing.T) string {
 		t.Fatal("Could not read response body: ", err)
 	}
 	dataStr := string(data)
-	if resp.StatusCode != 200 {
```
sargun: When will we receive not 200? When it's range request not satisfiable?

sargun: If that's the case, we should probably explicitly check for that.

cHYzZQo: If a range is requested http specs say the server should respond with a  206 "Partial Content".

In general though, any code between 200 and 399 is considered a success. And success checks should allow for anything in that range. For example 304s would be pretty common. But I can change the test to only allow 200 or 206.

cHYzZQo: BTW the code for unsatisfiable range requests is 416.

sargun: Do we want add a specific test checking for the 416 case, and for the 206 case?

cHYzZQo: Sure, added. Though at that point we are testing the http lib not our code.

PR #109, opened 2018-05-09
Change the executor state machine

sargun:


sargun:
btw -- coveralls ONLY looks at local tests, whereas codecov merges standalone + local tests.

executor/runner/runner.go
```diff
@@ -349,6 +352,26 @@ no_launchguard:
 	}
 }
 
+func (r *Runner) handleTaskRunningMessage(ctx context.Context, msg string, lastMessage *string, runningSent *bool, startTime time.Time, details *runtimeTypes.Details) {
```
sargun: This exists mostly to keep the cyclomatic complexity of the monitorContainer function low.

executor/runner/runner.go
```diff
@@ -301,43 +301,46 @@ no_launchguard:
...
-			status, err := r.runtime.Status(r.container)
-			if err != nil {
-				r.logger.Error("Status result error: ", err)
-			}
-			shouldQuit, titusTaskStatus, msg := parseStatus(status, err)
-			if shouldQuit {
-				r.logger.Info("Status: ", titusTaskStatus.String())
-				// TODO: Generate Update
-				r.updateStatus(ctx, titusTaskStatus, msg)
+		case statusMessage := <-statusChan:
```
fabiokung: this needs to handle `statusChan` being closed too, no? It seems it will be closed when we can't talk to the docker daemon anymore, in which case this will cause a `panic: nil pointer`. At least failing more gracefully and sending the `StatusUpdate.Lost` message may be better.

executor/runtime/docker/docker.go
```diff
@@ -1124,62 +1131,188 @@ func (r *DockerRuntime) processEFSMounts(c *runtimeTypes.Container) ([]efsMountI
...
 	// This means that the log watcher is not started
-	return "", nil
+	return "", details, statusMessageChan, nil
+}
+
+func (r *DockerRuntime) statusMonitor(cancel context.CancelFunc, c *runtimeTypes.Container, eventChan <-chan events.Message, errChan <-chan error, statusMessageChan chan runtimeTypes.StatusMessage) {
+	defer close(statusMessageChan)
+	defer cancel()
+
+	// This context should be tied to the lifetime of the container -- it wille get significantly less broken
```
fabiokung: minor: typo in `it wille`?

executor/runtime/docker/docker.go
```diff
@@ -1124,62 +1131,188 @@ func (r *DockerRuntime) processEFSMounts(c *runtimeTypes.Container) ([]efsMountI
...
+			}
+		} else {
+			statusMessageChan <- runtimeTypes.StatusMessage{
+				Status: runtimeTypes.StatusFailed,
+				Msg:    fmt.Sprintf("exited with code %s", exitCode),
+			}
+		}
+	case "health_status":
+		statusMessageChan <- runtimeTypes.StatusMessage{
+			Status: runtimeTypes.StatusRunning,
```
fabiokung: add a `Msg` here indicating the change in `health_status`?

executor/runtime/docker/docker.go
sargun: 👍

executor/runner/runner.go
sargun: :+1:

executor/runner/runner.go
```diff
@@ -301,43 +301,50 @@ no_launchguard:
...
 
-	// TODO(fabio): Start should return Details
-	details, err := r.runtime.Details(r.container)
-	if err != nil {
-		r.logger.Error("Error fetching details for task: ", err)
-		r.updateStatus(ctx, titusdriver.Lost, err.Error())
-		return
-	} else if details == nil {
-		r.logger.Error("Unable to fetch task details")
+	if details == nil {
```
andrew-leung: Worth checking this before err != nil above? Otherwise, we just updated a bunch of status with nil details.

sargun: Nah -- because details might be nil, for example, if we weren't able to start the image, because of a bad entrypoint. This here is more a sanity-check, and should never happen.

sargun: -- In those cases, the error will get reported back to the user, and the program wont shut down entirely.

executor/runtime/docker/docker.go
```diff
@@ -1124,62 +1131,189 @@ func (r *DockerRuntime) processEFSMounts(c *runtimeTypes.Container) ([]efsMountI
...
 		r.metrics.Counter("titus.executor.dockerStartContainerError", 1, nil)
 		// Check if bad entry point and return specific error
-		return "", maybeConvertIntoBadEntryPointError(err)
+		eventCancel()
+		return "", nil, statusMessageChan, maybeConvertIntoBadEntryPointError(err)
 	}
 
 	r.metrics.Timer("titus.executor.dockerStartTime", time.Since(dockerStartStartTime), c.ImageTagForMetrics())
 
+	if c.Allocation.IPV4Address == "" {
```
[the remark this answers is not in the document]
sargun: When you're running this in "old network driver mode" those are unset.

PR #122, opened 2018-05-15
Generate metatron credentials with IP address

sargun:
This makes the generation of metatron credentials a callback, versus
the previous, which was a generation "ahead of time". Although,
this introduces a higher chance of error, and latency at container
startup, it lets us pass the IP address of the container into
the metatron init script.


sargun:
This relates to TITUS-1757

PR #125, opened 2018-05-17
Fix Docker OOM Default Settings and add better testing around termination timeouts

sargun:


sargun:
This will become dynamically adjustable once: https://github.com/Netflix/titus-api-definitions/pull/44 is merged.

PR #127, opened 2018-05-18
Import executor/runtime/docker so it gets code coverage

sargun:


sargun:
@andrew-leung @fabiokung for posterity.

PR #132, opened 2018-05-24
Remove Launchguard

sargun:


executor/runtime/docker/docker.go
```diff
@@ -661,14 +671,29 @@ func prepareNetworkDriver(cfg Config, c *runtimeTypes.Container) error {
 		"--batch-size", strconv.Itoa(cfg.batchSize),
 	}
 
+	// This blocks, and ignores kills.
+	if !c.TitusInfo.GetIgnoreLaunchGuard() {
```
andrew-leung: If IgnoreLaunchGuard is true, why would we want to skip this step?

executor/runtime/docker/docker.go
```diff
@@ -1595,33 +1637,50 @@ func setupNetworking(burst bool, c *runtimeTypes.Container, cred ucred) error {
...
+	go func() {
+		defer close(c.SetupCommandStatus)
+		e := c.SetupCommand.Wait()
+		if e == nil {
+			return
+		}
+		if exitErr, ok := e.(*exec.ExitError); ok {
+			c.SetupCommandStatus <- exitErr
+		} else {
+			log.Error("Could not handle exit error of setup command: ", e)
```
andrew-leung: Should we push e through the channel?

fslocker/fslocker.go
```diff
@@ -118,8 +120,9 @@ func (locker *FSLocker) ExclusiveLock(path string, timeout *time.Duration) (*Exc
 }
 
 // SharedLock tries to get an exclusive Lock on the path
```
andrew-leung: Comment should be s/exclusive/shared/

vpc/allocate/allocate_network.go
```diff
@@ -51,10 +51,15 @@ var AllocateNetwork = cli.Command{ // nolint: golint
 			Usage: "How long to wait for security groups to converge, in seconds",
 			Value: 10 * time.Second,
 		},
+		cli.DurationFlag{
+			Name:  "wait-for-sg-lock-timeout",
+			Usage: "How long to wait for other users, if the SG is in use",
```
andrew-leung: In what scenario would we now expect contention on the same SG? If the SG should be released prior to signaling master of completion, is this just waiting in unexpected/error scenarios?

sargun: This timeout will never get hit in the case of the same set of SGs. Only when changing SGs. It's basically, because v2 engine is going to stick around for the forseeable, and we need a solution for that. This is kind of like launchguard-lite.

executor/runtime/docker/docker.go
sargun: If IgnoreLaunchGuard is true, we should be using the kill initiated state, so we should never have two tasks using the same interface for different security groups.

executor/runtime/docker/docker.go
sargun: That was an error.

PR #136, opened 2018-06-03
Allow adjusting Ip refresh timeout

sargun:


executor/runtime/docker/docker.go
```diff
@@ -150,6 +151,11 @@ func NewConfig() (*Config, []cli.Flag) {
 			Value:       time.Minute * 1,
 			Destination: &cfg.waitForSecurityGroupLockTimeout,
 		},
+		cli.DurationFlag{
+			Name:        "titus.exwecutor.networking.ipRefreshTimeout",
```
andrew-leung: Typo?

vpc/allocate/ip_pool_manager.go
```diff
@@ -59,7 +59,8 @@ func (mgr *IPPoolManager) assignMoreIPs(ctx *context.VPCContext, batchSize int)
 	}
 
 	originalIPCount := len(mgr.networkInterface.IPv4Addresses)
-	for i := 0; i < 10; i++ {
+	now := time.Now()
+	for time.Since(now) < ipRefreshTimeout {
```
andrew-leung: In the comment on line 69, aren't we not retrying in that case?

executor/runtime/docker/docker.go
sargun: Typo.

vpc/allocate/ip_pool_manager.go
sargun: Er, that means we retry to "lease" or "allocate" an IP from from the IPs assigned to this ENI, whereas this is to "assign" IPs. I can change the comment if it helps?

PR #145, opened 2018-06-25
Reduce the available IP addresses 24 per ENI on the r4.16xlarge

sargun:
This is to prevent "overshoot" - otherwise we would do 31. The reason
why we're doing this is TITUS-2255. :( Bugs in AWS.

### DIFF
--- a/vpc/limits.go
+++ b/vpc/limits.go
@@ -125,8 +125,8 @@ var interfaceLimits = map[string]map[string]limits{
 		},
 		"16xlarge": limits{
 			interfaces:               15,
-			ipAddressesPerInterface:  50,
-			ip6AddressesPerInterface: 50,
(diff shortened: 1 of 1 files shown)


corindwyer:
What the reasoning for 24 as opposed to say 28?

sargun:
@corindwyer Because of the brokeness of the AWS allocate IP address API, we don't have a way to actually "know" what the 32nd+ IP is, because allocation is asynchronous. A pretty common thing we see today is allocation races, where allocation (1) doesn't show new IPs for N minutes, and then we do allocation (2). If we do 28, and 4 of these allocations stack, we've broken our headroom (which happens when the VPC APIs slow down). I feel like 8 is a better number.

In addition, I'd like to leave the batching on initially to see if we can maybe live with 2, or 3 as the batch as opposed to 4.

PR #151, opened 2018-06-29
Added Atlas counter titus.executor.S3Uploader.uploadMB to report log …

rickkw:
Report metrics on log file size uploaded in MB

### DIFF
--- a/uploader/s3.go
+++ b/uploader/s3.go
@@ -9,6 +9,7 @@ import (
 
 	"context"
 
+	"github.com/Netflix/metrics-client-go/metrics"
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/ec2metadata"
 	"github.com/aws/aws-sdk-go/aws/session"
@@ -28,10 +29,11 @@ type S3Uploader struct {
 	log        logrus.FieldLogger
 	bucketName string
 	s3Uploader *s3manager.Uploader
+	metrics    metrics.Reporter
 }
 
 // NewS3Uploader creates a new instance of an S3 uploader
-func NewS3Uploader(log logrus.FieldLogger, bucket string) Uploader {
+func NewS3Uploader(m metrics.Reporter, log logrus.FieldLogger, bucket string) Uploader {
 	region, err := getEC2Region()
 	if err != nil {
 		panic(err)
@@ -40,6 +42,7 @@ func NewS3Uploader(log logrus.FieldLogger, bucket string) Uploader {
 	u := &S3Uploader{
 		log:        log,
 		bucketName: bucket,
+		metrics:    m,
 	}
 
 	session, err := session.NewSession(&aws.Config{
@@ -90,6 +93,21 @@ func (u *S3Uploader) Upload(ctx context.Context, local string, remote string, ct
 	return u.uploadFile(ctx, f, remote, contentType)
 }
 
+// countingReader is a wrapper of io.Reader to count number of bytes read
+type countingReader struct {
+	reader    io.Reader
+	bytesRead int
+}
+
+// Read aggregates number of bytes read
+func (r *countingReader) Read(p []byte) (n int, err error) {
+	n, err = r.reader.Read(p)
+	if err == nil {
+		r.bytesRead += n
+	}
+	return
+}
+
 // UploadFile writes a single file only to S3!
 func (u *S3Uploader) uploadFile(ctx context.Context, local io.Reader, remote string, contentType string) error {
 	u.log.Printf("Attempting to upload file from: %s to: %s", local, path.Join(u.bucketName, remote))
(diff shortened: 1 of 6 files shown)


uploader/s3.go
```diff
@@ -87,6 +90,11 @@ func (u *S3Uploader) Upload(ctx context.Context, local string, remote string, ct
 		}
 	}()
 
+	// Record number of bytes being uploaded.  At this point, file must exist.
+	if stat, err := os.Stat(local); err == nil {
+		u.metrics.Counter("titus.executor.S3Uploader.uploadMB", int(stat.Size()/1024/1024), nil)
```
corindwyer: I think this needs tags or the only information it will tell me is that an agent uploaded some megabytes to S3 during some time frame.

rickkw: what tags can you think of?  Filename may not be a good choice if it rotates and have a numerical or timestamp suffix.  I assume the standard tags already includes container ID, etc.

corindwyer: I was just thinking container ID. Does the `u.metrics.Counter` method automatically add container information?

sargun: nf.node is usually set to the container's task ID from my understanding. I might suggest just asking the insights team all the the nf.* tags they associate with task metrics and use those.

uploader/s3.go
```diff
@@ -87,6 +90,11 @@ func (u *S3Uploader) Upload(ctx context.Context, local string, remote string, ct
 		}
 	}()
 
+	// Record number of bytes being uploaded.  At this point, file must exist.
+	if stat, err := os.Stat(local); err == nil {
```
sargun: Don't use os.Stat here because then you wont catch stuff being uploaded via `UploadPartOfFile`.

You should move the code into `uploadFile`. I might see if there's a simple io.Reader wrapper to get callbacks as data is read.

rickkw: fixed

uploader/s3.go
```diff
@@ -90,24 +93,43 @@ func (u *S3Uploader) Upload(ctx context.Context, local string, remote string, ct
 	return u.uploadFile(ctx, f, remote, contentType)
 }
 
+// CountingReader is a wrapper of io.Reader to count number of bytes read
+type CountingReader struct {
```
sargun: Why is `CountingReader` exposed outside of the package? It should be `countingReader` if it's not used outside of the package (equivalent to private in Java).

uploader/s3.go
```diff
@@ -90,24 +93,43 @@ func (u *S3Uploader) Upload(ctx context.Context, local string, remote string, ct
...
+	reader    io.Reader
+	BytesRead int
```
sargun: Might want to make this an int64?

uploader/s3.go
```diff
@@ -90,24 +93,43 @@ func (u *S3Uploader) Upload(ctx context.Context, local string, remote string, ct
...
+
 // UploadFile writes a single file only to S3!
 func (u *S3Uploader) uploadFile(ctx context.Context, local io.Reader, remote string, contentType string) error {
 	u.log.Printf("Attempting to upload file from: %s to: %s", local, path.Join(u.bucketName, remote))
 	if contentType == "" {
 		contentType = defaultS3ContentType
 	}
 
+	// wrap input io.Reader with a counting reader
+	reader := &CountingReader{local, 0}
```
sargun: In general, it's better to initialize with explicit field names. Also, by default in Go, uninitialized fields have default values. For ints, this is 0.

uploader/s3.go
```diff
@@ -90,24 +93,43 @@ func (u *S3Uploader) Upload(ctx context.Context, local string, remote string, ct
...
+
 	result, err := u.s3Uploader.UploadWithContext(ctx, &s3manager.UploadInput{
 		ACL:         aws.String(defaultS3ACL),
 		ContentType: aws.String(contentType),
 		Bucket:      aws.String(u.bucketName),
 		Key:         aws.String(remote),
-		Body:        local,
+		Body:        io.Reader(reader),
```
sargun: Is this cast neccessary here?

uploader/s3_test.go
```diff
@@ -0,0 +1,25 @@
+package uploader
+
+import (
+	"io"
+	"strings"
+	"testing"
+)
+
+func TestCountingReader_Read(t *testing.T) {
```
sargun: `TestCountingReader_Read` -> `TestCountingReaderRead`

uploader/s3_test.go
```diff
@@ -0,0 +1,25 @@
...
+	const data = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"
+	reader := strings.NewReader(data)
+	countingReader := &CountingReader{reader: reader}
+
+	buffer := make([]byte, 10)
+	for {
+		_, err := countingReader.Read(buffer)
```
sargun: You can use:
* https://golang.org/pkg/io/ioutil/#ReadAll
* https://golang.org/pkg/io/#Copy with `ioutil.discard`

There's also a library which makes asserting on these errors easier: https://github.com/stretchr/testify

You can do something like `assert.Equal(t, len(data), n)`

uploader/s3_test.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"io/ioutil"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCountingReader_Read(t *testing.T) {
```
sargun: Huh, I thought there was a comment to remove this underscore?

PR #152, opened 2018-06-29
allow CMD on images to be overriden

fabiokung:
### Description of the Change

The control plane will set the new `Process` field for jobs that specify a custom command. The new (array) entrypoint and command follow the same rules as Docker.

The old flat string entrypoint has precedence, to allow for a smooth transition without breaking existing jobs. The rollout plan described at Netflix/titus-api-definitions#52


executor/runtime/types/types_test.go
```diff
@@ -23,10 +27,71 @@ func TestEntrypointParse(t *testing.T) {
...
+			TitusInfo: &titus.ContainerInfo{
+				Process: &titus.ContainerInfo_Process{
+					Entrypoint: entrypoint,
+					Command:    []string{"sleep", "1"},
+				},
+			},
+		}
+
+		entrypoint, cmd, err := c.Process()
+		if err != nil {
```
sargun: `require.NoError(err)`?

executor/runtime/types/types_test.go
```diff
@@ -23,10 +27,71 @@ func TestEntrypointParse(t *testing.T) {
...
+	assert.Len(t, cmd, 0)
+}
+
+func TestDefaultIsNil(t *testing.T) {
+	c := Container{
+		TitusInfo: &titus.ContainerInfo{},
+	}
+
+	entrypoint, cmd, err := c.Process()
+	if err != nil {
```
sargun: `require.NoError`

hack/test-images/shell-entrypoint/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM ubuntu:xenial
```
sargun: Please add instructions to build this in the .circle config.

.gitignore
```diff
@@ -94,5 +94,5 @@ checkstyle-result.xml
 # unignore directories named titus-executor
 !*/titus-executor/
 
-test-standalone.log
```
sargun: Why did you remove these two?

hack/test-images/shell-entrypoint/Dockerfile
sargun: Err, call the build command

.gitignore
fabiokung: the IDE told me to, since there's another rule for `*.log`

executor/runtime/types/types_test.go
fabiokung: no other tests here are using that lib. I'd personally rather not to, unless you feel strongly about it.

.gitignore
sargun: neat.

PR #158, opened 2018-07-20
TITUS-895 Publish S3 upload size metrics with additional t.jobId and …

rickkw:
1. Enable titus.executor.S3Uploader.successfullyUploadedBytes metics
2. All titus-executor metrics are tagged with t.jobId and t.taskId by default
3. Additional tags can be enabled/disabled by fast properties titus.executor.enableDefaultTitusTags.  Default is true.

### DIFF
--- a/cmd/titus-executor/main.go
+++ b/cmd/titus-executor/main.go
@@ -28,6 +28,7 @@ func init() {
 }
 
 var logLevel string
+var enableDefaultTitusTags = true
 
 func setupLogging() {
 	switch logLevel {
@@ -49,6 +50,11 @@ var flags = []cli.Flag{
 		Name:        "titus.executor.logLevel",
 		Value:       "info",
 		Destination: &logLevel,
+	},
+	cli.BoolTFlag{
+		Name:        "titus.executor.enableDefaultTitusTags",
+		Destination: &enableDefaultTitusTags,
+		Usage:       "When enabled, titus-executor Atlas metrics are decorated with additional tags: t.jobId, t.taskId",
 	},
 }
 
@@ -88,7 +94,12 @@ func mainWithError(c *cli.Context, dockerCfg *docker.Config, cfg *config.Config)
 	case true:
 		m = metrics.Discard
 	default:
-		m = runner.NewReporter(metrics.New(ctx, log.StandardLogger(), tag.Defaults))
+		m = metrics.New(ctx, log.StandardLogger(), tag.Defaults)
+		if enableDefaultTitusTags {
+			// If fast property titus.executor.enableDefaultTitusTags is unset or set to true,
+			// wrap metrics.Reporter with runner.reporter that has default Atlas tags (t.jobId, t.taskId)
+			m = runner.NewReporter(m)
+		}
 		defer m.Flush()
 	}
 
--- a/cmd/titus-executor/main.go
+++ b/cmd/titus-executor/main.go
@@ -88,7 +88,7 @@ func mainWithError(c *cli.Context, dockerCfg *docker.Config, cfg *config.Config)
 	case true:
 		m = metrics.Discard
 	default:
-		m = metrics.New(ctx, log.StandardLogger(), tag.Defaults)
+		m = runner.NewReporter(metrics.New(ctx, log.StandardLogger(), tag.Defaults))
 		defer m.Flush()
 	}
 
--- a/executor/runner/runner.go
+++ b/executor/runner/runner.go
@@ -43,10 +43,11 @@ type task struct {
 // Runner maintains in memory state for the task runner
 type Runner struct { // nolint: maligned
 	// const:
-	metrics metrics.Reporter
-	runtime runtimeTypes.Runtime
-	config  config.Config
-	logger  *logrus.Entry
+	metrics   metrics.Reporter
+	runtime   runtimeTypes.Runtime
+	config    config.Config
+	dockerCfg *docker.Config
+	logger    *logrus.Entry
 
 	container *runtimeTypes.Container
 	watcher   *filesystems.Watcher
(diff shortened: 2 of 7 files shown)


executor/runner/runner.go
```diff
@@ -72,16 +73,17 @@ func New(ctx context.Context, m metrics.Reporter, logUploaders *uploader.Uploade
 	dockerRuntime := func(ctx context.Context, cfg config.Config) (runtimeTypes.Runtime, error) {
 		return docker.NewDockerRuntime(ctx, m, dockerCfg, cfg)
 	}
-	return WithRuntime(ctx, m, dockerRuntime, logUploaders, cfg)
+	return WithRuntime(ctx, m, dockerRuntime, logUploaders, cfg, &dockerCfg)
```
sargun: So, `dockerCfg` is supposed be Docker specific. Whereas WithRuntime, or runner.go is supposed to be container runtime agnostic.

sargun: `runner.go` is meant to be container runtime agnostic. `docker.Config` is specific to the Docker runtime. If there's a way we can avoid picking this up, that'd be ideal.

executor/runtime/docker/docker.go
```diff
@@ -165,10 +166,21 @@ func NewConfig() (*Config, []cli.Flag) {
...
+			Destination: &cfg.enableDefaultTitusTags,
+			Usage:       "When enabled, titus-executor Atlas metrics are decorated with additional tags, t.jobId, t.taskId",
+		},
 	}
 	return cfg, flags
 }
 
+// IsEnableDefaultTitusTags returns true if Atlas metrics should be tagged with additional tags
+// including t.jobId and t.taskId; return false otherwise
+func (c *Config) IsEnableDefaultTitusTags() bool {
```
sargun: This should really be in its own config. You can just make this a flag inside of the main.go for the titus-executor, and then submit a true / false to the NewRunner function.

executor/runner/runner.go
```diff
@@ -304,6 +307,21 @@ func (r *Runner) runContainer(ctx context.Context, startTime time.Time, updateCh
...
 }
 
+func (r *Runner) maybeSetDefaultTags() {
+	if r.dockerCfg.IsEnableDefaultTitusTags() {
+		// Initialize metrics.Reporter with default tags t.jobId and t.taskId
+		tagger, ok := r.metrics.(tagger)
+		if ok {
+			tags := map[string]string{
+				"t.jobId":  r.container.TitusInfo.TitusProvidedEnv["TITUS_JOB_ID"],
+				"t.taskId": r.container.TaskID,
```
sargun: You need to check if the `TITUS_JOB_ID` is there, otherwise do not set it.

executor/runner/runner.go
```diff
@@ -304,6 +307,21 @@ func (r *Runner) runContainer(ctx context.Context, startTime time.Time, updateCh
 	r.monitorContainer(ctx, startTime, statusChan, updateChan, details)
 }
 
+func (r *Runner) maybeSetDefaultTags() {
+	if r.dockerCfg.IsEnableDefaultTitusTags() {
+		// Initialize metrics.Reporter with default tags t.jobId and t.taskId
+		tagger, ok := r.metrics.(tagger)
```
sargun: I would do this cast at Runner initialization time, and not here.

executor/runner/runner.go
rickkw: docker.Config is already a parameter passed into the runner.New() function, so the Runner already knows about it.  I only pass that along into the runner.WithRuntime.  

Do you still want me to remove docker.Config from WithRuntime?  I can do it, but it won't address your comment about runner.go being agnostic of docker.Config.  Please let me know.

sargun: New uses  Docker.Config because it's the default runtime. It gets passed through to the Docker runtime, but the actual runner doesn't use it at all. We could remove it, and hoist that to a higher layer if we wanted to, but for now, please don't add additional dependencies.

rickkw: I agree with your original comment.  I don't really need docker.Config.  I will remove it.

cmd/titus-executor/main.go
```diff
@@ -50,6 +51,11 @@ var flags = []cli.Flag{
 		Value:       "info",
 		Destination: &logLevel,
 	},
+	cli.BoolTFlag{
+		Name:        "titus.executor.enableDefaultTitusTags",
+		Destination: &enableDefaultTitusTags,
+		Usage:       "When enabled, titus-executor Atlas metrics are decorated with additional tags: t.jobId, t.taskId",
+	},
```
rickkw: @sargun moved flag into main.go

cmd/titus-executor/main.go
```diff
@@ -89,6 +95,11 @@ func mainWithError(c *cli.Context, dockerCfg *docker.Config, cfg *config.Config)
 		m = metrics.Discard
 	default:
 		m = metrics.New(ctx, log.StandardLogger(), tag.Defaults)
+		if enableDefaultTitusTags {
+			// If fast property titus.executor.enableDefaultTitusTags is unset or set to true,
+			// wrap metrics.Reporter with runner.reporter that has default Atlas tags (t.jobId, t.taskId)
+			m = runner.NewReporter(m)
+		}
```
rickkw: that m implements the tagger interface is enough for the Runner to tell if extra Atlas tags is needed.  We don't need to pass any extra arguments into runner.New()

executor/runner/runner.go
```diff
@@ -73,21 +73,22 @@ func New(ctx context.Context, m metrics.Reporter, logUploaders *uploader.Uploade
...
-		logUploaders: logUploaders,
-		config:       cfg,
-		dockerCfg:    dockerCfg,
-		taskChan:     make(chan task, 1),
-		killChan:     make(chan struct{}),
-		UpdatesChan:  make(chan Update, 10),
-		StoppedChan:  make(chan struct{}),
+		logger:        logrus.NewEntry(logrus.StandardLogger()),
+		metrics:       m,
+		metricsTagger: metricsTagger,
```
sargun: Did these fields get reordered?

executor/runner/runner.go
```diff
@@ -308,17 +309,15 @@ func (r *Runner) runContainer(ctx context.Context, startTime time.Time, updateCh
...
-			r.logger.Infof("Set Atlas default tags to: %s", tags)
+	// If extra Atlas tags is enabled, initialize metrics.Reporter with default tags t.jobId and t.taskId
+	jobID := r.container.TitusInfo.TitusProvidedEnv["TITUS_JOB_ID"]
+	if r.metricsTagger != nil && len(jobID) > 0 {
+		tags := map[string]string{
+			"t.jobId":  jobID,
+			"t.taskId": r.container.TaskID,
 		}
+		r.metricsTagger.append(tags)
+		r.logger.Infof("Set Atlas default tags to: %s", tags)
```
sargun: `r.logger.WithFields(tags).Info("Set Atlas default tags")`

executor/runner/runner.go
rickkw: no, I stuck in metricsTagger after metrics because they are closely related.  Other than that I did not reorder anything else.

executor/runner/runner.go
rickkw: can't assign map[string]string to map[string]interface{}

sargun: Ok. :+1:

PR #159, opened 2018-07-24
Add lxcfs integration

sargun:
Add integration into LXCFS, so /proc/meminfo reflects the cgroup memory limit.

### DIFF
--- a/executor/runtime/docker/lxcfs.go
+++ b/executor/runtime/docker/lxcfs.go
@@ -11,7 +11,7 @@ const (
 	lxcfs = "/var/lib/lxcfs"
 )
 
-func getlxcfsbindmounts() []string {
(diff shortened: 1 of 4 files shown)


executor/runtime/docker/lxcfs.go
```diff
@@ -0,0 +1,25 @@
...
+	"path/filepath"
+
+	"golang.org/x/sys/unix"
+)
+
+const (
+	lxcfs = "/var/lib/lxcfs"
+)
+
+func getlxcfsbindmounts() []string {
```
fabiokung: minor nit: camel case

sargun:
Both. I noticed some tools read /proc/meminfo, and were broken.

PR #160, opened 2018-07-26
 add custom CMD (new style entrypoint/cmd) to metatron validation

fabiokung:
### Description of the Change

`entrypoint` and `command` are now verified as arrays by metatron. Backwards and forwards compatibility is being done by the metatron service itself.

### DIFF
--- a/executor/metatron/metatron.go
+++ b/executor/metatron/metatron.go
@@ -170,15 +170,6 @@ func createPassportDir(taskID string) error {
 	return os.MkdirAll(getMetatronOutputPath(taskID), os.FileMode(0700))
 }
 
-// RemovePassports removes a task's Metatron credential directory on the host
-func RemovePassports(taskID string) error {
-	err := os.RemoveAll(getPassportHostPath(taskID))
-	if err != nil && !os.IsNotExist(err) {
-		return fmt.Errorf("Failed to remove Metatron host path %s: %s", getPassportHostPath(taskID), err)
-	}
-	return nil
-}
-
 // GetPassports gets Metatron passports for a container/task and stores
 // them in a file system location.
 func (mts *TrustStore) GetPassports(ctx context.Context, encodedAppMetadata *string, encodedAppSig *string, taskID string, titusMetadata TitusMetadata) (*CredentialsConfig, error) {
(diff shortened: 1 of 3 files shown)


executor/metatron/metatron.go
```diff
@@ -170,19 +171,14 @@ func createPassportDir(taskID string) error {
 	return os.MkdirAll(getMetatronOutputPath(taskID), os.FileMode(0700))
 }
 
-// RemovePassports removes a task's Metatron credential directory on the host
-func RemovePassports(taskID string) error {
```
andrew-leung: I notice we removed the call to this function back in May (https://github.com/Netflix/titus-executor/commit/885ba4bfb9a8111b2a34c263c61e50baaeb80a3c#diff-957e82dddf339c98741109a28268e372). Does that mean we are leaving these creds laying around the host? I realize this is unrelated to this specific PR.

executor/runner/runner.go
```diff
@@ -468,25 +468,39 @@ func mkGetMetatronConfigFunc(mts *metatron.TrustStore) func(ctx context.Context,
...
 			LaunchTime:   (time.Now().UnixNano() / int64(time.Millisecond)),
 		}
-
 		metatronConfig, err := mts.GetPassports(
 			ctx,
-			c.TitusInfo.MetatronCreds.AppMetadata,
-			c.TitusInfo.MetatronCreds.MetadataSig,
-			c.TaskID,
-			titusMetadata)
+			*c.TitusInfo.MetatronCreds.AppMetadata,
```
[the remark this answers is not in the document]
fabiokung: that's a good idea, the old code wasn't checking it (there is a `nil` check in the code that calls this) so this should be safe as-is. But our future selves will probably appreciate it is a bit safer.

executor/metatron/metatron.go
sargun: Yeah. We are. It'd be nice to move to a model where the agent never has to write / read metatron creds from disk, and instead the metatron agent wrote a tarball to stdout.

fabiokung: Yes, I think you may be right.

PR #161, opened 2018-07-29
Fix fast properties

sargun:
This is a fix for properties which are dynamically loaded

### DIFF
--- a/properties/properties.go
+++ b/properties/properties.go
@@ -216,3 +216,33 @@ func incorrectTypeForFlagError(name, expectedTypeName string, value interface{})
 
 	return fmt.Errorf("Mismatched type for flag '%s'. Expected '%s' but actual is '%s'", name, expectedTypeName, valueTypeName)
 }
+
+// ConvertFlagsForAltSrc enables a flag to be used by altsrc
+func ConvertFlagsForAltSrc(flags []cli.Flag) []cli.Flag { // nolint: gocyclo
+	ret := make([]cli.Flag, len(flags))
+	for idx, untypedflag := range flags {
+		switch f := untypedflag.(type) {
+		case cli.IntFlag:
+			ret[idx] = altsrc.NewIntFlag(f)
+		case cli.DurationFlag:
+			ret[idx] = altsrc.NewDurationFlag(f)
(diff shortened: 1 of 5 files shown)


properties/properties.go
```diff
@@ -60,11 +60,14 @@ func (qis *QuiteliteInputSource) getVal(name string) (interface{}, bool) {
 func (qis *QuiteliteInputSource) Int(name string) (int, error) {
 	otherGenericValue, exists := qis.getVal(name)
 	if exists {
-		otherValue, isType := otherGenericValue.(int)
+		otherValue, isType := otherGenericValue.(float64)
```
rickkw: Does quitelite treat numbers, including integer, as float64?  Can its implementation change in the future that returns int for integer value?

properties/properties.go
```diff
@@ -45,7 +39,7 @@ func fetchQuiteLiteSource(alternateURIFlag string) (altsrc.InputSourceContext, e
 
 	}()
 	ret := QuiteliteInputSource{}
-	err = json.NewDecoder(resp.Body).Decode(&ret)
+	err = json.NewDecoder(resp.Body).Decode(&ret.valueMap)
```
rickkw: is this the main bug?

sargun: There were a few bugs:
1. This one
2. There was a problem with conversions of floats / ints.
3. The actual function wasn't being called (https://github.com/Netflix/titus-executor/pull/161/files#diff-7cae71039937aae66ba05a7cdcf970fcR75)
4. The flags needed to be enabled with altsrc. https://github.com/Netflix/titus-executor/pull/161/files#diff-587d2988572f886f6e2554c7e0ea6ba7R229

properties/properties.go
sargun: Ye
Yes, see here: https://golang.org/pkg/encoding/json/

> To unmarshal JSON into an interface value, Unmarshal stores one of these in the interface value:

```
bool, for JSON booleans
float64, for JSON numbers
string, for JSON strings
[]interface{}, for JSON arrays
map[string]interface{}, for JSON objects
nil for JSON null
```

PR #162, opened 2018-07-30
TITUS-2326 Cloned titus-log-viewer executable.  Preparing shared log …

rickkw:
1. Cloned titus-darion to titus-log-viewer executable
2. Modified darion shared code to work with log root in both inside and outside of container

### DIFF
--- a/cmd/titus-log-viewer/main.go
+++ b/cmd/titus-log-viewer/main.go
@@ -1,0 +1,32 @@
+package main
+
+import (
+	"net/http"
+
+	"github.com/Netflix/titus-executor/darion/api"
+	"github.com/Netflix/titus-executor/logsutil"
+	log "github.com/sirupsen/logrus"
+)
+
+func pingHandler(w http.ResponseWriter, r *http.Request) {
+	if _, err := w.Write([]byte("pong")); err != nil {
+		log.Error("Unable to respond with pong on ping handler: ", err)
+	}
+}
+
+func main() {
+	logsutil.MaybeSetupLoggerIfUnderSystemd()
+	log.Println("Titus log viewer is starting")
+	r := newMux()
+	if err := http.ListenAndServe(":8004", r); err != nil {
+		log.Error("Error: HTTP ListenAndServe: ", err)
+	}
+}
+
+func newMux() *http.ServeMux {
+	r := http.NewServeMux()
+	r.HandleFunc("/ping", pingHandler)
+	r.HandleFunc("/logs/", api.LogHandler)
+	r.HandleFunc("/listlogs/", api.ListLogsHandler)
+	return r
+}
--- a/cmd/titus-log-viewer/main_test.go
+++ b/cmd/titus-log-viewer/main_test.go
@@ -1,0 +1,18 @@
+package main
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+)
+
+func TestPing(t *testing.T) {
+	r := newMux()
+	server := httptest.NewServer(r)
+	defer server.Close()
+
+	_, err := http.Get(server.URL + "/ping")
+	if err != nil {
+		t.Fatal("Unexpected error")
+	}
+}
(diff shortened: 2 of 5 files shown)


cmd/titus-log-viewer/main.go
```diff
@@ -0,0 +1,32 @@
...
+	logsutil.MaybeSetupLoggerIfUnderSystemd()
+	log.Println("Titus log viewer is starting")
+	r := newMux()
+	if err := http.ListenAndServe(":8004", r); err != nil {
+		log.Error("Error: HTTP ListenAndServe: ", err)
+	}
+}
+
+func newMux() *http.ServeMux {
+	r := http.NewServeMux()
```
sargun: If you're cloning this, it probably makes sense to split up this (shared) code.

darion/api/handlers.go
```diff
@@ -39,8 +39,8 @@ func LogHandler(w http.ResponseWriter, r *http.Request) {
 }
 
 func logHandler(w http.ResponseWriter, r *http.Request, containerID, fileName string) {
-	filePath := filepath.Join(conf.ContainersHome, containerID, "logs", fileName)
-	fout, err := os.Open(filePath)
+	logPath := filepath.Join(buildLogLocationBase(containerID), fileName)
```
sargun: I don't know if the cloned version is meant to be multi-tenant, or single-tenant. If it's meant to be single-tenant (i.e. Darion per container), then you should remove everything around containerID.

darion/conf/conf.go
```diff
@@ -9,6 +9,10 @@ import (
 var (
 	// ContainersHome is the location in which the directories with the Task IDs lie, (and under that /logs/...)
 	ContainersHome = "/var/lib/docker/containers/"
+
+	// TitusTaskID initialized using TITUS_TASK_ID from OS environment variable.
+	// It can be used as a crude way to determine if this process is running inside a container.
+	TitusTaskID = os.Getenv("TITUS_TASK_ID")
```
sargun: See Above, RE: multi-tenant vs. single-tenant.

darion/api/handlers.go
rickkw: @sargun the goal of this PR is to fudge the darion code just enough that allows both multi-tenant and single-tenant to work (initially), so I can work on injecting it into the container.  This is not meant to be the final code.  I don't want to spend time refactoring this code (just yet).  Rather, I think it's better to focus on the process injection which has more unknowns.

sargun: Then I would keep this on your own branch for now, and wait to merge until it's ready, because this part of the code isn't touched very much.

cmd/titus-log-viewer/main.go
```diff
@@ -0,0 +1,32 @@
...
+	if _, err := w.Write([]byte("pong")); err != nil {
+		log.Error("Unable to respond with pong on ping handler: ", err)
+	}
+}
+
+func main() {
+	logsutil.MaybeSetupLoggerIfUnderSystemd()
+	log.Println("Titus log viewer is starting")
+	r := newMux()
+	if err := http.ListenAndServe(":8004", r); err != nil {
```
andrew-leung: Is this meant to run in a different network namespace where it won't clash on port with Darion?

cmd/titus-log-viewer/main_test.go
```diff
@@ -0,0 +1,18 @@
+package main
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+)
+
+func TestPing(t *testing.T) {
```
andrew-leung: What is this test actually testing? Looks like it's just using golang http packages.

fabiokung:
@rickkw are you still looking for reviews here, or should we wait until you have done testing in your own branch and are ready to open the final PR?

cmd/titus-log-viewer/main.go
rickkw: yes

cmd/titus-log-viewer/main_test.go
rickkw: honest I don't know.  I just cloned the test from darion.

rickkw:
@fabiokung per @sargun advice, I am leaving this PR open while working on injecting Darion inside containers.  No need to merge this PR and introduce risk, while this effort is still WIP.  I won't be merging this PR until I make progress with the actual injection work.

PR #164, opened 2018-08-16
Make the titus VPC globalgc try to scope enumeration of ENIs by VPC

sargun:
According to AWS, DescribeNetworkInterfaces should be scoped by VPC for efficiency


sargun:
Given this is currently causing issues in production, I'm going to merge it, and pick it up so we can use it on the global gc logic.

andrew-leung:
Sorry for the delay. LGTM. Just curious, a) what's the main motivation for scoping to a VPC (e.g., future multi-account support)? and b) what was the production issue given that each agent pool is in a single VPC today?

fabiokung:
LGTM

@andrew-leung I believe this was suggested by EC2 engineers to reduce the cost of these calls.

PR #165, opened 2018-08-16
Lock Down FUSE

sargun:


executor/runtime/docker/capabilities.go
```diff
@@ -0,0 +1,68 @@
...
+func setupAdditionalCapabilities(c *runtimeTypes.Container, hostCfg *container.HostConfig) error {
+	addedCapabilities := addAdditionalCapabilities(c, hostCfg)
+
+	// Privileged containers automaticaly deactivate seccomp and friends, no need to do this
+	fuseEnabled, err := c.GetFuseEnabled()
+	if err != nil {
+		return err
+	}
+	if fuseEnabled || c.TitusInfo.GetAllowNestedContainers() {
+		if _, ok := addedCapabilities["SYS_ADMIN"]; !ok {
```
fabiokung: minor: extract `"SYS_ADMIN"` to a `const`

executor/runtime/docker/capabilities.go
```diff
@@ -0,0 +1,68 @@
...
+			PathInContainer:   fuseDev,
+			CgroupPermissions: "rmw",
+		})
+		apparmorProfile = "docker-fuse"
+		seccompProfile = "fuse-container.json"
+
+	}
+	// We can do this here because nested containers can do everything fuse containers can
+	if c.TitusInfo.GetAllowNestedContainers() {
+		apparmorProfile = "docker-nested"
```
fabiokung: do the nested apparmor and seccomp profiles include everything that's in the fuse ones? (in case both `FUSE && nestedContainers` is enabled for a task)

sargun: @fabiokung Yeah, I thought the comment indicated that: `	// We can do this here because nested containers can do everything fuse containers can` -- The capabilities of nested containers are a superset of fuse containers.

sargun:
@fabiokung The only thing the FUSE profiles have beyond the default Docker profile are mounting.

executor/runtime/docker/capabilities.go
fabiokung: ugh, I missed that comment. LGTM

PR #167, opened 2018-08-17
IPv6

sargun:


sargun:
This PR is based on #165, because that has some API improvements and some testing bits that were added. Once that gets merged, I'll rebase on master to clean up the diff.
CC: @andrew-leung @fabiokung

vpc/allocate/ip_pool_manager.go
```diff
@@ -81,7 +83,37 @@ func (mgr *IPPoolManager) assignMoreIPs(ctx *context.VPCContext, batchSize int,
...
-func (mgr *IPPoolManager) allocate(ctx *context.VPCContext, batchSize int, ipRefreshTimeout time.Duration) (string, *fslocker.ExclusiveLock, error) {
+func (mgr *IPPoolManager) allocateIPv6(ctx *context.VPCContext, networkinterface *ec2wrapper.EC2NetworkInterface) (string, *fslocker.ExclusiveLock, error) {
+	configLock, err := mgr.lockConfiguration(ctx)
+	if err != nil {
+		ctx.Logger.Warning("Unable to get lock during allocation: ", err)
+		return "", nil, err
+	}
+	defer configLock.Unlock()
+
+	iface, err := ctx.Cache.DescribeInterface(ctx, networkinterface.InterfaceID)
```
fabiokung: do we assume each interface will always have enough IPv6s? A short comment about how we expect to allocate IPs to each ENI would be good here

sargun:
Unfortunately, this patchset is much messier than intended, because it moves around a lot of code in place.

PR #169, opened 2018-08-22
ENI GC version 2.0

sargun:
Switch to new ENI GC strategy  …
This ENI GC strategy does two things:
1. It marks all ENIs with a time they were created as a tag
2. When walking over all ENIs, and doing the describe call, it
   looking at the above tag

--
Of course, there will be old Titus agents out there that will
be deploying ENIs which are unlabeled, or where labeling might
fail. Therefore, we periodically walk all ENIs, and label
them with when we first see them.

### DIFF
--- a/vpc/backfilleni/backfilleni.go
+++ b/vpc/backfilleni/backfilleni.go
@@ -1,1 +1,171 @@
 package backfilleni
+
+import (
+	"time"
+
+	"github.com/Netflix/titus-executor/vpc"
+	"github.com/Netflix/titus-executor/vpc/context"
+	"github.com/Netflix/titus-executor/vpc/ec2util"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"gopkg.in/urfave/cli.v1"
+)
+
+type backfillConfiguration struct {
+	TagChunkSize int
+	Timeout      time.Duration
+	VPCID        string
+}
+
+// BackfillEni is the configuration for the command that labels creation time on new ENIs
+func BackfillEni() cli.Command {
+	cfg := &backfillConfiguration{}
+	backkFillEni := func(parentCtx *context.VPCContext) error {
+		if err := doBackfillEni(parentCtx, cfg); err != nil {
+			return cli.NewMultiError(cli.NewExitError("Unable to generate backfill", 1), err)
+		}
+		return nil
+	}
+	return cli.Command{ // nolint: golint
+		Name:   "backfill-eni-labels",
+		Usage:  "For ENIs which do not have a creation timestamp tag, this will go ahead and do its best to backfill it",
+		Action: context.WrapFunc(backkFillEni),
+		Flags: []cli.Flag{
+			cli.IntFlag{
+				Name:        "tag-chunk-size",
+				Value:       50,
+				Destination: &cfg.TagChunkSize,
+			},
+			cli.DurationFlag{
+				Name:        "timeout",
+				Value:       30 * time.Minute,
+				Destination: &cfg.Timeout,
+			},
+			cli.StringFlag{
+				Name:        "vpc-id",
+				Usage:       "Optionally specify a VPC, to speed up filtering requests",
+				EnvVar:      "EC2_VPC_ID",
+				Value:       "",
+				Destination: &cfg.VPCID,
+			},
+		},
+	}
+}
+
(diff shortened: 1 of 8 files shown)


vpc/backfilleni/backfilleni.go
```diff
@@ -0,0 +1,171 @@
...
+		untaggedEnis = append(untaggedEnis, describeAvailableResponse.NetworkInterfaces...)
+
+		if describeAvailableResponse.NextToken == nil {
+			return untaggedEnis, nil
+		}
+		describeAvailableRequest.SetNextToken(*describeAvailableResponse.NextToken)
+	}
+}
+
+func filterNetworkInterfaces(enis []*ec2.NetworkInterface) []*ec2.NetworkInterface {
```
fabiokung: nit: rename to `filterUntaggedInterfaces`

vpc/backfilleni/backfilleni.go
```diff
@@ -0,0 +1,171 @@
...
+				EnvVar:      "EC2_VPC_ID",
+				Value:       "",
+				Destination: &cfg.VPCID,
+			},
+		},
+	}
+}
+
+func getENIs(parentCtx *context.VPCContext, cfg *backfillConfiguration, svc *ec2.EC2) ([]*ec2.NetworkInterface, error) {
+	filters := []*ec2.Filter{
```
fabiokung: should we be tagging only `titus-managed` ENIs?

vpc/globalgc/globalgc.go
```diff
@@ -4,303 +4,175 @@ import (
...
-				Name:   aws.String("status"),
-				Values: aws.StringSlice([]string{"in-use"}),
-			},
-			{
-				Name:   aws.String("tag-key"),
-				Values: aws.StringSlice([]string{markTag}),
-			},
-		}, baseFilters...),
+	if *eni.Status != "available" {
+		panic(fmt.Sprintf("Interface status is %s instead of available", *eni.Status))
```
fabiokung: shouldn't this just skip this interface? (or is this the TODO on this function?)

vpc/globalgc/globalgc.go
```diff
@@ -4,303 +4,175 @@ import (
...
-					Key: aws.String(markTag),
-				},
-			},
-		}
-		_, err = ec2Client.DeleteTagsWithContext(parentCtx, deleteTagsInput)
-
-		if err != nil {
-			return err
-		}
+	_, err = svc.DeleteNetworkInterfaceWithContext(ctx, deleteNetworkInterfaceInput)
```
fabiokung: do we need to rate limit this? Or are we relying on the SDK to backoff on rate limit errors?

vpc/globalgc/globalgc.go
```diff
@@ -4,303 +4,175 @@ import (
...
-		cli.StringFlag{
-			Name:   "vpc-id",
-			Usage:  "Optionally specify a VPC, to speed up filtering requests",
-			EnvVar: "EC2_VPC_ID",
-			Value:  "",
-		},
-	},
+type globalGcConfig struct {
+	timeout    time.Duration
+	detachTime time.Duration
```
fabiokung: should probably call this `creationTime`?

vpc/globalgc/globalgc.go
```diff
@@ -4,303 +4,175 @@ import (
...
+			cli.DurationFlag{
+				Name:        "timeout",
+				Usage:       "Maximum amount of time allowed running GC",
+				Value:       time.Minute * 5,
+				Destination: &cfg.timeout,
+			},
+			cli.DurationFlag{
+				Name:        "detach-time",
+				Usage:       "How long an ENI has to be detached before we will clean it up",
+				Value:       time.Minute * 30,
```
fabiokung: now that this represents "how long since disconnected ENIs were created", should this be bumped to a few days? It seems we will be potentially cleaning up ENIs that have been in use for more than 30min much earlier now. Or is it also the intent?

vpc/backfilleni/backfilleni.go
fabiokung: nvm, I see the filter on the description below.

vpc/globalgc/globalgc.go
sargun: The SDK itself has backoffs built-in to it.

vpc/globalgc/globalgc.go
sargun: If we try to delete an ENI and it's currently attached, the call fails.

sargun:
@fabiokung ENIs don't have CreatedAt timestamps as far as I can tell.

PR #174, opened 2018-08-30
track network preparation latencies

fabiokung:
### Description of the Change

emit a metric for network preparation latencies (when successful)

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -827,7 +827,12 @@ func (r *DockerRuntime) Prepare(parentCtx context.Context, c *runtimeTypes.Conta
 
 	if r.cfg.UseNewNetworkDriver {
 		group.Go(func() error {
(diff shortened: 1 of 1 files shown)


executor/runtime/docker/docker.go
```diff
@@ -827,7 +827,11 @@ func (r *DockerRuntime) Prepare(parentCtx context.Context, c *runtimeTypes.Conta
 
 	if r.cfg.UseNewNetworkDriver {
 		group.Go(func() error {
-			return prepareNetworkDriver(errGroupCtx, r.dockerCfg, c)
+			netErr := prepareNetworkDriver(errGroupCtx, r.dockerCfg, c)
+			if netErr == nil {
+				r.metrics.Timer("titus.executor.prepareNetworkTime", time.Since(dockerCreateStartTime), c.ImageTagForMetrics())
```
sargun: Is there a reason to tag this with image?

sargun: Also, it might be easier to just time this span alone?

PR #176, opened 2018-09-15
docker-runtime: Use single quotes in bash env file template

cHYzZQo:
Using single quotes ensures that values are exported as literal values and not evaluated. Also escapes any single quotes in the value.


cHYzZQo:
Are tests expected to pass? I get the same lint failures on master. @sargun does master pass for you?

sargun:
I know that linting has hit some issues temporarily. I think rob G. Will
look at it when he’s back. What’s the urgency of the change?

Sent from my iPhone

On Sep 14, 2018, at 23:02, cHYzZQo <notifications@github.com> wrote:

Are tests expected to pass? I get the same lint failures on master. @sargun
<https://github.com/sargun> does master pass for you?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/Netflix/titus-executor/pull/176#issuecomment-421534056>,
or mute the thread

cHYzZQo:
Not urgent, just wanted to make sure it wasn't something i missing. Thanks

cHYzZQo:
Rebased now that the linting fixes are in.

sargun:
@cHYzZQo Are you ready for review?

cHYzZQo:
Yes

sargun:
lgtm.

sargun:
@cHYzZQo This looks good to me, but the one thing that might be valuable is to see is '$' is in the string, and fall back to the old behaviour. I don't think anyone is doing string interpolation, but might as well shield ourselves?

cHYzZQo:
> one thing that might be valuable is to see is '$' is in the string, and fall back to the old behaviour.

Is that really what we want? I don't think that would have ever worked well because 1) the initial environment wouldn't do any interpolation so it makes the initial environment that entrypoint is running under and the environment after ```bash -l``` inconsistent. 2) order isn't guaranteed so it's hard to reliably reference any other user defined variable.

But if you are sure you want it i'll add it in.

sargun:
@cHYzZQo Makes sense. This looks good to me. @rgulewich do you want to take a look as well?

PR #178, opened 2018-09-24
Fix metadata service block

sargun:
### Description of the Change

This fixes a potential bypass allowing users to get to the real AWS metadata service.


sargun:
Waiting on @rgulewich's changes before this will be merged / pass.

vpc/bpf/filter.c
```diff
@@ -36,8 +36,10 @@ int classifier_egress_filter(struct __sk_buff *skb) {
 
 	if (eth->h_proto == __constant_htons(ETH_P_IP) && iph->daddr == __constant_htonl(METADATA_SERVICE_IP))
 		return TC_ACT_SHOT;
+
 	/* See the explanation behind this in allocation_network_linux.go */
```
fabiokung: that file does not exist anymore

vpc/setup/setup_linux.go
```diff
@@ -144,15 +144,15 @@ func configureQdiscsForLink(ctx *context.VPCContext, link netlink.Link) error {
...
 
 func setupIFBs(ctx *context.VPCContext) error {
-	err := setupIFB(ctx, vpc.IngressIFB, "classifier_ingress")
+	err := setupIFB(ctx, vpc.IngressIFB, "classifier_ingress", unix.ETH_P_IP)
 	if err != nil {
 		return err
 	}
 
-	return setupIFB(ctx, vpc.EgressIFB, "classifier_egress")
+	return setupIFB(ctx, vpc.EgressIFB, "classifier_egress", unix.ETH_P_ALL)
```
fabiokung: it seems that for `!unix.ETH_P_ALL`, the bpf filter is a noop and will always do `TC_ACT_OK`, or are you relying on it to validate the L2 and L3 header lengths and formats?

sargun: That should never happen (we should never get non-ethernet traffic on the interface veth that goes into the ifb).

fabiokung: my bad, I meant `!unix.ETH_P_IP`

sargun: Yes, except for the block rule, it passes through all IP traffic. The amount of non-IP traffic we receive should be minimal, but it is on my long-term TODO list (there's a JIRA for it), to make this capture and classify all container traffic, not just IPv4 traffic.

PR #181, opened 2018-09-28
This sets the default timeout and retries for the IAM service

sargun:
When the container hits the metadata service for its IAM credentials, we lazily request them.
The Boto client expects this API to return in less than one second and return credentials.
This bumps the retries, and timeout to a bigger number.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -68,6 +68,14 @@ func NewConfig() (*Config, []cli.Flag) {
 		hardCodedEnv: []string{
 			"NETFLIX_APPUSER=appuser",
 			"EC2_DOMAIN=amazonaws.com",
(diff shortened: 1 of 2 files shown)


config/config.go
```diff
@@ -68,6 +68,8 @@ func NewConfig() (*Config, []cli.Flag) {
 		hardCodedEnv: []string{
 			"NETFLIX_APPUSER=appuser",
 			"EC2_DOMAIN=amazonaws.com",
+			"AWS_METADATA_SERVICE_TIMEOUT=5",
```
rgulewich: Maybe add a comment that these are standard AWS SDK environment vars?  Other than that, LGTM.

PR #182, opened 2018-09-28
Don't crash when the docker daemon doesn't respond

rgulewich:
`titus-reaper` was crashing with "panic: runtime error: invalid memory address or nil pointer dereference" when there was an error talking to the Docker daemon.  The offending line in the stacktrace was titus-executor/reaper/reaper.go:109


sargun:
This should be turned back into a cron job.

sargun:
(but you don't need to do that; good job, this is a long outstanding fix)

PR #184, opened 2018-10-03
Remove unused cpu isolation mechanisms

gabrielhartmann:
1. CPU isolation behavior is no longer specified via fast properties.
1. The "opportunistic workload" mode was unused.
1. The "nanocpus" bandwidth control was unused.

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -71,17 +71,10 @@ export {{ $key }}='{{ $val | escape_sq }}'
 {{ end }}
 `
 
-const (
-	bandwidthMode = "bandwidth"
-	nanocpusMode  = "nanocpus"
-	sharesMode    = "shares"
-)
-
 // Config represents the configuration for the Docker titus runtime
 type Config struct { // nolint: maligned
 	userNamespaceFDEnabled          bool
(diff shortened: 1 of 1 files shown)


executor/runtime/docker/docker.go
```diff
@@ -332,32 +320,17 @@ func setupLoggingInfra(dockerRuntime *DockerRuntime) error {
...
-	logEntry := log.WithField("taskID", c.TaskID).WithField("bandwidthMode", cfsBandwidthMode).WithField("cpuBurst", cpuBurst)
+	logEntry := log.WithField("taskID", c.TaskID).WithField("cpuBurst", cpuBurst)
 
 	if cpuBurst {
 		logEntry.Info("Falling back to shares since CPU bursting is enabled")
 		setShares(logEntry, c, hostCfg)
 		return
 	}
 
-	if c.Resources.CPU == 0 {
```
sargun: Wanna just inline the setCFSBandwidth in here?

gabrielhartmann: That method still has internal fallback logic for `setNanoCpus`.  I'd prefer to leave that inside the `setCFSBandwidth` method.  I'm working on a subsequent PR to address that.

PR #185, opened 2018-10-04
Add Docker config validation

gabrielhartmann:
Remove fallback logic to nanos on violation of CFS bandwidth configuration.  Fail fast instead.  This change also moves code for configuration out of the large `docker.go` file.

### DIFF
--- a/executor/runtime/docker/config.go
+++ b/executor/runtime/docker/config.go
@@ -1,0 +1,127 @@
+package docker
+
+import (
+	"errors"
+	"time"
+
+	log "github.com/sirupsen/logrus"
+	"gopkg.in/urfave/cli.v1"
+)
+
+const minCfsBandwidth = 1000
+const maxCfsBandwidth = 1000000
+
+// Config represents the configuration for the Docker titus runtime
+type Config struct { // nolint: maligned
+	userNamespaceFDEnabled          bool
+	cfsBandwidthPeriod              uint64
+	tiniVerbosity                   int
+	batchSize                       int
+	burst                           bool
+	securityConvergenceTimeout      time.Duration
+	pidLimit                        int
+	prepareTimeout                  time.Duration
+	startTimeout                    time.Duration
+	bumpTiniSchedPriority           bool
+	waitForSecurityGroupLockTimeout time.Duration
+	ipRefreshTimeout                time.Duration
+}
+
+// NewConfig generates a configuration, with a set of flags tied to it for the docker runtime
+func NewConfig() (*Config, []cli.Flag) {
+	cfg := &Config{}
+	flags := []cli.Flag{
+		cli.BoolTFlag{
+			Name:        "titus.executor.userNamespacesFDEnabled",
+			Destination: &cfg.userNamespaceFDEnabled,
+		},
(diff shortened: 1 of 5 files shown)


executor/runtime/docker/config.go
```diff
@@ -0,0 +1,126 @@
+package docker
+
+import (
+	"errors"
+	log "github.com/sirupsen/logrus"
+	"gopkg.in/urfave/cli.v1"
+	"time"
+)
+
+const MinCfsBandwidth = 1000
```
sargun: use const (
MinCfsBandwidth = 1000
MaxCfsBandwidth = 1000000
)

sargun: actually, it appears these aren't used outside of this package. they should be locally scoped.

executor/runtime/docker/config.go
```diff
@@ -0,0 +1,126 @@
+package docker
+
+import (
+	"errors"
```
sargun: Please run goimports.

executor/runtime/docker/config.go
```diff
@@ -0,0 +1,126 @@
...
+	if args == nil {
+		args = []string{}
+	}
+
+	args = append([]string{"fakename"}, args...)
+
+	return cfg, app.Run(args)
+}
+
+func validate(dockerCfg *Config) error {
```
sargun: Why don't you use a validator (i.e. apply with error)  -- you need to implement the interface `errorableFlag`:

```

// Flag is a common interface related to parsing flags in cli.
// For more advanced flag parsing techniques, it is recommended that
// this interface be implemented.
type Flag interface {
        fmt.Stringer
        // Apply Flag settings to the given flag set
        Apply(*flag.FlagSet)
        GetName() string
}

// errorableFlag is an interface that allows us to return errors during apply
// it allows flags defined in this library to return errors in a fashion backwards compatible
// TODO remove in v2 and modify the existing Flag interface to return errors
type errorableFlag interface {
        Flag

        ApplyWithError(*flag.FlagSet) error
}
```

you should be able to do this by extending the IntFlag type with something like BoundedIntFlag.

executor/runtime/docker/config_test.go
```diff
@@ -0,0 +1,29 @@
+package docker
+
+import (
+	"github.com/stretchr/testify/assert"
```
sargun: goimports

executor/runtime/docker/docker_test.go
```diff
@@ -4,11 +4,11 @@ import (
 	"bytes"
 	"context"
 	"errors"
+	"github.com/Netflix/titus-executor/properties"
```
sargun: goimports.

executor/runtime/docker/config.go
gabrielhartmann: Done

executor/runtime/docker/config.go
gabrielhartmann: Thanks.  I looked around but didn't find this option.  I'll do it this way.

gabrielhartmann: `errorableFlag` is a private interface right?

PR #186, opened 2018-10-04
Allow attaching TTYs to containers

rgulewich:
This exposes a new passthrough parameter `titusParameter.agent.ttyEnabled`, which causes the container to be created with a tty (same as doing a `docker run --tty`).

### DIFF
--- a/executor/mock/jobrunner.go
+++ b/executor/mock/jobrunner.go
@@ -57,6 +57,8 @@ type JobInput struct {
 	CPU *int64
 	// StopTimeoutSeconds is the duration we wait after SIGTERM for the container to exit
 	KillWaitSeconds uint32
+	// Tty attaches a tty to the container via a passthrough attribute
+	Tty bool
 }
 
 // JobRunResponse returned from RunJob
@@ -201,14 +203,16 @@ func (jobRunner *JobRunner) StopExecutorAsync() {
 }
 
 // StartJob starts a job on an existing JobRunner and returns once the job is started
-func (jobRunner *JobRunner) StartJob(jobInput *JobInput) *JobRunResponse {
+func (jobRunner *JobRunner) StartJob(jobInput *JobInput) *JobRunResponse { // nolint: gocyclo
 	// Define some stock job to run
 	var jobID string
+
 	if jobInput.JobID == "" {
 		jobID = fmt.Sprintf("Titus-%v%v", r.Intn(1000), time.Now().Second())
 	} else {
 		jobID = jobInput.JobID
 	}
+
 	taskID := fmt.Sprintf("Titus-%v%v-Worker-0-2", r.Intn(1000), time.Now().Second())
 	env := map[string]string{
 		"TITUS_TASK_ID":          taskID,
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -140,5 +140,5 @@ $(GOVENDOR):
 .PHONY: testdeps
 testdeps: $(GOVENDOR)
 	$(GOVENDOR) install +local
-	# Fail if gometalinter is not present in $PATH:
+	# Fail if gometalinter is not present in PATH:
```
sargun: Wanna break this out into another PR / commit?

README.md
```diff
@@ -46,6 +46,16 @@ If you want to build a dpkg, without Docker, once the code is checked out, you c
 make build-standalone
 ```
 
+## Linting
+Linting is done via the [gometalinter](https://github.com/alecthomas/gometalinter) package, which runs various linters. In order to ensure
```
sargun: See above.

README.md
```diff
@@ -46,6 +46,16 @@ If you want to build a dpkg, without Docker, once the code is checked out, you c
...
+consistent versions of linters, install version 2.0.11 as per the instructions in "Initial setup steps" above.
+
+To run lint checks:
+```sh-session
+make validate
```
sargun: make metalinter FAST=1 should probably be a thing.

executor/mock/standalone/standalone_test.go
```diff
@@ -847,3 +848,16 @@ func testTwoCPUs(t *testing.T, jobID string) {
 		t.Fail()
 	}
 }
+
+func testTty(t *testing.T, jobID string) {
+	ji := &mock.JobInput{
+		ImageName:     ubuntu.name,
+		Version:       ubuntu.tag,
+		EntrypointOld: `/usr/bin/tty`,
```
sargun: Does this test to see if TTY is enabled (does it fail if TTY is disabled, if so, does it make sense to add a negative test?)

executor/mock/jobrunner.go
```diff
@@ -201,14 +203,21 @@ func (jobRunner *JobRunner) StopExecutorAsync() {
...
 	var jobID string
+	attrs := make(map[string]string)
+
 	if jobInput.JobID == "" {
 		jobID = fmt.Sprintf("Titus-%v%v", r.Intn(1000), time.Now().Second())
 	} else {
 		jobID = jobInput.JobID
 	}
+
+	if jobInput.Tty {
```
sargun: Why don't you put this after creating titus.ContainerInfo? kinda like how this gets set: https://github.com/Netflix/titus-executor/blob/master/executor/mock/jobrunner.go#L249-L251

Makefile
rgulewich: They're already separate commits, but I can break them into another PR, sure.

executor/mock/standalone/standalone_test.go
rgulewich: Yeah - `tty` will fail if there's no tty present.  I can add a negative test.

executor/mock/jobrunner.go
rgulewich: Good point - missed that.  Done.

PR #187, opened 2018-10-05
Linting docs, test and build changes

rgulewich:
This includes:
* Adding linting instructions to the README
* Fixing a build issue when you've been doing multiple builds on the same machine
* Allowing the docker in docker test timeout to be overridden
* A minor Makefile comment fix

### DIFF
--- a/hack/builder/titus-executor-builder.sh
+++ b/hack/builder/titus-executor-builder.sh
@@ -117,7 +117,7 @@ if [[ $num_debs -ne 1 ]]; then
     exit 1
 fi
 
-filename=${outdir}/*.deb
+filename=$(ls -t ${outdir}/*.deb | grep -v latest | head -n 1)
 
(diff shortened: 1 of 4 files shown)


hack/builder/titus-executor-builder.sh
```diff
@@ -117,7 +117,7 @@ if [[ $num_debs -ne 1 ]]; then
     exit 1
 fi
 
-filename=${outdir}/*.deb
+filename=$(ls -t ${outdir}/*.deb | grep -v latest | head -n 1)
```
sargun: `-1t`

sargun: Wait, what's wrong with selecting on latest?

rgulewich: `-1` is the default when `ls` is not run in a shell.  Latest is the symlink: we want to grab the source file here.

PR #188, opened 2018-10-10
fix #126: Don't query registry if image is already downloaded

### LINKED ISSUE #126: Check if Digest Image Already Downloaded
Right now, the executor always forces us to go back to the registry. It would be nice to have a mode where it didn't go back to registry and checked if we already downloaded an image.

rgulewich:
Fixes issue #126 - do an image inspect to determine if the image already exists.  If it does, don't do a pull.


rgulewich:
Circle CI build failing on push of the `no-entrypoint` image with "received unexpected HTTP status: 520 Origin Error"

executor/mock/jobrunner.go
```diff
@@ -168,6 +167,14 @@ func NewJobRunner() *JobRunner {
 		panic(err)
 	}
 
+	return cfg, dockerCfg
+}
+
+// NewJobRunner creates a new JobRunner with its executor started
+// in the background and the test driver configured to use it.
+func NewJobRunner() *JobRunner {
```
sargun: TODO: We should move the job runner into the module standalone, given it gets used nowhere else.

executor/mock/jobrunner.go
```diff
@@ -181,14 +188,40 @@ func NewJobRunner() *JobRunner {
...
 
 	return jobRunner
 }
 
+// DockerPull does a docker pull via DockerRuntime
+func DockerPull(imgName string, version string) (bool, error) {
+	cfg, dockerCfg := generateConfigs()
+
+	ctx, cancel := context.WithCancel(context.Background())
+	rt, err := docker.NewDockerRuntime(ctx, metrics.Discard, *dockerCfg, *cfg)
```
sargun: Add `defer cancel()`, that way cancel() gets called any time the stack unwinds. It's generally a good pattern.

executor/mock/jobrunner.go
```diff
@@ -181,14 +188,40 @@ func NewJobRunner() *JobRunner {
...
+		cancel:          cancel,
 		runner:          r,
 		shutdownChannel: make(chan struct{}),
 	}
 
 	return jobRunner
 }
 
+// DockerPull does a docker pull via DockerRuntime
+func DockerPull(imgName string, version string) (bool, error) {
```
sargun: Dumb question, any reason this code isn't in standalone_test? Are there any private methods that need to be accessed?

sargun: One nicety of putting it in standalone test, is you can pass testing.T to it, and then use assertify.

executor/mock/standalone/standalone_test.go
```diff
@@ -875,3 +876,28 @@ func testTtyNegative(t *testing.T, jobID string) {
 		t.Fail()
 	}
 }
+
+func testCachedDockerPull(t *testing.T, jobID string) {
+
+	var imgName = "nginx"
```
sargun: Is there a titusops image worth pulling that's small, so we can ensure it's not reused? 

Also, it's probably worth deleting the image before it's pulled the first time, because people run this test with long-running VMs.

executor/mock/standalone/standalone_test.go
```diff
@@ -875,3 +876,28 @@ func testTtyNegative(t *testing.T, jobID string) {
...
+	var imgVersion = "latest"
+	res, err := mock.DockerPull(imgName, imgVersion)
+	if err != nil {
+		t.Errorf("Got error during first docker pull: %+v", err)
```
sargun: Probably worth using testify:
example: https://github.com/Netflix/titus-executor/blob/master/fslocker/fslocker_test.go#L20

You can even add messages:
https://godoc.org/github.com/stretchr/testify/require#NoErrorf

executor/mock/standalone/standalone_test.go
```diff
@@ -875,3 +876,28 @@ func testTtyNegative(t *testing.T, jobID string) {
...
+	}
+
+	// Shouldn't be cached:
+	if res {
```
sargun: You can use assert, and add a message.

executor/runtime/docker/docker.go
```diff
@@ -558,8 +558,36 @@ func sleepWithCtx(parentCtx context.Context, d time.Duration) error {
...
+	}
+
+	log.Debugf("Image '%s' exists. Response: %+v", ref, resp)
+	return true, nil
+}
+
+// DockerPull returns true if the image was cached, and we didn't need to pull
+func (r *DockerRuntime) DockerPull(ctx context.Context, c *runtimeTypes.Container) (bool, error) {
+	imgName := c.QualifiedImageName()
+	imgExists, err := imageExists(ctx, r.client, imgName)
```
sargun: We should only do this for digests. Right now, the behaviour with labels is we always try to re-pull the image in case somewhat mutated it.

executor/runtime/docker/docker.go
```diff
@@ -558,8 +558,36 @@ func sleepWithCtx(parentCtx context.Context, d time.Duration) error {
...
+
+		return false, err
+	}
+
+	log.Debugf("Image '%s' exists. Response: %+v", ref, resp)
+	return true, nil
+}
+
+// DockerPull returns true if the image was cached, and we didn't need to pull
+func (r *DockerRuntime) DockerPull(ctx context.Context, c *runtimeTypes.Container) (bool, error) {
```
sargun: Might be worth tossing a metric or two in here.

executor/runtime/docker/docker.go
```diff
@@ -558,8 +558,36 @@ func sleepWithCtx(parentCtx context.Context, d time.Duration) error {
...
+	imgName := c.QualifiedImageName()
+	imgExists, err := imageExists(ctx, r.client, imgName)
+	if err != nil {
+		log.Warningf("Error inspecting image '%s', due to reason: %+v", imgName, err)
```
sargun: So, this might be worth logging at Error.

One of the neat features of out logging library is it supports structured fields. 

For arbitrary string fields:
https://godoc.org/github.com/sirupsen/logrus#WithField

For error fields:
https://godoc.org/github.com/sirupsen/logrus#WithError

executor/runtime/docker/docker.go
```diff
@@ -558,8 +558,36 @@ func sleepWithCtx(parentCtx context.Context, d time.Duration) error {
...
+		return false, err
+	}
+
+	if imgExists {
+		log.Infof("Not pulling image '%s': already exists", imgName)
```
sargun: 👍 Might be worth adding this as a structured field.

It probably makes sense to add a thing at the top like:
`entry := log.WithField("imageName", imgName)`

executor/mock/jobrunner.go
rgulewich: Nope - was just trying to follow convention.  It looked like all of the tests were in `standalone_test.go`, and all of the helpers were in `jobrunner.go`.

executor/mock/standalone/standalone_test.go
rgulewich: Oh, thank goodness.  The Go built in testing lib is pretty bare bones.

executor/runtime/docker/docker.go
sargun: ^^

PR #190, opened 2018-10-17
Add labels describing resource allocation to containers

gabrielhartmann:
Docker container creation events don't contain information about resources.  In order to avoid a second call to inspect a container after construction we put this information in _small_ labels on the container here.

```bash
$ docker run --rm --label com.netflix.titus.cpu=1 --cpu-shares 100 ubuntu:latest ls
...
```

```bash
$ docker events --format "{{json .}}" | jq .
{
  "status": "create",
  "id": "d15607d8bb28b8a8038cc80b49ef65e22b1669c25e3af07131be4e60595320e9",
  "from": "ubuntu:latest",
  "Type": "container",
  "Action": "create",
  "Actor": {
    "ID": "d15607d8bb28b8a8038cc80b49ef65e22b1669c25e3af07131be4e60595320e9",
    "Attributes": {
      "com.netflix.titus.cpu": "1",
      "image": "ubuntu:latest",
      "name": "gifted_vaughan"
    }
  },
  "scope": "local",
  "time": 1539819620,
  "timeNano": 1539819620158171600
}
...
```

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -19,7 +19,8 @@ const (
 )
 
 // Config contains the executor configuration
-type Config struct { // nolint: maligned
+type Config struct {
+	// nolint: maligned
 	// MetatronEnabled returns if Metatron is enabled
 	MetatronEnabled bool
 	// PrivilegedContainersEnabled returns whether to give tasks CAP_SYS_ADMIN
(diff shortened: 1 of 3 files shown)


sargun:
Does it make sense to add an integration test?

gabrielhartmann:
@sargun I did a quick search for tests when I was writing this but didn't see any tests checking for label correctness.  Do you have a pointer?

rgulewich:
@gabrielhartmann @sargun What about just adding a unit test for `NewContainer()` to `executor/runtime/container_test.go`?

sargun:
@rgulewich :+1:

gabrielhartmann:
@sargun @rgulewich: I added a unit test.  I also added a field for workload type.  Please have a second look.

PR #197, opened 2018-11-06
Add first portion of container SSH Daemon

sargun:
This PR is a WIP of adding a thing that makes it so we can embed sshd into the container.


sargun:
@rgulewich this is ready-to-review.

gabrielhartmann:
@sargun : could you please provide a little more description about what this PR does?  For instance, does this intend to take care of the lifecycle of the systemd services you'll be launching or just provide the capability to launch these services?

gabrielhartmann:
Thanks @sargun.  So what happens when the executor crashes?  Leaked systemd service?

sargun:
Are you asking what happens when the executor crash-stops, or when it gets into undefined state? @gabrielhartmann

gabrielhartmann:
crash-stops

sargun:
I'm glad you asked. 
So, the executor mount /proc/$PID1OFCONTAINER on a directory, which is like /var/lib/titus-inits/%i (where %i is the Task ID)

So, at the (Re)start of the systemd unit, it checks for the presence of the /var/lib/titus-inits/%i/ns (https://github.com/Netflix/titus-executor/pull/197/files#diff-8351ac04b205be9439973e0d6b78e916R3), which checks if the unit is alive. 

Since we join the pid namespace of the task using nsenter, the task will actually stop when the pid namespace of the container stops.

So, basically the crash-stop behaviour should be something like this:

1. Executor crashes
1. Mesos realizes this, and kills all processes in pid cgroup for the container
1. The pid namespace of the container gets torn down, because PID 1 in the container's pid namespace exits
1. All processes (including SSHD) stop
1. Systemd thinks about restarting the unit, but then realizes that the PID 1 of the container is dead, and since it's an ephemeral / templatized systemd unit, it goes away.

sargun:
No, it's a template unit. So, it doesn't leak.

sargun:
Templated units by themselves cannot start, they need an "instance" to be started. The instance lifecycle is tied to the lifecycle (whether or not it's up) of the PID1.

executor/runtime/docker/docker_ssh.go
```diff
@@ -0,0 +1,162 @@
...
+	iamProfileARN, err := c.GetIamProfile()
+	if err != nil {
+		return err
+	}
+	iamProfile, err := arn.Parse(iamProfileARN)
+	if err != nil {
+		return err
+	}
+
+	caData, err := ioutil.ReadFile(c.Config.ContainerSSHDCAFile)
```
rgulewich: Is this shared between all of the containers?

executor/runtime/docker/docker.go
```diff
@@ -806,8 +739,76 @@ func prepareNetworkDriver(parentCtx context.Context, cfg Config, c *runtimeTypes
...
+			return err
+		}
+	}
+
+	// so we replace @ with - to match Docker's image naming scheme, a la:
+	// [a-zA-Z0-9][a-zA-Z0-9_.-]
+	noDashes := strings.Replace(r.cfg.ContainerSSHDImage, ":", "-", -1)
+	noAts := strings.Replace(noDashes, "@", "-", -1)
+	noSlashes := strings.Replace(noAts, "/", "_", -1)
+	*containerName = "titus-sshd-" + noSlashes
```
rgulewich: Why is this based on the image, and not on the ID of the container that it's meant to ssh into?

executor/runtime/docker/docker_ssh.go
sargun: So, the configuration can be set at executor start up time to point to a different file. It gets copied into the container at startup time, so if the file changes, the container wont reflect it.

sargun: I guess, different question, what is "this" that you refer to?

executor/runtime/docker/docker.go
sargun: because this container / volume is shared amongst all containers that refer to the same image. As you can see, it's not deleted / terminated. This way, we can reuse the container for "VolumesFrom"

rgulewich:
@sargun Could you give a quick overview of how IAM accounts, certs, and ssh user accounts tie together?  As an app owner, what do I need to configure to get ssh access, and what accounts can I ssh in as?

executor/runtime/docker/docker_ssh.go
rgulewich: The CA file that this is pointing to.  I'm wondering if it's one CA file used for all ssh containers on the host, or if it's intended to change per SSH container.

executor/runtime/docker/docker.go
```diff
@@ -806,8 +739,76 @@ func prepareNetworkDriver(parentCtx context.Context, cfg Config, c *runtimeTypes
...
+	// Check if this container exists, if not create it.
+	// We don't check the error here, because there's no way
+	// to prevent us from accidentally calling this concurrently
+	_, tmpErr := r.client.ContainerCreate(ctx, cfg, hostConfig, nil, *containerName)
+	if tmpErr == nil {
+		return nil
+	}
+
+	// Do this with backoff
+	timer := time.NewTicker(100 * time.Millisecond)
```
rgulewich: Does this ever stop?  If the `ContainerCreate` call above fails for a reason that's not related to another container creation in progress (so it will never be created), what's the behaviour?

executor/runtime/docker/docker.go
rgulewich: So the container is shared, but there's a separate running sshd executable per container to ssh into (as per the titus-sshd systemd unit below)?

executor/runtime/docker/docker_ssh.go
sargun: It's configurable, but by default it's one CA file for all containers on the host (we can override it via config, or mutable deploy)

executor/runtime/docker/docker.go
sargun: Yes.

sargun:
@rgulewich that's managed by the CA infrastructure. It's...complicated.

executor/runtime/docker/docker.go
sargun: The timer itself will stop when we return from this function, see the `defer`. 

The loop itself can terminate under one of two conditions:
1) The container shows up and responds to inspect
2) If the context is cancelled. The context descends from:

```
// Prepare host state (pull image, create fs, create container, etc...) for the container
func (r *DockerRuntime) Prepare(parentCtx context.Context, c *runtimeTypes.Container, binds []string) error { // nolint: gocyclo
	var sshdContainerName string
	l := log.WithField("taskID", c.TaskID)
	l.WithField("prepareTimeout", r.dockerCfg.prepareTimeout).Info("Preparing container")

	ctx, cancel := context.WithTimeout(parentCtx, r.dockerCfg.prepareTimeout)
	defer cancel()
```

`	group, errGroupCtx := errgroup.WithContext(ctx)`

So, if the errgroup has an error, the context is cancelled. Alternatively, if the `func (r *Runner) prepareContainer(ctx context.Context, updateChan chan update) update` times out, it cancels the context.

Also, if we get a message from the master saying to die, it exits.

inject/titus-nsenter.c
```diff
@@ -0,0 +1,260 @@
...
+	},
+	{
+		.nstype	= CLONE_NEWUSER,
+		.name	= "user",
+	},
+	{
+		.nstype	= CLONE_NEWUTS,
+		.name	= "uts",
+	},
+	/* Order here is intentional. We do pid NS last. */
```
rgulewich: Why do we do pid last?

sargun: It opens up an exploit path if we do PID before any of the other ones. Specifically, the attack vector is that the container shows up in the PID namespace (/proc/...) in the container before it's changed over all its namespaces to container namespaces. 

If someone breaks in using some new vector that we don't know of while this change is happening, they can traverse back up to the host OS's namespaces.

PR #203, opened 2018-11-14
Generate interface names randomly when inserting them into a container

sargun:
### Description of the Change

This fixes the fact we always generate tmp-8081 as the temporary interface

### DIFF
--- a/vpc/allocate/setup_container_linux.go
+++ b/vpc/allocate/setup_container_linux.go
@@ -58,7 +58,8 @@ func doSetupContainer(parentCtx *context.VPCContext, netnsfd int, bandwidth uint
 		mtu = 1500
 	}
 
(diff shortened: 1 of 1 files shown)


vpc/allocate/setup_container_linux.go
```diff
@@ -58,7 +58,8 @@ func doSetupContainer(parentCtx *context.VPCContext, netnsfd int, bandwidth uint
 		mtu = 1500
 	}
 
-	containerInterfaceName := fmt.Sprintf("tmp-%d", rand.Intn(10000))
+	r := rand.New(rand.NewSource(time.Now().UnixNano()))
```
fabiokung: it might be a good idea to reuse the same `rand` object in `vpc/allocate/ip_pool_manager.go` as well (or at least seed that one too).

There are other un-seeded `rands` used for jitter in other `vpc/...` packages as well, might as well do the same for them?

PR #204, opened 2018-11-19
Fix short circuiting quitelite

sargun:


rgulewich:
@sargun - Changes look fine.  What will this be used for?

sargun:
@rgulewich To run Titus in OSS

PR #207, opened 2018-12-01
Optimistic metadataserver

sargun:
Refactor metadataserver to enable optimistic IAM fetch
    
* Cleanup titus metadata service init
* Make the metadata server us the local region's STS service

This also makes it so that when the metadata server is hit, and it fetches credentials, it'll keep refreshing them automatically, rather than requiring force refresh.

### DIFF
--- a/executor/runtime/types/types.go
+++ b/executor/runtime/types/types.go
@@ -26,9 +26,10 @@ import (
 const (
 	hostnameStyleParam = "titusParameter.agent.hostnameStyle"
 	// FuseEnabledParam is a container atttribute set to enable FUSE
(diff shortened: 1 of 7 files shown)


executor/runtime/types/types.go
```diff
@@ -380,6 +381,15 @@ func (c *Container) GetKeepLocalFileAfterUpload() (bool, error) {
 	return strconv.ParseBool(keepLocalFileAfterUploadStr)
 }
 
+// GetOptimisticIAMTokenFetch indicates whether or not we should delete log files after uploading them
```
rgulewich: Copy and pasta?

PR #209, opened 2018-12-13
make it easier to override the nflx env file

fabiokung:
All other nflx env files provided by nflx base images follow the convention of prefixing files in /etc/nflx/base-environment.d/ with a number in the NNN format. This will make it easier to
override variables defined in the file supplied by the titus executor, and ensure that overrides are loaded after the titus file (for the common use case of loading files in the alphabetical order).

cc @bmoyles @amit-git


fabiokung:
@bmoyles would you prefer we make it explicit it's a titus-provided file (e.g.: `200titus`) instead, to avoid issues with the existing `200runtime_context` being in a different format? Who knows what code out there is making assumptions about its contents...

bmoyles:
Yeah, that seems like a reasonable compromise.

PR #211, opened 2018-12-20
Systemd support

rgulewich:
For all containers:
- Mount /run as tmpfs (default size 128 MiB)

For systemd labeled containers (those running and image with the
`com.netflix.titus.systemd` label set to "true"):
- Mount `/run/lock` as its own tmpfs mount
- Tini exec's the container's init command so that it runs as pid 1
- Run them using the standard apparmor and seccomp profiles: no
  CAP_SYS_ADMIN requirement

Other notes:
- This requires that cgroup namespaces are enabled in docker, otherwise
  the systemd container will fail to come up due to not being able to
  create new cgroups.
- Move to Bionic for the systemd test image: the version of systemd that
  ships with it is able to start without CAP_SYS_ADMIN


.circleci/config.yml
```diff
@@ -32,7 +32,7 @@ jobs:
           name: Login into Docker hub
           command: docker login -u $DOCKER_USER -p $DOCKER_PASS
       - run: ./build.sh ubuntu
-      - run: ./build.sh ubuntu-systemd-xenial
+      - run: ./build.sh ubuntu-systemd-bionic
```
sargun: Can we add both images?

executor/runtime/docker/capabilities.go
```diff
@@ -58,9 +59,12 @@ func setupAdditionalCapabilities(c *runtimeTypes.Container, hostCfg *container.H
 	if c.TitusInfo.GetAllowNestedContainers() {
 		apparmorProfile = "docker-nested"
 		seccompProfile = "nested-container.json"
+		c.Env["TINI_UNSHARE"] = trueString
```
sargun: We can rip out the nested containers code if you want?

executor/runtime/docker/docker.go
```diff
@@ -468,6 +471,20 @@ func (r *DockerRuntime) dockerConfig(c *runtimeTypes.Container, binds []string,
 	// Maybe set cfs bandwidth has to be called _after_
 	maybeSetCFSBandwidth(r.dockerCfg.cfsBandwidthPeriod, c, hostCfg)
 
+	// Always setup tmpfs: it's needed to ensure Metatron credentials don't persist across reboots and for SystemD to work
+	tmpFsSize := int64(defaultRunTmpFsSize)
+	if hostCfg.Memory < tmpFsSize {
```
sargun: I don't think we need this. It should never happen, since container memory should never be so low. In fact, if system memory = tmpfs memory, aren't we effectively leaking double the RAM? 

We "trust" the master to do validation on resource dimensions.

sargun: I think this would also simplify the code a little bit, because then you could just hard code the below sizes, rather than fmt.Sprintf'ing

executor/runtime/docker/docker.go
```diff
@@ -638,6 +655,26 @@ func vpcToolPath() string {
...
 }
 
+// Use image labels to determine if the container should be configured to run SystemD
+func setSystemdRunning(imageInfo types.ImageInspect, c *runtimeTypes.Container) error {
+	l := log.WithField("imageName", c.QualifiedImageName())
+
+	if systemdBool, ok := imageInfo.Config.Labels[systemdImageLabel]; ok {
+		val, err := strconv.ParseBool(systemdBool)
+		if err != nil {
+			return err
```
sargun: wrap this error so we know where it's coming from (IIRC, the function is `errors.wrapf`

executor/runtime/docker/docker.go
```diff
@@ -638,6 +655,26 @@ func vpcToolPath() string {
 	return ret
 }
 
+// Use image labels to determine if the container should be configured to run SystemD
+func setSystemdRunning(imageInfo types.ImageInspect, c *runtimeTypes.Container) error {
+	l := log.WithField("imageName", c.QualifiedImageName())
+
```
sargun: Why not log the value here, rather than below, so if the value is "corrupted" we can at least figure out what's going on?

executor/runtime/docker/docker.go
```diff
@@ -638,6 +655,26 @@ func vpcToolPath() string {
...
+			return err
+		}
+
+		c.IsSystemD = val
+		l.Infof("SystemD image label set to %t", val)
+		return nil
+	}
+
+	l.Info("SystemD image label not set: not configuring container to run SystemD")
+	c.IsSystemD = false
```
sargun: Isn't this the default anyways, so this is redundant?

executor/runtime/docker/docker.go
sargun: Also, this logger should descend from the one already in `Prepare`

executor/runtime/docker/docker.go
```diff
@@ -910,7 +947,7 @@ func (r *DockerRuntime) Prepare(parentCtx context.Context, c *runtimeTypes.Conta
 		}
 
 		myImageInfo = imageInfo
-		return nil
+		return setSystemdRunning(*imageInfo, c)
```
sargun: You cannot do it in here because that's a concurrency violation.

sargun: Since this modifies "c" when other people could be using it

.circleci/config.yml
rgulewich: Why continue to keep the old one around, it it's never going to work without elevated privileges, and we're not going to use it in tests?

executor/runtime/docker/capabilities.go
rgulewich: Are we never going to use it again?  I wasn't sure.

sargun: We might, but for now, it's cruft that's not supported, and should never be set. Probably worth just getting rid of the code and throwing an error if we find it's set to true.

.circleci/config.yml
sargun: Good point, derp.

executor/runtime/docker/docker.go
rgulewich: Sure, that's fine.  When you pass in tmpfs volumes to docker at create time, the used tmpfs bytes count toward the cgroup's limit, so it doesn't double count.

rgulewich:
> I guess we cannot add a full end-to-end test until we get the new Docker, correct?

@sargun - Correct.  In the meantime, I added a separate functional test for this.

executor/runtime/container.go
```diff
@@ -61,6 +61,7 @@ func NewContainer(taskID string, titusInfo *titus.ContainerInfo, resources *runt
 		TitusInfo:          titusInfo,
 		Resources:          resources,
 		Env:                env,
+		IsSystemD:          false,
```
sargun: Isn't false the default, any reason to set it?

executor/runtime/docker/docker.go
```diff
@@ -660,6 +667,25 @@ func vpcToolPath() string {
...
 
+// Use image labels to determine if the container should be configured to run SystemD
+func setSystemdRunning(log *log.Entry, imageInfo types.ImageInspect, c *runtimeTypes.Container) error {
+	l := log.WithField("imageName", c.QualifiedImageName())
+
+	if systemdBool, ok := imageInfo.Config.Labels[systemdImageLabel]; ok {
+		l.Infof("SystemD image label set to %s", systemdBool)
+		val, err := strconv.ParseBool(systemdBool)
+		if err != nil {
+			return errors.Wrap(err, "Error parsing SystemD image label")
```
sargun: SystemD -> systemd

executor/runtime/docker/docker.go
```diff
@@ -660,6 +667,25 @@ func vpcToolPath() string {
 	return ret
 }
 
+// Use image labels to determine if the container should be configured to run SystemD
+func setSystemdRunning(log *log.Entry, imageInfo types.ImageInspect, c *runtimeTypes.Container) error {
+	l := log.WithField("imageName", c.QualifiedImageName())
+
+	if systemdBool, ok := imageInfo.Config.Labels[systemdImageLabel]; ok {
+		l.Infof("SystemD image label set to %s", systemdBool)
```
sargun: use withfield.

executor/runtime/docker/docker.go
```diff
@@ -660,6 +667,25 @@ func vpcToolPath() string {
...
+		val, err := strconv.ParseBool(systemdBool)
+		if err != nil {
+			return errors.Wrap(err, "Error parsing SystemD image label")
+		}
+
+		c.IsSystemD = val
+		return nil
+	}
+
+	l.Info("SystemD image label not set: not configuring container to run SystemD")
```
sargun: See above. Also, does it make sense to log on the negative?

executor/runtime/docker/docker.go
```diff
@@ -1699,6 +1728,11 @@ func (r *DockerRuntime) setupPostStartLogDirTiniHandleConnection2(parentCtx cont
 		return err
 	}
 
+	if err := setCgroupOwnership(parentCtx, c, cred); err != nil {
+		log.Error("Unable to setup container nesting: ", err)
```
sargun: Is this error message correct, given we've removed nesting (or at least begun to remove it), shouldn't it be something like "Unable to delegate cgroups?"? also, use withErr

executor/runtime/docker/docker_linux.go
```diff
@@ -13,6 +13,9 @@ import (
 	"time"
 	"unsafe"
```
sargun: Extra space.

executor/runtime/docker/docker_linux.go
```diff
@@ -181,6 +185,41 @@ func cleanupCgroups(cgroupPath string) error {
 	return nil
 }
 
+func setCgroupOwnership(parentCtx context.Context, c *runtimeTypes.Container, cred ucred) error {
+	if !c.IsSystemD {
+		return nil
+	}
+
+	cgroupPath := filepath.Join("/proc/", strconv.FormatInt(int64(cred.pid), 10), "cgroup")
```
sargun: stronv.itoa?

PR #213, opened 2019-01-12
Begin adding tests the the VPC tool

sargun:
The VPC tool was built under duress. This begins to add some real
tests to it.

### DIFF
--- a/vpc/allocate/ip_pool_manager_test.go
+++ b/vpc/allocate/ip_pool_manager_test.go
@@ -1,0 +1,150 @@
+package allocate
+
+import (
+	"context"
+	"io/ioutil"
+	"os"
+	"reflect"
+	"runtime"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/Netflix/titus-executor/fslocker"
+	vpcContext "github.com/Netflix/titus-executor/vpc/context"
+	"github.com/Netflix/titus-executor/vpc/ec2wrapper"
+	"github.com/aws/aws-sdk-go/aws/client"
+	"github.com/sirupsen/logrus"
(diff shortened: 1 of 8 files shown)


vpc/allocate/ip_pool_manager_test.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+	tni.freeIPs = func(deallocationList []string) error {
+		called = true
+		assert.Equal(t, ipList[1:], deallocationList)
+		return nil
+	}
+	tni.ipv4Addresses = ipList
+	assert.NoError(t, ipPoolManager.DoGc(ctx, time.Second))
+	assert.NoError(t, ipPoolManager.DoGc(ctx, time.Second))
+	time.Sleep(3 * time.Second)
```
fabiokung: I know you are just starting on this, but adding an abstraction for an injectable `type Clock interface` (or `func Now() time.Time`) that the `IPPoolManager` uses will help keep execution times for these tests manageable. Something to keep an eye on as you add more tests.

fabiokung: that way you can control the wallclock in tests with things like `clock.AdvanceBy(3 * time.Second)` instead of `time.Sleep(3 * time.Second)`.

sargun: Yeah, there's a bunch of places in the tests where I think it might make a lot of sense to mock-out the clock. In standalone tests too. 

Is there a clock library you know that allows that, but with a way to avoid skipping over multiple events?

fabiokung: Unfortunately not, I haven't used any in Go.

PR #214, opened 2019-01-21
Pessimistically allocate IPv6 addresses

sargun:
* This changes it so that it assumes the IPv4 and IPv6 address limit are
   the same. This is true for all the instance types we care about
 * This adds basic tests for IPv6 allocation
 * This only allocates 1 IPv6 address at ENI creation time (vs 3)
 * This adds tests around the allocation logic

### DIFF
--- a/vpc/allocate/ip_pool_manager.go
+++ b/vpc/allocate/ip_pool_manager.go
@@ -2,7 +2,6 @@ package allocate
 
 import (
 	"errors"
-	"math/rand"
 	"path/filepath"
 	"time"
 
@@ -10,8 +9,6 @@ import (
 	"github.com/Netflix/titus-executor/vpc"
 	"github.com/Netflix/titus-executor/vpc/context"
 	"github.com/Netflix/titus-executor/vpc/ec2wrapper"
-	"github.com/aws/aws-sdk-go/aws"
-	"github.com/aws/aws-sdk-go/service/ec2"
 	"golang.org/x/sys/unix"
 )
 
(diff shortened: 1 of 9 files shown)


vpc/allocate/ip_pool_manager.go
```diff
@@ -42,38 +39,31 @@ func (mgr *IPPoolManager) lockConfiguration(parentCtx *context.VPCContext) (*fsl
 	return parentCtx.FSLocker.ExclusiveLock(path, &timeout)
 }
 
-func (mgr *IPPoolManager) assignMoreIPs(ctx *context.VPCContext, batchSize int, ipRefreshTimeout time.Duration) error {
-	if len(mgr.networkInterface.GetIPv4Addresses()) >= vpc.GetMaxIPv4Addresses(ctx.InstanceType) {
+func assignMoreIPs(ctx *context.VPCContext, batchSize int, ipRefreshTimeout time.Duration, networkInterface ec2wrapper.NetworkInterface, ipAssignmentFunction ec2wrapper.IPAssignmentFunction, ipFetchFunction ec2wrapper.IPFetchFunction) error {
+	if len(ipFetchFunction()) >= vpc.GetMaxIPAddresses(ctx.InstanceType) {
```
rgulewich: Might be worth moving the various `len(ipFetchFunction())` calls to a variable for readability.

vpc/ec2wrapper/ec2metadata.go
```diff
@@ -174,6 +196,18 @@ func (mdc *EC2MetadataClientWrapper) getMetadata(path string) (string, error) {
 	return val, err
 }
 
+func ipStringToList(ips string) []string {
+	// This expects a non-\n terminated list of IPs separated by \n, and returns a normalized form
+	addresses := strings.Split(ips, "\n")
+	result := make([]string, len(addresses))
+	for idx, addr := range addresses {
+		sanitized := strings.Trim(strings.TrimSpace(addr), "\x00")
+		ip := net.ParseIP(sanitized)
```
rgulewich: Worth checking if this returns `nil`, since this list comes from an untrusted source?  I mean, it's _probably_ fine...

sargun: Fixed in next PR.

PR #215, opened 2019-01-22
Make it so you can disable IPv6

sargun:


rgulewich:
What happens if setup is called with `disable-ipv6=true`, but then allocate-network is called with `allocate-ipv6-address=true`?  Does it error out in a reasonable way?

PR #216, opened 2019-01-29
Use runc instead of docker-runc

rgulewich:
Docker 18.09.1 doesn't ship with docker-runc: it uses containerd's runc instead


rgulewich:
> Might make sense to add a dependency on docker greater than version X in the repo?

Good idea - done.

PR #219, opened 2019-02-06
Fix race in writing to volume container list

rgulewich:
Fix two goroutines racing when concurrently writing to `volumeContainers`.


executor/runtime/docker/docker.go
```diff
@@ -868,7 +868,10 @@ func (r *DockerRuntime) createVolumeContainer(ctx context.Context, containerName
 
 // Prepare host state (pull image, create fs, create container, etc...) for the container
 func (r *DockerRuntime) Prepare(parentCtx context.Context, c *runtimeTypes.Container, binds []string, startTime time.Time) error { // nolint: gocyclo
+	var metatronContainerName string
```
sargun: oh man. This sucks.

PR #220, opened 2019-02-07
Wait For Titus Isolate to Isolate Workload Before Starting Workload

sargun:
Titus isolate waits for Docker events, and such. This sometimes
can take a moment. In order to avoid racing, and have the application
read the CPU state, and then configure itself based on that CPU
state, we wait for titus-isolate up to some timeout.

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -83,6 +83,9 @@ type Config struct { // nolint: maligned
 	bumpTiniSchedPriority           bool
 	waitForSecurityGroupLockTimeout time.Duration
 	ipRefreshTimeout                time.Duration
+
+	titusIsolateBlockTime   time.Duration
+	enableTitusIsolateBlock bool
 }
 
 // NewConfig generates a configuration, with a set of flags tied to it for the docker runtime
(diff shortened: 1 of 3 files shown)


executor/runtime/docker/docker.go
```diff
@@ -138,6 +141,19 @@ func NewConfig() (*Config, []cli.Flag) {
 			Destination: &cfg.ipRefreshTimeout,
 			Value:       time.Second * 10,
 		},
+		cli.DurationFlag{
+			Name:   "titus.executor.titusIsolateBlockTime",
+			EnvVar: "TITUS_EXECUTOR_TITUS_ISOLATE_BLOCK_TIME",
+			// The default value inside of the Titus Isolate code is 5 seconds.
```
gabrielhartmann: 10 seconds

executor/runtime/docker/docker.go
```diff
@@ -1265,6 +1281,27 @@ func (r *DockerRuntime) processEFSMounts(c *runtimeTypes.Container) ([]efsMountI
 	return efsMountInfos, nil
 }
 
+func (r *DockerRuntime) waitForTini(ctx context.Context, listener *net.UnixListener, efsMountInfos []efsMountInfo, c *runtimeTypes.Container) (string, error) {
+	// This can block for up the the full ctx timeout
```
gabrielhartmann: ```... up to the ...```

executor/runtime/docker/docker_isolate.go
```diff
@@ -0,0 +1,57 @@
...
+	defer cancel()
+	requestURL := &url.URL{
+		Scheme: "http",
+		Host:   host,
+		Path:   fmt.Sprintf("/isolate/%s", taskID),
+	}
+
+	rq, err := http.NewRequest("GET", requestURL.String(), http.NoBody)
+	if err != nil {
+		logrus.WithError(err).Warn("Could not form HTTP Request to Isolate")
```
gabrielhartmann: return after this?

PR #222, opened 2019-02-09
Use root to run titus-metatrond, not the image's USER

rgulewich:
Starting the metatron-sync service was failing with:
```
"exec: \"/titus/metatron/bin/titus-metatrond\": stat /titus/metatron/bin/titus-metatrond: permission denied"
```
This is because `runc exec` will use the `USER` value of the container's image if it's set, rather than always running as root.  The process also requires `CAP_DAC_OVERRIDE` in order to actually run the executable.


Makefile
```diff
@@ -99,7 +99,7 @@ endif
 .PHONY: builder
 builder:
 	@echo '---> Building titusoss/titus-executor-builder'
-	$(DOCKER) build -t titusoss/titus-executor-builder hack/builder
+	$(DOCKER) build --no-cache -t titusoss/titus-executor-builder hack/builder
```
sargun: Why no-cache?

sargun: This will make builds kind of slow.

rgulewich: Temporary push to see if the tests were failing due to a cached systemd service file.

That's not the issue, though: I _think_ it might have something to do with how the `--user` option to `runc` plays with user namespaces (executing as 0:0 still gives permission denied), but I'm not sure yet.

PR #230, opened 2019-02-16
Apparmor nsenter

sargun:


executor/runtime/docker/capabilities.go
```diff
@@ -42,25 +44,21 @@ func setupAdditionalCapabilities(c *runtimeTypes.Container, hostCfg *container.H
 		}
 	}
 	seccompProfile := "default.json"
-	apparmorProfile := "docker-titus"
+	apparmorProfile := "docker_titus"
```
rgulewich: Why the move to underscores for the profiles?

inject/titus-nsenter.c
```diff
@@ -97,10 +97,28 @@ int populate_namespaces(int titus_pid_1_fd, int namespace_fds[]) {
...
 
-int __set_up_apparmor(int titus_pid_1_fd) {
+static int set_up_apparmor(char apparmor_profile[1024], int apparmor_fd) {
 	// This will break on older kernels
-	int containerfd, our_fd, n;
-	char buf[1024], writebuf[1024];
+	char writebuf[1024];
+	int n, ret = 0;
+
+	// we can us dprintf, but this just makes error handling cleaner
```
rgulewich: "use"

executor/runtime/docker/capabilities.go
sargun: https://gitlab.com/jo-so/apparmor/commit/0e0663e99e993e578e6092b5907fcbe5ae8988b1

sargun: Dashes break apparmor 🤦‍♂️

PR #239, opened 2019-03-12
Add API Protect

sargun:
API protect, when optionally enabled, is a feature that allows
for protection of automatically generated AWS credentials
by bounding them to a given host (VPC, NAT gateways).


cmd/titus-metadata-service/main.go
```diff
@@ -146,6 +162,21 @@ func main() {
 			EnvVar:      types.TitusMetatronVariableName,
 			Destination: &metatronEnabled,
 		},
+		cli.StringFlag{
+			Name:        "vpc-id",
+			EnvVar:      "EC2_VPC_ID",
+			Destination: &vpcID,
+		},
+		cli.StringFlag{
```
sargun: TODO: Remove.

cmd/titus-metadata-service/main.go
```diff
@@ -158,23 +189,43 @@ func main() {
...
+
 		if metatronEnabled {
 			log.Info("Metatron enabled!")
-			if signer, err = identity.NewDefaultSigner(); err != nil {
-				log.Fatal(err)
+			if signer, err := identity.NewDefaultSigner(); err != nil {
+				log.WithError(err).Fatal("Cannot instantiate new default signer")
+				mdscfg.Signer = signer
+			}
+			if container, err := readTaskConfigFile(titusTaskInstanceID); err != nil {
```
sargun: @rgulewich I hope it was okay to hoist this in here.

cmd/titus-metadata-service/main.go
```diff
@@ -158,23 +189,43 @@ func main() {
...
+				log.WithError(err).Fatal("Cannot read container config file")
+				mdscfg.Container = container
 			}
 		}
-		ms := metadataserver.NewMetaDataServer(context.Background(), backingMetadataServer, iamARN, titusTaskInstanceID, ipv4Address, region, optimistic, container, signer)
+
+		if ipv6Addresses != "" {
```
sargun: Technically EC2_IPV6s can have multiple IPs. I believe they get newline separated. We only care for the primary IP.

cmd/titus-metadata-service/main.go
rgulewich: Yeah - looks like I didn't push my local change that also did this, sorry. :(

metadataserver/iam.go
```diff
@@ -96,7 +111,18 @@ func newIamProxy(ctx context.Context, router *mux.Router, iamArn, titusTaskInsta
 	*/
 	router.HandleFunc("/security-credentials/{iamProfile}", proxy.specificInstanceProfile)
 
-	if optimistic {
+	if config.APIProtectEnabled {
+		apiProtectInfoLock.Lock()
+		ec2AWSCfg := aws.NewConfig().WithMaxRetries(3)
```
rgulewich: Why is `WithMaxRetries` set for this AWS config object, but not the STS one?

sargun: Because the ec2 APIs fail often due to rate limits and such. Whereas we hand-do retries in the STS codepath.

sargun: It fails open, which is nice as well.

cmd/titus-metadata-service/main.go
rgulewich: Might be helpful to move this comment into the code :)

PR #240, opened 2019-03-12
Add owner e-mail and job type labels for consumption by the cpu prediction model

gabrielhartmann:
### Description of the Change

These should be the last attributes needed by the CPU prediction model for some time.

### DIFF
--- a/executor/runtime/container.go
+++ b/executor/runtime/container.go
@@ -9,13 +9,17 @@ import (
 )
 
 const (
-	appNameLabelKey        = "com.netflix.titus.appName"
-	cpuLabelKey            = "com.netflix.titus.cpu"
-	memLabelKey            = "com.netflix.titus.mem"
-	diskLabelKey           = "com.netflix.titus.disk"
-	networkLabelKey        = "com.netflix.titus.network"
-	workloadTypeLabelKey   = "com.netflix.titus.workload.type"
-	titusTaskInstanceIDKey = "TITUS_TASK_INSTANCE_ID"
+	appNameLabelKey          = "com.netflix.titus.appName"
+	cpuLabelKey              = "com.netflix.titus.cpu"
+	memLabelKey              = "com.netflix.titus.mem"
+	diskLabelKey             = "com.netflix.titus.disk"
+	networkLabelKey          = "com.netflix.titus.network"
+	workloadTypeLabelKey     = "com.netflix.titus.workload.type"
+	ownerEmailLabelKey       = "com.netflix.titus.owner.email"
+	ownerEmailPassThroughKey = "titus.agent.ownerEmail"
+	jobTypeLabelKey          = "com.netflix.titus.job.type"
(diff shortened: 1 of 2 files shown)


executor/runtime/container.go
```diff
@@ -45,6 +49,8 @@ func NewContainer(taskID string, titusInfo *titus.ContainerInfo, resources *runt
 
 	env := cfg.GetNetflixEnvForTask(titusInfo, strMem, strCPU, strDisk, strNetwork)
 	labels[appNameLabelKey] = titusInfo.GetAppName()
+	labels[ownerEmailLabelKey] = titusInfo.PassthroughAttributes[ownerPassThroughKey]
+	labels[jobTypeLabelKey] = titusInfo.PassthroughAttributes[jobTypePassThroughKey]
```
gabrielhartmann: It's my understanding that if a value isn't present in the map you get the `zero value` which for strings is the empty string, which for me is desired behavior here.

executor/runtime/container.go
```diff
@@ -51,6 +55,12 @@ func NewContainer(taskID string, titusInfo *titus.ContainerInfo, resources *runt
 	labels[diskLabelKey] = strDisk
 	labels[networkLabelKey] = strNetwork
 
+	passthroughAttributes := titusInfo.GetPassthroughAttributes()
+	if passthroughAttributes != nil {
+		labels[ownerEmailLabelKey] = passthroughAttributes[ownerPassThroughKey]
+		labels[jobTypeLabelKey] = passthroughAttributes[jobTypePassThroughKey]
```
gabrielhartmann: It's my understanding that if a value isn't present in the map you get the zero value which for strings is the empty string, which for me is desired behavior.

executor/runtime/container.go
```diff
@@ -9,13 +9,17 @@ import (
...
-	workloadTypeLabelKey   = "com.netflix.titus.workload.type"
-	titusTaskInstanceIDKey = "TITUS_TASK_INSTANCE_ID"
+	appNameLabelKey          = "com.netflix.titus.appName"
+	cpuLabelKey              = "com.netflix.titus.cpu"
+	memLabelKey              = "com.netflix.titus.mem"
+	diskLabelKey             = "com.netflix.titus.disk"
+	networkLabelKey          = "com.netflix.titus.network"
+	workloadTypeLabelKey     = "com.netflix.titus.workload.type"
+	ownerEmailLabelKey       = "com.netflix.titus.owner.email"
+	ownerEmailPassThroughKey = "titus.agent.ownerEmail"
```
sargun: why not `com.netflix.titus`?

gabrielhartmann: This is the labeling scheme from the control plane as suggested by @corindwyer.
https://github.com/Netflix/titus-control-plane/blob/master/titus-server-master/src/main/java/com/netflix/titus/master/mesos/DefaultV3TaskInfoFactory.java#L66

PR #243, opened 2019-03-29
Fix titus isolate command wait

sargun:
Coveralls broke, so I'm removing that.

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -1669,6 +1669,7 @@ func (r *DockerRuntime) setupPostStartLogDirTiniHandleConnection(parentCtx conte
 	}
 
 	rootFile := files[0]
+	c.RegisterRuntimeCleanup(rootFile.Close)
 
 	// r.logDir(c), &cred, rootFile, nil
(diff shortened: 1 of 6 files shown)


executor/runtime/docker/docker.go
```diff
@@ -1678,7 +1679,16 @@ func (r *DockerRuntime) setupPostStartLogDirTiniHandleConnection(parentCtx conte
 func (r *DockerRuntime) setupPostStartLogDirTiniHandleConnection2(parentCtx context.Context, c *runtimeTypes.Container, cred ucred, rootFile *os.File) error { // nolint: gocyclo
 	group, errGroupCtx := errgroup.WithContext(parentCtx)
 
+	// This required (write) access to c.RegisterRuntimeCleanup
+	if err := mountContainerProcPid1InTitusInits(parentCtx, c, cred); err != nil {
```
gabrielhartmann: It was only the `/var/lib/titus-inits` path that was a problem, not `/var/lib/titus-environments`?

sargun: Correct.

PR #245, opened 2019-04-10
Add globalgc logging details

rgulewich:
Right now, GC jobs failing due to rate limit errors print "Unable to run global gc", but give no details as to what call failed.

### DIFF
--- a/vpc/globalgc/globalgc.go
+++ b/vpc/globalgc/globalgc.go
@@ -100,6 +100,7 @@ func fetchDisconnectedENIs(parentCtx *context.VPCContext, vpcID string, networkI
 		}
 		describeAvailableResult, err := svc.DescribeNetworkInterfacesWithContext(parentCtx, describeAvailableRequest)
 		if err != nil {
+			parentCtx.Logger.WithError(err).Error("DescribeNetworkInterfaces error")
 			return err
 		}
 		for _, networkInterface := range describeAvailableResult.NetworkInterfaces {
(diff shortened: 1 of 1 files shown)


vpc/globalgc/globalgc.go
```diff
@@ -115,20 +116,21 @@ func fetchDisconnectedENIs(parentCtx *context.VPCContext, vpcID string, networkI
 // TODO: Find fatal errors and return errors
 func cleanupENI(parentCtx *context.VPCContext, svc *ec2.EC2, detachTime time.Duration, eni *ec2.NetworkInterface) error {
 	ctx := parentCtx.WithField("NetworkInterfaceId", *eni.NetworkInterfaceId)
+	logger := ctx.Logger.WithField("eni-name", *eni.NetworkInterfaceId)
 	tags := ec2util.TagSetToMap(eni.TagSet)
 	networkCreationTime, ok := tags[vpc.ENICreationTimeTag]
 	if !ok {
-		ctx.Logger.Warning("ENI does not have creation time field")
+		logger.Warning("ENI does not have creation time field")
```
sargun: why not use the context logger?

vpc/globalgc/globalgc.go
```diff
@@ -115,20 +116,21 @@ func fetchDisconnectedENIs(parentCtx *context.VPCContext, vpcID string, networkI
 // TODO: Find fatal errors and return errors
 func cleanupENI(parentCtx *context.VPCContext, svc *ec2.EC2, detachTime time.Duration, eni *ec2.NetworkInterface) error {
 	ctx := parentCtx.WithField("NetworkInterfaceId", *eni.NetworkInterfaceId)
+	logger := ctx.Logger.WithField("eni-name", *eni.NetworkInterfaceId)
```
sargun: If you use the context logger, you shouldn't have to do this.

vpc/globalgc/globalgc.go
```diff
@@ -115,20 +116,21 @@ func fetchDisconnectedENIs(parentCtx *context.VPCContext, vpcID string, networkI
...
 	creationTime, err := time.Parse(time.RFC3339, *networkCreationTime)
 	if err != nil {
-		ctx.Logger.WithField(vpc.ENICreationTimeTag, *networkCreationTime).WithError(err).Error("Cannot parse")
+		logger.WithField(vpc.ENICreationTimeTag, *networkCreationTime).WithError(err).Error("Cannot parse")
 		return nil
 	}
 	timeSinceCreation := time.Since(creationTime)
 	if timeSinceCreation < detachTime {
-		ctx.Logger.WithField("timeSinceCreation", timeSinceCreation.String()).Debug("Not cleaning up ENI, too young")
+		logger.WithField("timeSinceCreation", timeSinceCreation.String()).Debug("Not cleaning up ENI, too young")
```
sargun: You can create a logEntry using withfield specifically here.

vpc/globalgc/globalgc.go
rgulewich: Nice, OK - I didn't realize that doing a `ctx.WithField()` also added fields to the logger.

vpc/globalgc/globalgc.go
rgulewich: No reason - was just trying to pick one of the loggers - this function was using both `parentCtx.Logger` and `ctx.Logger` before.  Based on your comment above, using `ctx.Logger` makes more sense.

PR #251, opened 2019-05-07
Remove Logging to Journald only under systemd

sargun:
This makes it so that the executor doesn't rely on CGO. It needed
cgo before to figure out if it was running as a systemd service
or not. This makes it assume it will always run as a systemd service
unless otherwise stated by console flags, rather than relying on
environment detection.


cmd/titus-reaper/main.go
```diff
@@ -18,9 +18,10 @@ func main() {
 	flag.BoolVar(&debug, "debug", false, "Turn on debug logging")
 	flag.Parse()
 
-	logsutil.MaybeSetupLoggerIfUnderSystemd()
 	if debug {
 		log.SetLevel(log.DebugLevel)
+	} else {
+		logsutil.MaybeSetupLoggerIfOnJournaldAvailable()
```
rgulewich: Why do this only when !debug?

sargun: Because if you're running with -debug, you're probably running from the console, and not a service?

PR #252, opened 2019-05-07
Return better error message on metatron initialization failure

rgulewich:
Return a useful error message to users, rather than a systemd error message.

### DIFF
--- a/executor/mock/standalone/standalone_test.go
+++ b/executor/mock/standalone/standalone_test.go
@@ -97,6 +97,8 @@ var (
 	}
 )
 
+const defaultFailureTimeout = time.Minute
+
 // This file still uses log as opposed to using the testing library's built-in logging framework.
 // Since we do not configure Logrus, we will just log to stderr.
 func TestStandalone(t *testing.T) {
@@ -146,6 +148,7 @@ func TestStandalone(t *testing.T) {
 		testTtyNegative,
 		testCachedDockerPull,
 		testMetatron,
+		testMetatronFailure,
 		testRunTmpFsMount,
 		testExecSlashRun,
 		testSystemdImageMount,
@@ -278,7 +281,7 @@ func testInvalidFlatStringAsCmd(t *testing.T, jobID string) {
 	jobRunner := mock.NewJobRunner(nil)
 	defer jobRunner.StopExecutor()
 	jobResponse := jobRunner.StartJob(ji)
-	ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
+	ctx, cancel := context.WithTimeout(context.Background(), defaultFailureTimeout)
 	defer cancel()
 	if err := jobResponse.WaitForFailureWithStatus(ctx, 127); err != nil {
 		t.Fatal(err)
@@ -312,7 +315,7 @@ func testEntrypointAndCmdFromImage(t *testing.T, jobID string) {
 	jobRunner := mock.NewJobRunner(nil)
 	defer jobRunner.StopExecutor()
 	jobResponse := jobRunner.StartJob(ji)
-	ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
+	ctx, cancel := context.WithTimeout(context.Background(), defaultFailureTimeout)
 	defer cancel()
 	if err := jobResponse.WaitForFailureWithStatus(ctx, 123); err != nil {
 		t.Fatal(err)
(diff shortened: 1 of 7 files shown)


executor/mock/standalone/standalone_test.go
```diff
@@ -97,6 +97,8 @@ var (
 	}
 )
 
+const defaultFailureTimeout = 60 * time.Second
```
sargun: time.Minute

executor/runtime/docker/docker_linux.go
```diff
@@ -37,15 +39,44 @@ type schedParam struct {
...
+	systemServiceStartTimeout = 90 * time.Second
+	umountNoFollow            = 0x8
+	sysFsCgroup               = "/sys/fs/cgroup"
+)
+
+var (
+	atlasServiceOpts = serviceOpts{
+		humanName: "atlas",
+		unitName:  "atlas-titus-agent",
+		required:  false,
```
sargun: You can leave this out.

sargun: (just let required default to true)

executor/runtime/docker/docker_linux.go
```diff
@@ -37,15 +39,44 @@ type schedParam struct {
...
-	systemServiceStartTimeout  = 90 * time.Second
-	umountNoFollow             = 0x8
-	sysFsCgroup                = "/sys/fs/cgroup"
+	titusInits                = "/var/lib/titus-inits"
+	systemServiceStartTimeout = 90 * time.Second
+	umountNoFollow            = 0x8
+	sysFsCgroup               = "/sys/fs/cgroup"
+)
+
+var (
```
sargun: In fact, you could make this a list, and then the start unit stuff down below could just walk the list

executor/runtime/docker/docker_linux.go
```diff
@@ -116,37 +147,85 @@ func setupSystemPods(parentCtx context.Context, c *runtimeTypes.Container, cfg c
...
-func startSystemdUnit(ctx context.Context, conn *dbus.Conn, required bool, taskID, unitName string) error {
-	qualifiedUnitName := fmt.Sprintf("%s@%s.service", unitName, taskID)
-	l := logrus.WithField("taskID", taskID).WithField("unit", unitName)
-	l.Infof("Starting systemd unit %s", qualifiedUnitName)
+func runServiceInitCommand(ctx context.Context, log *logrus.Entry, cID string, opts serviceOpts) error {
+	if opts.initCommand == "" {
+		return nil
+	}
+
+	var stdout bytes.Buffer
```
sargun: Why don't we combine sydout / stderr?

executor/runtime/docker/docker_linux.go
```diff
@@ -116,37 +147,85 @@ func setupSystemPods(parentCtx context.Context, c *runtimeTypes.Container, cfg c
...
 
+	l.WithField("args", cmdArgs).Infof("Running init command for %s service", opts.unitName)
+	cmd := exec.CommandContext(ctx, opts.initCommand, cmdArgs...)
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
+
+	err := cmd.Run()
+	if err != nil {
+		outputStr := stdout.String()
+		l.WithField("stdout", outputStr).WithField("stderr", stderr.String()).WithError(err).Errorf("error running init command for %s service: exit code %d", opts.unitName, cmd.ProcessState.ExitCode())
```
sargun: exit code and service should be structured fields. I would actually almost say stderr / stdout should be in the message because we can't index / query on them

executor/runtime/docker/docker_linux.go
```diff
@@ -116,37 +147,85 @@ func setupSystemPods(parentCtx context.Context, c *runtimeTypes.Container, cfg c
...
+
+		if len(outputStr) != 0 {
+			splitOutput := strings.Split(outputStr, "\n")
+			errStr := ""
+			// Find the last non-empty line in stdout and use that as the error message
```
sargun: Couldn't you do something like strings.trimwhitespace, split, and then just get the last thing in the slice?

executor/runtime/docker/docker_linux.go
rgulewich: That's a great idea.  Will do.

executor/runtime/docker/docker_linux.go
rgulewich: 2 reasons:

The first is that cmd.CombinedOutput doesn't accept a Context, and I want the process to be killed when the executor times out.

The second is that I'd like this "contract" with service owners to be as explicit and deterministic as possible: I don't want the message that gets bubbled up to users to change if the service starts printing a statement to stderr that it didn't previously.

executor/runtime/docker/docker_linux.go
rgulewich: OK, I think I'll just log stdout and stderr as messages separately.

My reason for including the service in the message is that it makes it much easier to figure out what's going on when you're using `journalctl` in normal, non-json output mode.  If there's a better way of getting that info without looking through piles of JSON output, I'd love to know it!

PR #254, opened 2019-05-08
Allow setting the size of /dev/shm

rgulewich:
Add a `titusParameter.agent.shmSize` passthrough property to set the size of `/dev/shm` in MiB.


executor/mock/standalone/standalone_test.go
```diff
@@ -1014,3 +1015,20 @@ func testSystemdImageMount(t *testing.T, jobID string) {
 		t.Fail()
 	}
 }
+
+// Test that the size of `/dev/shm` can be set
+func testShm(t *testing.T, jobID string) {
+	var mem int64 = 256
+	var shmSize int64 = 192
```
andrew-leung: What happens if `shmSize > mem`?

executor/mock/jobrunner.go
```diff
@@ -282,6 +285,9 @@ func (jobRunner *JobRunner) StartJob(jobInput *JobInput) *JobRunResponse { // no
 	if jobInput.Batch == "idle" {
 		ci.PassthroughAttributes["titusParameter.agent.batchPriority"] = "idle"
 	}
+	if jobInput.ShmSize != nil {
+		ci.PassthroughAttributes["titusParameter.agent.shmSize"] = strconv.FormatInt(*jobInput.ShmSize, 10)
```
andrew-leung: Should the name be `shmSizeMB` to indicate the expected unit?

executor/mock/jobrunner.go
```diff
@@ -66,6 +67,8 @@ type JobInput struct {
 	MetatronEnabled bool
 	// Mem sets the memory resource attribute in MiB
 	Mem *int64
+	// ShmSize sets the shared memory size of `/dev/shm` in MiB
+	ShmSize *int64
```
sargun: why int64?

rgulewich: Was trying to be consistent with docker's ShmSize - though an int probably makes more sense.

executor/mock/standalone/standalone_test.go
rgulewich: This isn't relevant now that we're relying on the control plane to enforce this, but this is allowed.  The container will come up with a `/dev/shm` of whatever size you request.  However, you can only write to `/dev/shm` up to the cgroup memory limit of the container.

PR #257, opened 2019-05-27
Configure GPU containers with the nvidia-container-runtime hook

rgulewich:
- Use the [nvidia-container-runtime](https://github.com/NVIDIA/nvidia-container-runtime) instead of nvidia-docker 1.0 to configure containers
- This relies on using a non-default OCI runtime, as per [the dockerd docs](https://docs.docker.com/engine/reference/commandline/dockerd/#docker-runtime-execution-options)
- Default to a runtime named `oci-add-hooks`, eg: https://github.com/Netflix-Skunkworks/oci-add-hooks - this is preferable to the nvidia-docker2 runtime, as it doesn't rely on patching runc.
- Allow an alternate runtime to be configured via a config option
- The executor only sets the visible devices in the container - it leaves all other Nvidia config environment variables at their defaults, which matches the nvidia-docker 1.0 behaviour. More details can be found in the [nvidia-container-runtime README](https://github.com/NVIDIA/nvidia-container-runtime#environment-variables-oci-spec)
- Run standalone tests serially when the `DISABLE_PARALLEL_TESTS` env var is set

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,8 +79,8 @@ Requires docker:
 ```sh-session
 make test-standalone
 
-# Change the test timeout (useful on slower systems):
-TEST_TIMEOUT=10m make test-standalone
+# Disable running tests in parallel and change the test timeout (useful on slower systems):
+TEST_FLAGS="-v -parallel 1" TEST_TIMEOUT=10m make test-standalone
 # If you're iterating on tests and don't want to build a new executor .deb every time:
 ./hack/tests-with-dind.sh
 ```
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ Tests will run inside a Docker container and run a dedicated docker daemon as do
 
 AWS specific features (VPC integration, metadata service proxy, GPU, EFS, ...) are disabled during these tests.
 
-## Generated Code 
+## Generated Code
 There are places inside of the executor where we've checked in binaries, or prebuilt pieces of code. This may be considered harmful by some, but from the perspective of pragmatism, we have such code in:
 
 In order to generate code you need `go-bindata`, which you can get via:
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -79,8 +79,8 @@ Requires docker:
 ```sh-session
 make test-standalone
 
-# Change the test timeout (useful on slower systems):
-TEST_TIMEOUT=10m make test-standalone
+# Disable running tests in parallel and change the test timeout (useful on slower systems):
+DISABLE_PARALLEL_TESTS=true TEST_TIMEOUT=10m make test-standalone
```
sargun: Rather than DISABLE_PARALLEL_TEST, can we expose setting the parallelism? Then, if you want to disable it, you can set it to 1?

nvidia/nvidia.go
```diff
@@ -1,92 +1,46 @@
...
 
 const (
-	nvidiaPluginURL           = "http://localhost:3476"
-	nvidiaPluginDockerJSONURI = "/docker/cli/json"
-	nvidiaPluginInfoJSONURI   = "/v1.0/gpu/info/json"
-	nvidiaPluginTimeout       = time.Minute
-	nvidiaPluginName          = "nvidia-docker"
-	stateDir                  = "/run/titus-executor-nvidia"
+	stateDir     = "/run/titus-executor-nvidia"
+	nvidiaSmiExe = "/usr/bin/nvidia-smi"
```
sargun: can we use exec.lookpath here?

nvidia/nvidia.go
```diff
@@ -1,92 +1,46 @@
...
 )
 
 const (
-	nvidiaPluginURL           = "http://localhost:3476"
-	nvidiaPluginDockerJSONURI = "/docker/cli/json"
-	nvidiaPluginInfoJSONURI   = "/v1.0/gpu/info/json"
-	nvidiaPluginTimeout       = time.Minute
-	nvidiaPluginName          = "nvidia-docker"
-	stateDir                  = "/run/titus-executor-nvidia"
+	stateDir     = "/run/titus-executor-nvidia"
```
sargun: in the future, we should probably !hardcode this directory?

nvidia/nvidia.go
```diff
@@ -118,159 +72,31 @@ func (n *PluginInfo) initHostGpuInfo() error {
...
-	volumesToCreate := make(map[string]struct{})
-	for _, vol := range n.Volumes {
-		volumesToCreate[strings.Split(vol, ":")[0]] = struct{}{}
-	}
-
-	// Fetch the volumes that Docker knows about
-	args := filters.NewArgs()
-	args.Add("driver", volumeDriver)
-	volumes, err := n.dockerClient.VolumeList(context.TODO(), args)
+	cmd := exec.Command(nvidiaSmiExe, "--query-gpu=gpu_uuid", "--format=csv,noheader")
```
sargun: It'd be nice if this was run with a timeout (with context?)

nvidia/nvidia.go
```diff
@@ -118,159 +72,31 @@ func (n *PluginInfo) initHostGpuInfo() error {
...
 	}
 
-	// Determine the volumes that need to be created
-	for _, vol := range volumes.Volumes {
-		if vol.Driver == volumeDriver {
-			delete(volumesToCreate, vol.Name)
-		}
+	n.gpuIds = strings.Split(strings.TrimSpace(stdout.String()), "\n")
+	if len(n.gpuIds) == 0 {
+		log.Errorf("nvidia-smi returned 0 devices: stdout='%s', stderr='%s'", stdout.String(), stderr.String())
```
sargun: I think the `%q` formatter does this.

nvidia/nvidia.go
```diff
@@ -118,159 +72,31 @@ func (n *PluginInfo) initHostGpuInfo() error {
...
+		return errors.New("no GPU devices found")
```
sargun: Define this error as an error type.

PR #258, opened 2019-05-31
Update debian dependencies to cleanly install on Bionic

rgulewich:
- Don't hardcode the xenial apparmor version
- The postinstall explicitly restarts the lxc service, so make it an explicit dependency

### DIFF
--- a/hack/builder/titus-executor-builder.sh
+++ b/hack/builder/titus-executor-builder.sh
@@ -77,7 +77,6 @@ systemctl reload apparmor || echo "Could not reload apparmor"
 EOF
 chmod +x /tmp/post-install.sh
 
-
 fpm -t deb -s dir -C root \
   -a amd64 \
   -n titus-executor \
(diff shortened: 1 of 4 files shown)


hack/builder/titus-executor-builder.sh
```diff
@@ -99,9 +98,9 @@ fpm -t deb -s dir -C root \
   --deb-field "Branch: ${git_sha}" \
   --deb-activate ldconfig \
   --depends libc6 \
-  --depends 'apparmor >= 2.12-4ubuntu7' \
+  --depends 'apparmor >= 2.12' \
   --depends 'docker-ce >= 5:18.09.1~3-0~ubuntu-xenial' \
-  --deb-recommends lxcfs \
+  --depends lxcfs \
```
sargun: Do we want to make lxcfs a hard dep? IIRC, we can't (or don't?) use lxcfs while running in the test docker container on build kite.

rgulewich: Hmm, ok - I can back this out and install it explicitly instead.

PR #262, opened 2019-06-19
Fix system service init commands not working for non-runc runtimes

rgulewich:
The root path for a runtime changes based on its name, which was causing metatron init to fail on GPU-enabled containers (as they are now run with the `oci-add-hooks` runtime).

### DIFF
--- a/executor/runtime/docker/docker_linux.go
+++ b/executor/runtime/docker/docker_linux.go
@@ -43,19 +43,20 @@ type schedParam struct {
 type serviceEnabledFunc func(cfg *config.Config, c *runtimeTypes.Container) bool
 
 type serviceOpts struct {
-	humanName       string
-	initCommand     string
-	initCommandArgs string
-	required        bool
-	unitName        string
-	enabledCheck    serviceEnabledFunc
+	humanName    string
+	initCommand  string
+	required     bool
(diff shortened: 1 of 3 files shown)


executor/runtime/docker/docker.go
```diff
@@ -216,6 +216,7 @@ type DockerRuntime struct { // nolint: golint
 	pidCgroupPath     string
 	cfg               config.Config
 	dockerCfg         Config
+	hostCfg           *container.HostConfig
```
sargun: Since we only use our runtimes "once", does it make sense to change the API for the runtimes to look closer to that?

rgulewich: I'm not sure what you mean - are you saying store the runtime here instead of HostConfig, or something else?

rgulewich: OK - doing this breaks the contract for `type Runtime interface`: previously, you could (in theory), call `Prepare()` and `Start()` for more than one container.  Adding this dependency breaks that contract.

I changed the code to add the runtime to the `Container` struct instead, which preserves this behaviour - we might not use it currently, but I don't have a pressing need to break this contract right now, so I might as well keep it as is.

PR #268, opened 2019-06-26
Docker image metadata in the container ENV

fabiokung:
Some users report it would be useful to have introspection of what Docker image was used to launch their containers from inside the container, so it can for example be appended to logs.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -219,6 +219,16 @@ func (c *Config) getEnvBasedOnTask(taskInfo *titus.ContainerInfo, mem, cpu, disk
 	env1["TITUS_NUM_DISK"] = disk
 	env1["TITUS_NUM_NETWORK_BANDWIDTH"] = networkBandwidth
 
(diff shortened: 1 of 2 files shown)


sargun:
@fabiokung Are these size caps on any of these?

fabiokung:
Good point. I don't think we explicitly validate it anywhere in the control plane, but it seems docker itself will enforce some limits, so people can't just use arbitrarily big names/tags (digests are a sha256).

Quick test with a local docker daemon here gives me (moby/moby#11118):

```sh-session
$ docker tag ubuntu too-long-...
Error parsing reference: "too-long-..." is not a valid repository/tag: repository name must not be more than 255 characters
```

And: https://docs.docker.com/engine/reference/commandline/tag/

> A tag name must be valid ASCII and may contain lowercase and uppercase letters, digits, underscores, periods and dashes. A tag name may not start with a period or a dash and may contain a maximum of 128 characters.

PR #269, opened 2019-06-27
Metadata service: refresh Metatron certs every 5 minutes

rgulewich:
This prevents the task identity document being signed with a stale certificate.

### DIFF
--- a/cmd/titus-metadata-service/main.go
+++ b/cmd/titus-metadata-service/main.go
@@ -30,6 +30,8 @@ import (
 // 169 is the first octet of 169.254...
 const defaultListeningPort = 8169
 
+const certRefreshTime = 5 * time.Minute
+
 /* Either returns a listener, or logs a fatal error */
 func getListener(listenPort int, listenerFd int64) net.Listener {
 	if listenerFd != -1 && listenPort != defaultListeningPort {
@@ -76,6 +78,20 @@ func readTaskConfigFile(taskID string) (*titus.ContainerInfo, error) {
 	}
 
 	return &cInfo, nil
+}
+
+func reloadSigner(ms *metadataserver.MetadataServer) {
+	t := time.NewTicker(certRefreshTime)
+	defer t.Stop()
+	for range t.C {
+		if newSigner, err := identity.NewDefaultSigner(); err != nil {
+			log.WithError(err).Fatal("Cannot instantiate new default signer")
+		} else {
+			if err := ms.SetSigner(newSigner); err != nil {
+				log.WithError(err).Error("Error reloading signing certificate")
+			}
+		}
+	}
 }
 
 func main() {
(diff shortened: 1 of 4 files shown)


sargun:
Can you wire up SIGHUP as well?

cmd/titus-metadata-service/main.go
```diff
@@ -78,6 +80,17 @@ func readTaskConfigFile(taskID string) (*titus.ContainerInfo, error) {
 	return &cInfo, nil
 }
 
+func reloadSigner(ms *metadataserver.MetadataServer) {
+	for {
+		time.Sleep(certRefreshTime)
```
sargun:
```
t := time.NewTicker(5 * time.Second)
defer t.Stop()
for range <-t.C {
...
}
```

metadataserver/identity.go
```diff
@@ -29,6 +31,39 @@ func (ms *MetadataServer) generateTaskIdentity() *titus.TaskIdentity {
...
+			"newIssuer":       newCert.Issuer,
+			"newNotBefore":    newCert.NotBefore,
+			"newNotAfter":     newCert.NotAfter,
+			"newSerialNumber": newCert.SerialNumber,
+			"oldSubject":      oldCert.Subject,
+			"oldIssuer":       oldCert.Issuer,
+			"oldNotBefore":    oldCert.NotBefore,
+			"oldNotAfter":     oldCert.NotAfter,
+			"oldSerialNumber": oldCert.SerialNumber,
+		}).Infof("signer cert updated")
```
sargun: why infof?

metadataserver/identity.go
```diff
@@ -29,6 +31,39 @@ func (ms *MetadataServer) generateTaskIdentity() *titus.TaskIdentity {
...
+	}
+
+	ms.signLock.Lock()
+	oldSigner := ms.signer
+	oldCert, err := x509.ParseCertificate(oldSigner.Certificate.Certificate[0])
+	if err != nil {
+		log.WithError(err).Error("Error parsing old certificate")
+	}
+
+	if newCert != nil && oldCert != nil {
```
sargun: why not just do a blind 'return' in the `if err != nil` sections? or `x509.ParseCertificate` return errors?

cmd/titus-metadata-service/main.go
rgulewich: Cool, didn't know about `NewTicker()`.

metadataserver/identity.go
rgulewich: I got a little nervous about not doing the assignment when those parses fail.  I guess it's fine, though.

rgulewich:
> Can you wire up SIGHUP as well?

Why, if restarting the systemd unit does the same thing?

sargun:
Because my injection code is jank, and it fails sometimes. Systemd eventually retries restarts and it works again. But, again, jank.

rgulewich:
@sargun - OK, added.

cmd/titus-metadata-service/main.go
```diff
@@ -78,6 +81,29 @@ func readTaskConfigFile(taskID string) (*titus.ContainerInfo, error) {
...
+
+	for {
+		select {
+		case <-t.C:
+		case <-sigs:
+		}
+
+		if newSigner, err := identity.NewDefaultSigner(); err != nil {
+			log.WithError(err).Fatal("Cannot instantiate new default signer")
+		} else {
```
sargun: Drop the else. The above statement is terminating.

PR #272, opened 2019-07-03
Rename darion

rgulewich:
Rename titus-darion to titus-logviewer


sargun:
This means you own it now.

PR #274, opened 2019-07-04
System services should use the new runc path for newer containerd releases

rgulewich:
Also account for a non-runc runtime when running the service

### DIFF
--- a/root/lib/systemd/system/titus-metatron-sync@.service
+++ b/root/lib/systemd/system/titus-metatron-sync@.service
@@ -4,7 +4,7 @@ ConditionPathIsDirectory=/var/lib/titus-inits/%i/ns
 
 [Service]
 EnvironmentFile=/var/lib/titus-environments/%i.env
-ExecStart=/usr/sbin/runc --root /var/run/docker/runtime-runc/moby exec --user 0:0 --cap CAP_DAC_OVERRIDE ${TITUS_CONTAINER_ID} /titus/metatron/bin/titus-metatrond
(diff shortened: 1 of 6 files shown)


root/lib/systemd/system/titus-metatron-sync@.service
```diff
@@ -4,7 +4,7 @@ ConditionPathIsDirectory=/var/lib/titus-inits/%i/ns
 
 [Service]
 EnvironmentFile=/var/lib/titus-environments/%i.env
-ExecStart=/usr/sbin/runc --root /var/run/docker/runtime-runc/moby exec --user 0:0 --cap CAP_DAC_OVERRIDE ${TITUS_CONTAINER_ID} /titus/metatron/bin/titus-metatrond
+ExecStart=/usr/bin/runc --root /var/run/docker/runtime-${TITUS_OCI_RUNTIME}/moby exec --user 0:0 --cap CAP_DAC_OVERRIDE ${TITUS_CONTAINER_ID} /titus/metatron/bin/titus-metatrond
```
sargun: Does this unit now break on xenial?

rgulewich: It does.  Not really sure what to do about that, other than shipping a wrapper script that tries each in order.

rgulewich: Or just symlink `/usr/sbin/runc` -> `/usr/bin/runc` on Bionic, I suppose.

PR #278, opened 2019-07-12
Run a logviewer per container as a system service

rgulewich:
- Package the logviewer in the `titusoss/titus-logviewer` image
- Enable with `CONTAINER_LOGVIEWER` and `CONTAINER_LOGVIEWER_IMAGE`
- Make the logviewer detect when it's running in a Titus container and adjust log paths accordingly


executor/mock/jobrunner.go
```diff
@@ -252,7 +282,9 @@ func (jobRunner *JobRunner) StartJob(jobInput *JobInput) *JobRunResponse { // no
 	}
 
 	// TODO: refactor this all to use NewContainer()
-	taskID := fmt.Sprintf("Titus-%v%v-Worker-0-2", r.Intn(1000), time.Now().Second())
+	validContainerNameRE := regexp.MustCompile("[^a-zA-Z0-9_.-]")
```
sargun: Wait, why do we do this replacement?

hack/agent/Dockerfile
```diff
@@ -46,7 +49,7 @@ COPY hack/agent/daemon.json /etc/docker/daemon.json
 COPY hack/agent/titus-shared.env /etc/titus-shared.env
 COPY hack/agent/10-disable-atlas-agent.conf /etc/systemd/system/atlas-titus-agent@.service.d/10-disable-atlas-agent.conf
 
-RUN wget -O /tmp/apparmor.deb http://launchpadlibrarian.net/381273068/apparmor_2.12-4ubuntu7_amd64.deb
+RUN curl http://launchpadlibrarian.net/381273068/apparmor_2.12-4ubuntu7_amd64.deb > /tmp/apparmor.deb
```
sargun: Do we still need to do this under bionic?

executor/mock/jobrunner.go
rgulewich: This strips out characters that docker won't allow in container names.  Will add a comment.

hack/agent/Dockerfile
rgulewich: Probably not - I'll try removing.

PR #281, opened 2019-07-15
Add hedging to AWS requests

sargun:


rgulewich:
This looks more like retry logic with a backoff to me - is there no retry logic in the AWS client that we can use?

sargun:
@rgulewich It's hedging + retry logic kind of. Basically, the second, and third requests wont start except for some period after the first started. We already have retries. Retries work slightly differently, in that they happen after a call has failed. The AWS client already has backoff + retries.

sargun:
@rgulewich updated.

rgulewich:
@sargun Great - thanks for the detailed comment.  I probably won't be the last person with questions... :)

PR #285, opened 2019-08-05
Add network metadata to pod update

gabrielhartmann:
The control plane wants this network metadata information which was previously provided by the Mesos backend.

### DIFF
--- a/executor/runtime/types/types.go
+++ b/executor/runtime/types/types.go
@@ -486,6 +486,17 @@ type NetworkConfigurationDetails struct {
 	ResourceID   string
 }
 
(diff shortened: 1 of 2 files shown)


executor/runtime/types/types.go
```diff
@@ -486,6 +486,17 @@ type NetworkConfigurationDetails struct {
 	ResourceID   string
 }
 
+func (n *NetworkConfigurationDetails) ToMap() map[string]string {
+	m := make(map[string]string)
```
sargun: btw, you know you can just:

return map[string]string{
"x": "y",
...
}

PR #286, opened 2019-08-07
Add retries to titus isolate connection on isolation time

sargun:


sargun:
codecov seems down right now, which is problematic.

executor/runtime/docker/docker_isolate.go
```diff
@@ -12,15 +12,43 @@ import (
...
+			}
+		case <-firstTry:
+			if workloadIsolated(ctx, taskID, host) {
+				return true
+			}
+		}
+	}
+}
+
+func workloadIsolated(ctx context.Context, taskID, host string) bool {
```
rgulewich: Minor nitpick: could you rename to something like `isolateWorkload()` (or the like)?  `workloadIsolated` implies that you're checking that the workload was isolated at some time in the past, rather than actively telling titus-isolate to do the isolation.

sargun: We aren't telling the titus-isolate to do the isolation. I added a comment . accordingly.

PR #288, opened 2019-08-21
Static address assignment

sargun:
### Description of the Change


sargun:
@rgulewich this is a giant PR. It'd be nice to give it a high-level skim, but since it's new functionality, that doesn't effect the current hot path, I think it's okay.

vpc/service/allocate.go
```diff
@@ -0,0 +1,379 @@
...
+	ip := net.ParseIP(aws.StringValue(assignPrivateIPAddressesOutput.AssignedPrivateIpAddresses[0].PrivateIpAddress))
+	az := aws.StringValue(dummyNetworkInterface.AvailabilityZone)
+
+	ec2session := ec2.New(awsSession)
+	_, err = ec2session.CreateTagsWithContext(ctx, &ec2.CreateTagsInput{
+		DryRun:    nil,
+		Resources: []*string{dummyNetworkInterface.NetworkInterfaceId},
+		Tags: []*ec2.Tag{
+			{
+				Key:   aws.String(fmt.Sprintf("vpc-service:%s", allocationUUID)),
```
rgulewich: Don't know if there's a limit on the length of keys, but it might be worth calling this `titus-vpc-service:%s`, so that people poking around AWS know who to ask about these ENIs.

PR #326, opened 2019-12-06
Add KVM and tunnel devices to the container when launching a VM

gabrielhartmann:
In order to launch virtual machines in a Titus container the devices `/dev/kvm` and `/dev/net/tun` must be added.

### DIFF
--- a/executor/runtime/docker/capabilities.go
+++ b/executor/runtime/docker/capabilities.go
@@ -11,7 +11,11 @@ import (
 
 const (
 	SYS_ADMIN = "SYS_ADMIN" // nolint: golint
-	NET_ADMIN = "NET_ADMIN"
+	NET_ADMIN = "NET_ADMIN" // nolint: golint
+)
+
+var (
+	errNestedContainers = errors.New("nested containers no longer supported")
 )
 
 func addAdditionalCapabilities(c *runtimeTypes.Container, hostCfg *container.HostConfig) map[string]struct{} {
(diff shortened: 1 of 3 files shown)


executor/runtime/docker/capabilities.go
```diff
@@ -30,36 +30,49 @@ func addAdditionalCapabilities(c *runtimeTypes.Container, hostCfg *container.Hos
 }
 
 func setupAdditionalCapabilities(c *runtimeTypes.Container, hostCfg *container.HostConfig) error {
-	addedCapabilities := addAdditionalCapabilities(c, hostCfg)
+	if c.TitusInfo.GetAllowNestedContainers() {
+		return errors.New("nested containers no longer supported")
```
sargun: If you want this can be defined as a var

executor/runtime/docker/capabilities.go
```diff
@@ -30,36 +30,49 @@ func addAdditionalCapabilities(c *runtimeTypes.Container, hostCfg *container.Hos
...
+	}
 
-	// Privileged containers automaticaly deactivate seccomp and friends, no need to do this
 	fuseEnabled, err := c.GetFuseEnabled()
+	if err != nil {
+		return err
+	}
 
+	kvmEnabled, err := c.GetKvmEnabled()
```
sargun: Do KVM containers need CAP_SYS_ADMIN?

gabrielhartmann: no, not to launch the VM anyway post network configuration

sargun: I think for the network config, all you need is CAP_NET_ADMIN.

gabrielhartmann: Yes

PR #328, opened 2019-12-17
Add device mapping for root EBS volume

gabrielhartmann:
Allow container access to the attached EBS root volume.

### DIFF
--- a/executor/runtime/docker/capabilities.go
+++ b/executor/runtime/docker/capabilities.go
@@ -3,6 +3,7 @@ package docker
 import (
 	"errors"
 	"fmt"
+	"os"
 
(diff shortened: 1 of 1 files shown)


executor/runtime/docker/capabilities.go
```diff
@@ -83,6 +84,16 @@ func setupAdditionalCapabilities(c *runtimeTypes.Container, hostCfg *container.H
 			CgroupPermissions: "rmw",
 		})
 
+		value, exists := os.LookupEnv("ROOT_DEVICE_PATH")
```
sargun: What will set this environment variable?

gabrielhartmann: VK

PR #330, opened 2019-12-19
This adds server-side SSL to the Titus VPC service

sargun:


vpc/service/auth.go
```diff
@@ -0,0 +1,40 @@
...
+	if peer.AuthInfo != nil {
+		l.WithField("authtype", peer.AuthInfo.AuthType()).Debug("Authenticating peers via AuthFuncOverride")
+	} else {
+		l.Debug("not authenticating peers via AuthFuncOverride")
+
+	}
+	return ctx, nil
+}
+
+func (*vpcService) AuthFuncOverride(ctx context.Context, fullMethodName string) (context.Context, error) {
```
rgulewich: What's this function for, and where is it used?

PR #331, opened 2019-12-27
Introduce ssl auth

sargun:
This introduces SSL authentication for the VPC Service. Specifically, the authFunc only is attached to the VPC Service functions, as opposed to the service as a whole.

### DIFF
--- a/cmd/titus-vpc-tool/shared.go
+++ b/cmd/titus-vpc-tool/shared.go
@@ -2,11 +2,16 @@ package main
 
 import (
 	"context"
+	"crypto/tls"
+	x509 "crypto/x509"
+	"fmt"
+	"io/ioutil"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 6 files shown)


cmd/titus-vpc-tool/shared.go
```diff
@@ -37,9 +78,13 @@ func getSharedValues(ctx context.Context, v *pkgviper.Viper) (*fslocker.FSLocker
 
 	entry.WithField("serviceAddr", serviceAddr).Debug("Initializing client")
 	grpc_logrus.ReplaceGrpcLogger(entry)
+	securityDialOption, err := getSecurityConfiguration(ctx, v)
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "Cannot configure ")
```
rgulewich: Might want to mention what we can't configure.

PR #339, opened 2020-02-14
This commit allows customization of log location.

alekar:
S3 buckets and path prefixes can be customized using pass through
parameters. This commit adds two new passthrough parameters:

1. titusParameter.agent.log.s3BucketName
2. titusParameter.agent.log.s3PathPrefix

The s3 uploading backend has also been modified to assume the IAM
credentials of the running container. The user needs to make sure
that the specified log location is writable by the role.

The s3 uploadmanager shares the role session name with IAM system
service.

To allow this configuration, the initialization of the logger has
been delayed till the titus task configuration is available. This
is done by creating the loggers during container startup in pkg
runtime.

Finally, it is no longer possible to specify multiple uploaders in
configuration. The initializer prefers s3 destination, followed by
a local destination. If none are specified, logs will be lost.


uploader/uploader.go
```diff
@@ -1,149 +1,127 @@
...
+
+				wg.Add(1)
+				go func(fi os.FileInfo) {
+					qlocal := path.Join(local, fi.Name())
+					qremote := path.Join(remote, fi.Name())
+
+					log.WithField("local", qlocal).WithField("remote", qremote).Info("uploading")
+					if err = uploader.Upload(ctx, qlocal, qremote, ctypeFunc); err != nil {
+						log.WithField("local", qlocal).WithField("remote", qremote).Error(err)
+						errs = multierror.Append(errs, err)
```
rgulewich: Is this thread-safe?  Can multiple goroutines call this at the same time without losing data?

uploader/uploader.go
```diff
@@ -1,149 +1,127 @@
...
-	for range cfg.NoopUploaders {
-		copyUploaders = append(copyUploaders, NewNoopUploader())
+	if len(config.S3Uploaders) > 0 {
+		bucketName = config.S3Uploaders[0]
 	}
 
-	return copyUploaders
-}
+	if param, ok := titusInfo.GetPassthroughAttributes()[s3BucketNameParam]; ok {
+		bucketName = param
```
rgulewich: What's the behavior if we can't write to this S3 bucket?  Do we fall back to the default?

alekar: We drop the logs. The concern here is we write some sensitive information to a default bucket which does not have the right permissions.

uploader/uploader.go
alekar: Fixed via 3949578.

filesystems/watcher.go
```diff
@@ -18,9 +18,10 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/Netflix/titus-executor/uploader"
```
sargun: Formatting did something weird here?

uploader/uploader.go
```diff
@@ -1,149 +1,137 @@
...
+
+			// Iterate over each file and upload it
+			var wg sync.WaitGroup
+
+			for i, finfo := range finfos {
+				if finfo.IsDir() {
+					continue // don't upload subdirs
+				}
+
+				wg.Add(1)
```
sargun: I don't know if it's legal for multiple goroutines to simultaneously write to the same slice. I think you have to use a channel to collect the errors.

uploader/uploader.go
sargun: You might want to try: https://github.com/Netflix/titus-executor/blob/master/vpc/service/errgroupish.go

PR #347, opened 2020-03-02
CNI Plugin [WIP]

sargun:
Add CNI Plugin


sargun:
There's a bunch of TODOs here. And there's no background GC process, but it's a start in the right direction, IMHO.

If you want to try it out, drop a file in `/etc/cni/net.d/10-sargun.conf` (make sure there's no lower files)

with the contents:
```
{
  "cniVersion": "0.3.0",
  "type": "titus",
  "name": "sargun",
  "KubeletAPIURL": "https://localhost:10250/pods"
}
```

(I have absolutely no idea what the name does)

sargun:
@hankjacobs @rgulewich I'm also thinking of splitting out the bits under vpc/service/ into a separate PR. What do you think?

sargun:
Also, I'm sorry. This wasn't supposed to be 1000 lines. I have no idea how it became 1 thousand lines. :(.

hankjacobs:
+1 on splitting the vpc/service stuff into a separate PR. Will make it easier to review

sargun:
@hankjacobs Already split: https://github.com/Netflix/titus-executor/pull/362

PR #351, opened 2020-03-07
Long lived locks GRPC API

hankjacobs:
This PR adds a GRPC API for long lived locks.


sargun:
This is awesome. I'll take a look shortly.

sargun:
You may want to rebase on top of master. I noticed you also had to re-vendor our PB files in this PR. Master has that fixed.

vpc/proto/vpc.proto
```diff
@@ -262,6 +277,10 @@ service TitusAgentVPCService {
     rpc GCV2 (GCRequestV2) returns (GCResponseV2);
     rpc GCV3 (GCRequestV3) returns (GCResponseV3);
 
+    // Lock Management
```
sargun: In general the pattern has been something like:

```
message GetLocksRequest {
}
```

So, if we wanted to add filtering (one day?) we could.

vpc/proto/vpc.proto
```diff
@@ -245,6 +245,21 @@ message GCLegacySetupResponse {
     repeated NetworkInterfaceAttachment networkInterfaceAttachment = 1;
 }
 
+message LockId {
+    int64 id = 1;
+}
+
+message Lock {
+    int64 id = 1;
+    string lockName = 2;
```
sargun: This makes me thing that longer-term we should probably put the keyedItem + lock worker name into the database just as useful metadata.

vpc/service/long_lived_lock.go
```diff
@@ -19,6 +24,81 @@ const (
...
+			return nil, err
+		}
+
+		locks = append(locks, &lock)
+	}
+
+	return locks, nil
+}
+
+func (vpcService *vpcService) GetLocks(ctx context.Context, empty *empty.Empty) (*vpcapi.GetLocksResponse, error) {
```
sargun: Please start / stop a tracing span here, and use tracehelpers to set an error code on the span appropriately.

vpc/service/long_lived_lock.go
```diff
@@ -19,6 +24,81 @@ const (
...
+	rows, err := vpcService.db.QueryContext(ctx, "SELECT id, lock_name, held_by, held_until FROM long_lived_locks LIMIT 1000")
```
sargun: Please wrap this in (read only) transaction

vpc/service/long_lived_lock.go
```diff
@@ -19,6 +24,81 @@ const (
 	timeBetweenTryingToAcquireLocks = 15 * time.Second
 )
 
+func (vpcService *vpcService) readLocks(rows *sql.Rows) ([]*vpcapi.Lock, error) {
```
sargun: I don't think there's any reason for this function to have a pointer receiver?

vpc/service/long_lived_lock.go
```diff
@@ -19,6 +24,81 @@ const (
...
+
+	return locks, nil
+}
+
+func (vpcService *vpcService) GetLocks(ctx context.Context, empty *empty.Empty) (*vpcapi.GetLocksResponse, error) {
+	rows, err := vpcService.db.QueryContext(ctx, "SELECT id, lock_name, held_by, held_until FROM long_lived_locks LIMIT 1000")
+	if err != nil {
+		return nil, err
+	}
+	defer rows.Close()
```
sargun: I should probably be doing this everywhere. Good call.

vpc/service/long_lived_lock.go
```diff
@@ -19,6 +24,81 @@ const (
...
+
+	locks, err := vpcService.readLocks(rows)
+	if err != nil {
+		return nil, err
+	}
+
+	return &vpcapi.GetLocksResponse{Locks: locks}, nil
+}
+
+func (vpcService *vpcService) GetLock(ctx context.Context, id *vpcapi.LockId) (*vpcapi.Lock, error) {
```
sargun: See above about the span, transaction, and such.

vpc/service/long_lived_lock.go
```diff
@@ -19,6 +24,81 @@ const (
...
+	rows, err := vpcService.db.QueryContext(ctx, "SELECT id, lock_name, held_by, held_until FROM long_lived_locks WHERE id = $1", id.GetId())
```
sargun: I think you can simplify this to QueryRowContext (because this could return a maximum of one row).

And then if err == sql.ErrNoRows, you can go directly to the lock not found error. 

This may result in a *slight* bit of code duplication, if you don't break out readLocks -> readLock

sargun: You get this guarantee from the db because id is a PRIMARY / unique key:

```
titusvpcservice=> \d long_lived_locks
;                                        Table "public.long_lived_locks"
   Column   |            Type             | Collation | Nullable |                   Default                    
------------+-----------------------------+-----------+----------+----------------------------------------------
 id         | bigint                      |           | not null | nextval('long_lived_locks_id_seq'::regclass)
 lock_name  | text                        |           |          | 
 held_by    | text                        |           |          | 
 held_until | timestamp without time zone |           |          | 
Indexes:
    "long_lived_locks_pk" PRIMARY KEY, btree (id)
    "long_lived_locks_lock_name_uindex" UNIQUE, btree (lock_name)

```

vpc/service/long_lived_lock.go
```diff
@@ -19,6 +24,81 @@ const (
...
+	}
+
+	if len(locks) == 0 {
+		return nil, status.Error(codes.NotFound, "lock not found")
+	}
+
+	return locks[0], nil
+}
+
+func (vpcService *vpcService) DeleteLock(ctx context.Context, id *vpcapi.LockId) (*empty.Empty, error) {
```
sargun: So, is the intent behind this function to "unlock" the lock? It might make sense to build an actual mechanism to do that. 

Right now, the way you "knock" a lock holder out is by setting held_by to null, because that'll cause this to fail:

```
	result, err := tx.ExecContext(ctx, "UPDATE long_lived_locks SET held_until = now() + ($1 * interval '1 sec') WHERE id = $2 AND held_by = $3", lockTime.Seconds(), id, hostname)
	if err != nil {
		err = errors.Wrap(err, "Could update lock time")
		span.SetStatus(traceStatusFromError(err))
		return err
	}
	n, err := result.RowsAffected()
	if err != nil {
		err = errors.Wrap(err, "Could not get rows affected")
		span.SetStatus(traceStatusFromError(err))
		return err
	}
```

vpc/service/long_lived_lock.go
```diff
@@ -19,6 +24,81 @@ const (
...
+	affected, err := res.RowsAffected()
+	if err != nil {
+		return nil, err
+	}
+
+	if affected == 0 {
+		return nil, status.Error(codes.NotFound, "lock not found")
+	}
+
+	return &empty.Empty{}, nil
```
sargun: It might be neat to return who it was held by before?

vpc/service/long_lived_lock_test.go
```diff
@@ -0,0 +1,168 @@
+package service
```
sargun: This is awesome.

vpc/service/long_lived_lock_test.go
```diff
@@ -0,0 +1,168 @@
...
+
+	expected, rows := generateLockAndRows(t, mock)
+	mock.ExpectQuery("SELECT id, lock_name, held_by, held_until FROM long_lived_locks LIMIT 1000").WillReturnRows(rows)
+
+	service := vpcService{db: db}
+
+	ctx := context.Background()
+	res, err := service.GetLocks(ctx, &empty.Empty{})
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
```
sargun: One of the tools we use is gotools assert: https://pkg.go.dev/gotest.tools/v3@v3.0.2/assert?tab=doc -- it makes this slightly cleaner, if you want. It also has a package (cmp), that allows you to write things like:

`asssert.Assert(t, cmp.Len(got, 1))`

or

`assert.Assert(t, proto.Equal(expected, got))`

vpc/proto/vpc.proto
hankjacobs: Yep, good call. I was following the style of titus-api-definitions but I absolutely see your point.

vpc/service/long_lived_lock.go
hankjacobs: No, not technically. However, it is [generally best practice ](https://golang.org/doc/faq#methods_on_values_or_pointers) to use the same receiver type, e.g. if one method is a pointer receiver, all should be. This is so that the [method set](https://golang.org/doc/faq#different_method_sets) is consistent.

vpc/service/long_lived_lock_test.go
hankjacobs: Sure. I tend to be a minimalist when it comes to packages but I have no qualm using it so, at the very least, it's consistent .

vpc/proto/vpc.proto
hankjacobs: Hm, looking back at the titus-api-definitions I don't see what I was thinking. Anyway, I'll get that change in.

vpc/service/long_lived_lock.go
hankjacobs: Yeah, I got lazy and wanted to reuse some logic but I see what you mean.

vpc/service/long_lived_lock.go
hankjacobs: Done but it's unclear to me what the intention of it is. As far as I understand it, the default transaction isolation is read committed so that prevents dirty reads. I'm probably missing something though...

vpc/service/long_lived_lock_test.go
hankjacobs: thank you!

vpc/service/long_lived_lock.go
hankjacobs: Yep, totally. I added PreemptLock to do that. I was aiming for optionally allowing a lock to also be assigned to a new holder us that api but I don't think that's possible based on the acquire lock code. Please correct me if I'm wrong.

sargun:
@hankjacobs Please rebase on master. This currently breaks the buildkite agents.

hankjacobs:
done and done

sargun:
Feel free to merge.

PR #352, opened 2020-03-09
[WIP] POC of advisory lock based long lived locks

hankjacobs:
This is a working POC implementation of how long lived locks can be implemented with Postgres's session based advisory lock. It still needs to be cleaned up and what not but I'd like to decide if we want to switch to this before I do. 

Here's a runnable example of how to use it:

```
package main

import (
	"context"
	"database/sql"
	"fmt"
	"time"

	"github.com/Netflix/titus-executor/vpc/pglock"
	"github.com/lib/pq"
)

func main() {
	connector, err := pq.NewConnector("postgres://postgres:@localhost/titusvpcservice?sslmode=disable")
	checkError(err)
	db := sql.OpenDB(connector)

	lock := pglock.New("hank11", "computer1", db)

	fmt.Println("attempting lock")
	err = lock.Lock(context.Background())
	checkError(err)

	fmt.Println("locked")
	done := make(chan interface{})
	go func() {
		held := true
		for held {
			held, err = lock.IsHeld(context.Background())
			time.Sleep(1 * time.Second)
		}

		fmt.Println("closed")
		close(done)
	}()

	time.Sleep(30 * time.Second)
	fmt.Println("attempting unlock")
	err = lock.Unlock(context.Background())
	checkError(err)

	fmt.Println("unlocked")
	_ = <-done
	fmt.Println("exit")
}

func checkError(err error) {
	if err != nil {
		panic(err)
	}
}
```

### DIFF
--- a/vpc/pglock/pglock.go
+++ b/vpc/pglock/pglock.go
@@ -1,0 +1,126 @@
+package pglock
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+	"sync"
+
+	"github.com/pkg/errors"
+)
+
+// Lock lock
+type Lock struct {
+	mu     sync.RWMutex
+	held   bool
+	name   string
+	heldBy string
+	db     *sql.DB
+	conn   *sql.Conn
+	closed chan interface{}
+	done   chan interface{}
+}
+
+// New lock
+func New(name string, heldBy string, db *sql.DB) *Lock {
+	return &Lock{mu: sync.RWMutex{}, name: name, heldBy: heldBy, db: db}
+}
+
+// Lock the lock
+func (lock *Lock) Lock(ctx context.Context) error {
+	// Start critical section -- TODO: this can be done better
+	lock.mu.Lock()
+	if lock.held {
+		lock.mu.Unlock()
+		return fmt.Errorf("lock(%v) already locked", lock.name)
+	}
+	lock.held = true
+	lock.mu.Unlock()
+
+	resetHeld := func() {
+		lock.mu.Lock()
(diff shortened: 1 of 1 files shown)


sargun:
I thought I left a comment on this, but I guess I didn't. I think based on the current approach, your work around for sessions (or lack thereof is neat). 

My worry is around two things:
1. How do we preempt these locks? 
2. What happens if the network fails (especially in a way where the locking connection stays alive, but query connections fail, or vice-versa).

For some of the workers, problem #2 can be solved by making them do all their work over the session with the advisory lock. 

For preemption, I was thinking of `pg_terminate_backend`. It can terminate the connection.

hankjacobs:
For #1, yes I was thinking `pg_terminate_backend` to preempt the locks. Also, we could query the pg_lock table to see list what locks are currently held. 

For #2, the current design expects the lock holder to periodically call `IsHeld()`. The idea being the `held_by` column for the lock held will change if the caller suffered a network partition and lost the lock. `IsHeld()` will return false if that is the case. My original implementation had the lock poll itself but the api became pretty awkward and I generally try to push go routine management as far up the call stack as possible. 

All that being said, the question becomes "Is it worth pursuing further?". I do feel abstracting away the lock implementation (like is done here) will help minimize complexity and make the code easier to reason about. Do the ends justify the means?

solarkennedy:
@hankjacobs coming up on the 1 year anniversary of this PR. Is it sill relevant?

hankjacobs:
Nope. Thanks for the reminder!

PR #359, opened 2020-03-15
IMDS v2

hankjacobs:
Add support for enforcing tokens on the metadata service per [v2 of the IMDS API](https://aws.amazon.com/blogs/security/defense-in-depth-open-firewalls-reverse-proxies-ssrf-vulnerabilities-ec2-instance-metadata-service/)


hankjacobs:
Mostly done. Just need to add VCR tests.

hankjacobs:
The downside with this approach is the token validity is limited to the duration of which the metatron cert is valid and kept. I believe it’s currently only kept for 5 minutes. I may switch implementations to just keeping track of random tokens.

hankjacobs:
I also realized I still need to add support for fetching tokens when proxying requests

metadataserver/server.go
```diff
@@ -25,6 +25,7 @@ import (
 const defaultMacAddress = "00:00:00:00:00:00"
 
 var whitelist = set.NewSetFromSlice([]interface{}{
+	"/latest/api/token",
```
hankjacobs: This should be removed. This should never be proxied

hankjacobs:
Better yet, I’m going to switch to a unique key generated at startup + HMAC

.golangci.yml
```diff
@@ -16,3 +16,6 @@ run:
     - vpc/bpf/filter
     - vpc/bpf2
     - vpc/limitsold
+issues:
+  exclude:
+    - Using the variable on range scope .* in function literal
```
sargun: This has caught a number of real issue (primarily around launching goroutines in functions). I'd rather not exclude this.

cmd/titus-metadata-service/main.go
```diff
@@ -213,6 +214,11 @@ func main() {
 			EnvVar:      "IAM_STATE_DIR",
 			Destination: &stateDir,
 		},
+		cli.BoolFlag{
+			Name:        "require-token",
+			Usage:       "Set to true to require a token",
+			Destination: &requireToken,
```
sargun: Adding an EnvVar here would be nice.

metadataserver/auth/auth.go
```diff
@@ -0,0 +1,165 @@
...
+	env := envelope{Token: token, Expiration: exp}
+	envJSON, err := json.Marshal(env)
+	if err != nil {
+		return "", err
+	}
+	envEnc := base64.StdEncoding.EncodeToString(envJSON)
+	mac := a.hmac(envEnc)
+	sig := base64.StdEncoding.EncodeToString(mac)
+
+	return envEnc + "." + sig, nil
```
sargun: It looks like you've reimplemented / reinvented JWT / JWS here. Any reason not to just use off the shelf JWTs? They have expiration and stuff "built-in" to them?

sargun: (I used the first implementation of Go JWT examples I found)
https://play.golang.com/p/NhQDPykRQdY


Although your (our) certificate signer implemens a slightly different interface from the one they expect. I imagine we can map it across.

metadataserver/authentication.go
```diff
@@ -0,0 +1,78 @@
...
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		token := r.Header.Get("X-aws-ec2-metadata-token")
+		if !ms.tokenRequired && len(token) == 0 {
+			next.ServeHTTP(w, r)
+			return
+		}
+
+		auth := auth.HMACAuthenticator{Key: ms.tokenKey}
+		if !auth.VerifyToken(token) {
+			http.Error(w, "", http.StatusUnauthorized)
```
sargun: It might be worth indicating if the token is totally invalid vs. not present vs. expired

metadataserver/authentication.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	conn, bufrw, err := hj.Hijack()
+	if err != nil {
+		http.Error(w, "", http.StatusInternalServerError)
+		return
+	}
+	defer conn.Close()
+
+	if conn.RemoteAddr().(*net.TCPAddr).IP.To4() != nil {
```
sargun: We need this to work on IPv6 (in the future as well)

hankjacobs:
Well, at the beginning I wanted to avoid them because they felt too heavy but yeah, I agree, it’s already practically a JWT. Will swap it in.

metadataserver/authentication.go
hankjacobs: This is the way IMDS v2 actually behaves. I’m concern we’d break clients if we deviate.

metadataserver/auth/auth.go
hankjacobs: Yep, will use JWT

metadataserver/authentication.go
hankjacobs: Shouldn’t be a problem but I don’t think IMDS is available over v6. I could be wrong though.

sargun:
> I also realized I still need to add support for fetching tokens when proxying requests

Do you mean that the system "consume" IMDSv2 on the backend? I don't think that we need to do that.

metadataserver/authentication.go
sargun: Can we log the reason at least?

sargun: All the logs end up in journald, and we can setup alerting based on it.

sargun:
It seems like it's in the middle of moving to JWTs. Let me know when I should take another look.

hankjacobs:
JWT support is complete. The only things left are your code review changes
and updating the VCR tests. The review can totally wait until then.

hankjacobs:
What's probably a little confusing is I left the certificate implementation in there. I'll gut it.

metadataserver/authentication.go
hankjacobs: Yep Yep, good cal. I'll do a logging pass through all of the new code.

hankjacobs: Yep yep, good call. I'll do a logging pass on all of my new code. I'll take a look at adding handy metrics as well.

hankjacobs:
I've already added support for using IMDS v2 on the backend (in the proxy). It'll allow us to turn of v1 at some point in the future (I thought that was the intention and is why I implemented it).

.golangci.yml
```diff
@@ -16,3 +16,4 @@ run:
     - vpc/bpf/filter
     - vpc/bpf2
     - vpc/limitsold
+
```
sargun: Whitespace weirdness.

metadataserver/auth/auth.go
```diff
@@ -0,0 +1,132 @@
...
+
+// JWTAuthenticator authenticates using JWT
+type JWTAuthenticator struct {
+	Key []byte
+}
+
+// GenerateToken token
+func (a *JWTAuthenticator) GenerateToken(ttl time.Duration) (string, error) {
+	claims := &jwt.StandardClaims{
+		ExpiresAt: time.Now().Add(ttl).Unix(),
```
sargun: Please add the task ID to the token, or similar. I think they call this "Audience" in JWT terms. I believe partially it'll improve security, but more so, improve debugging.

metadataserver/auth/auth.go
```diff
@@ -0,0 +1,132 @@
...
+	return true, int64(remaining.Seconds())
+}
+
+// CertificateAuthenticator generates and authenticates tokens
+type CertificateAuthenticator struct {
+	Signer *identity.Signer
+}
+
+// GenerateToken generates a token
+func (a *CertificateAuthenticator) GenerateToken(ttl time.Duration) (string, error) {
```
sargun: Am I missing something here, or isn't most of this functionality in the `jwt-go` library?

sargun: Or is this being ripped out?

hankjacobs: It’s being ripped out.

sargun: :D

vendor/modules.txt
```diff
@@ -161,18 +163,9 @@ github.com/golang/glog
 # github.com/golang/groupcache v0.0.0-20200121045136-8c9f03a8e57e
 github.com/golang/groupcache/lru
 # github.com/golang/protobuf v1.3.3
-github.com/golang/protobuf/proto
```
sargun: Why did all these get removed?

sargun:
There are three make targets you might find helpful:

`lint - Runs lint on the files. Returns linting errors.`
`fmt - fmts all the files`
`protogen - generate new golang wrappers for the protobufs. Run the fmt target after this.`

hankjacobs:
Re #1: My thought when implementing it was that it's OK to generate a new key on startup because worse comes to worse, clients will fetch a new token. They already have to deal with token expiration and the error looks the same to the client in both cases . That being said, I don't have strong feelings about it so I'm game to do your suggestion if you do feel strongly about it.

Re #2: Great idea. It should simply the code a good deal. I'll go that route.

hankjacobs:
We could also use the task ID as the key that way it’s unique per task but is the same across restarts.

sargun:
I think we should have some kind of configurable value appended to task ID as the key for the HMAC, and then we're good to go.

hankjacobs:
So after going down a pretty deep rabbit hole I came to the conclusion that the aws sdk's metadata client isn't suitable for what we want to do. I could make it work but I'd have to abuse the hell out of the API it exposes. That being said, I replaced the proxying server implementation with one using an http client. I still have to fix up the tests and finish applying your feedback but its looking good so far.

sargun:
Looks good so far. Part of me almost thinks we should split it into two parts. 
1) Serving up the IMDSv2 API to our clients
2) Consuming the IMDSv2 API from AWS

#1 I see as almost done, except a little bit of logging, and making the key stable across restarts.

Do you want to to do that work first? and commit it, and then we can figure out the 2nd part later? I think the only thing that would be required is that the current http proxy would have to strip the token when proxying requests.

hankjacobs:
Sure. I'll go that route.

hankjacobs:
This should be good to go now.

cmd/titus-metadata-service/main.go
```diff
@@ -236,6 +251,8 @@ func main() {
 			Optimistic:          optimistic,
 			APIProtectEnabled:   apiProtectEnabled,
 			StateDir:            stateDir,
+			RequireToken:        requireToken,
```
sargun: Can we add a (logging) warning is the salt is empty?

hankjacobs:
Actually, I want to clean up the Verify api real quick, no need for the bool

metadataserver/auth/auth.go
```diff
@@ -0,0 +1,55 @@
...
+	var claims jwt.StandardClaims
+	jwtToken, err := jwt.ParseWithClaims(token, &claims, func(token *jwt.Token) (interface{}, error) {
+		if _, ok := token.Method.(*jwt.SigningMethodHMAC); !ok {
+			return nil, fmt.Errorf("Unexpected signing method: %v", token.Header["alg"])
+		}
+		return a.Key, nil
+	})
+
+	if err != nil {
+		return false, 0, errors.Wrap(err, "Token invalid")
```
sargun: Any reason not to make the function signature (int64, error). If err != nil, then you can assume token is invalid.

metadataserver/auth/auth.go
```diff
@@ -0,0 +1,55 @@
...
+
+	if !jwtToken.Valid {
+		return false, 0, errors.Wrap(err, "Token invalid")
+	}
+
+	if claims.Valid() != nil {
+		return false, 0, errors.Wrap(err, "Token invalid")
+	}
+
+	if !claims.VerifyAudience(a.Audience, false) {
```
sargun: Maybe errors.wrap should be errors.Wrapf("Token invalid, expected audience %q but have audience %q", claims.Audience)

metadataserver/auth/auth.go
```diff
@@ -0,0 +1,55 @@
...
+		}
+		return a.Key, nil
+	})
+
+	if err != nil {
+		return false, 0, errors.Wrap(err, "Token invalid")
+	}
+
+	if !jwtToken.Valid {
+		return false, 0, errors.Wrap(err, "Token invalid")
```
sargun: I don't think you want to return a wrapped error here, because err will = nil based on above.

metadataserver/auth/auth.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+
+	if claims.Valid() != nil {
+		return false, 0, errors.Wrap(err, "Token invalid")
```
sargun: I think you want to return the error here that was returned from claims.Valid(), and not the one from above

metadataserver/authentication.go
```diff
@@ -0,0 +1,109 @@
...
+			metrics.PublishIncrementCounter("auth.skipped.count")
+			next.ServeHTTP(w, r)
+			return
+		}
+
+		auth := auth.JWTAuthenticator{Key: ms.tokenKey, Audience: ms.titusTaskInstanceID}
+		valid, remaining, err := auth.VerifyToken(token)
+		if !valid {
+			if err != nil {
+				log.Error("Token invalid: ", err)
```
sargun: Please use log.WithError(err).Error("Token invalid")

metadataserver/authentication.go
```diff
@@ -0,0 +1,109 @@
...
+		next.ServeHTTP(w, r)
+	})
+}
+
+func (ms *MetadataServer) createAuthTokenHandler(w http.ResponseWriter, r *http.Request) {
+	metrics.PublishIncrementCounter("handler.createToken.count")
+
+	forwarded := r.Header.Get("x-forwarded-for")
+	if len(forwarded) > 0 {
+		log.Error("`x-forwarded-for` header present, blocking request`")
```
sargun: please make log the value of x-forwarded-for, so something like:
`log.WithField("xForwardedFor", forwarded).Error(..)`

Also, I don't think we need the backticks.

hankjacobs:
All cleaned up.

metadataserver/authentication.go
```diff
@@ -0,0 +1,109 @@
...
+	ttlStr := r.Header.Get("X-aws-ec2-metadata-token-ttl-seconds")
+
+	ttlSec, err := strconv.Atoi(ttlStr)
+	if err != nil {
+		log.Error("Could not decode ttl: ", err)
+		http.Error(w, "", http.StatusBadRequest)
+		return
+	}
+
+	if ttlSec < 0 || ttlSec > 21600 {
```
sargun: (6 * time.Hour).Seconds()?

metadataserver/authentication.go
```diff
@@ -0,0 +1,109 @@
...
+		log.Error("Invalid ttl: ", ttlSec)
```
sargun: Please use withfield

metadataserver/authentication.go
```diff
@@ -0,0 +1,109 @@
...
+		http.Error(w, "", http.StatusBadRequest)
+		return
+	}
+
+	token, err := auth.GenerateToken(time.Duration(ttlSec) * time.Second)
+	if err != nil {
+		log.Error("Could not generate token: ", err)
```
sargun: Please use witherror.

metadataserver/authentication.go
```diff
@@ -0,0 +1,109 @@
...
+		http.Error(w, "", http.StatusInternalServerError)
+		return
+	}
+
+	hj, ok := w.(http.Hijacker)
+	if !ok {
+		// Not a Hijacker, just treat it as a normal ResponseWriter
+		if _, err := fmt.Fprint(w, token); err != nil {
+			log.Error("Unable to write token: ", err)
```
sargun: Please use witherror.

metadataserver/authentication.go
```diff
@@ -0,0 +1,109 @@
...
+		}
+		return
+	}
+
+	conn, bufrw, err := hj.Hijack()
+	if err != nil {
+		log.Error("Unable to hijack connection: ", err)
```
sargun: Please use witherror.

metadataserver/authentication.go
```diff
@@ -0,0 +1,109 @@
...
+	hj, ok := w.(http.Hijacker)
+	if !ok {
+		// Not a Hijacker, just treat it as a normal ResponseWriter
+		if _, err := fmt.Fprint(w, token); err != nil {
+			log.Error("Unable to write token: ", err)
+		}
+		return
+	}
+
+	conn, bufrw, err := hj.Hijack()
```
sargun: TIL. about this API. Neat.

metadataserver/authentication.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	if conn.RemoteAddr().(*net.TCPAddr).IP.To16() != nil {
+		p := ipv6.NewConn(conn)
+		_ = p.SetHopLimit(1)
+	}
+
+	_, err = bufrw.Write([]byte(token))
+	if err != nil {
+		log.Error("Unable to write token: ", err)
```
sargun: Please use withErr

metadataserver/authentication.go
```diff
@@ -0,0 +1,109 @@
...
+		return
+	}
+	defer conn.Close()
+
+	if conn.RemoteAddr().(*net.TCPAddr).IP.To4() != nil {
+		p := ipv4.NewConn(conn)
+		_ = p.SetTTL(1)
+	}
+
+	if conn.RemoteAddr().(*net.TCPAddr).IP.To16() != nil {
```
sargun: maybe else if?

metadataserver/authentication.go
```diff
@@ -0,0 +1,109 @@
...
+		p := ipv6.NewConn(conn)
+		_ = p.SetHopLimit(1)
+	}
```
sargun: Maybe else {
log.warning("could not determine protocol type?
}

metadataserver/identity/verify.go
```diff
@@ -0,0 +1,126 @@
+package identity
```
sargun: I'm a little bit confused. Is this code only for testing? Or is it used elsewhere?

sargun:
Whoops, I reviewed a bunch of outdated bits.

metadataserver/server_test.go
```diff
@@ -633,3 +635,69 @@ func TestTaskIdentityWithECDSA(t *testing.T) {
...
+func TestRequireToken(t *testing.T) {
+	ess, err := setupStubServer(t)
+	if err != nil {
+		t.Fatal("Could not get stub server: ", err)
+	}
+
+	ms := setupMetadataServer(t, ess, ecdsaCerts[0], true)
+
+	// Get Token
+	tokenPath := fmt.Sprintf("http://%s%s", ess.proxyListener.Addr().String(), "/latest/api/token")
```
sargun: Any reason we don't use the "official" ec2metadata client here? and just override the EC2 metdata URL?

metadataserver/identity/verify.go
hankjacobs: It was used. It is no longer used. I was on the fence on whether I should remove it or not because I can see having signature verification being potentially handy at some point. I don't have a strong feeling either way. Your call.

metadataserver/authentication.go
hankjacobs: The back ticks are just for stylization purposes when printed but I'll remove.

metadataserver/identity/verify.go
sargun: They seem useful. Maybe just add a comment indicating they're helper functions or something? I can imagine the metatron client consuming these at some point.

metadataserver/server_test.go
hankjacobs: Good point. I’ll probably add that as another test since you can’t really inspect the response status codes directly.

hankjacobs:
I'm really glad you suggested using the ec2metadata client. Once a connection is hijacked, the full lifecycle of the connection and http response is the responsibility of the hijacker. My first implementation just wrote the token and none of the HTTP status or headers. That resulted in a malformed response which, funnily enough, curl just happily accepts so I didn't catch it when I was doing testing on an instance. The ec2metadata client did.

hankjacobs:
@sargun I'd like your eyes on the new changes to createAuthTokenHandler when you have a moment. Specifically with the manually writing the http status and headers to the hijacked connection. Thanks.

sargun:
Dumb question. Is the entire point of the hijacker to be able to set the connection TCP max TTL to 1?

hankjacobs:
Yep exactly. I almost ditched it but it does feel like a good security feature.

sargun:
1) Can we make the X-Forwarded-For check a warning right now, and then put it behind a command line option. We don't know if anyone in the fleet is relying on this behaviour right now, and breaking people's IMDSs sucks (it's hard for them to debug)

sargun:
2) You may have already looked at this option, but can we instead wrap the listener, like:
https://repl.it/repls/PaleCrushingSubweb

sargun:
This way, the TTL-setting happens at the listener. If you want to get really exciting, you can figure out how to set the TTL on the listener itself. (Keep in mind, this code only runs on Linux and Mac OS X).

sargun:
That way there's significantly less exciting issues that can occur with the world of HTTP.

sargun:
Right now, it's kind of jank, because the listener "work" happens in `cmd/titus-vpc-tool`, I don't see a strong reason not to pass down the "raw" (unwrapped) listener down through mdsCfg

sargun:
(I might be suggesting something stupid, or you may have already gone down this route. In which case, I'll look at the thing more closely. At first glance, it looks okay)

PR #360, opened 2020-03-16
Wait for network to teardown / shutdown prior to deallocating network

sargun:
There is a race condition where the "network" can be still running
in the container, but we will tell the IP service that it's no longer
in use.

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -86,6 +86,8 @@ type Config struct { // nolint: maligned
 
 	titusIsolateBlockTime   time.Duration
 	enableTitusIsolateBlock bool
+
+	networkTeardownTimeout time.Duration
 }
 
 type volumeContainerConfig struct {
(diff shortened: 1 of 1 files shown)


executor/runtime/docker/docker.go
```diff
@@ -1997,8 +1997,15 @@ func (r *DockerRuntime) Kill(c *runtimeTypes.Container) error { // nolint: gocyc
 	}
 
 stopped:
+	l := log.WithField("taskId", c.TaskID)
 	if c.SetupCommand != nil && c.SetupCommand.Process != nil {
 		_ = c.SetupCommand.Process.Signal(unix.SIGTERM) // nolint: gosec
+		time.AfterFunc(30*time.Second, func() {
```
gabrielhartmann: timeout should be a constant / configurable maybe

gabrielhartmann:
There are 3 cases where errors are dropped.  It would be nice to log when they're not nil so we know what happened.  I understand hitting errors here will probably never happen.  But if they were hit it, it would be really informative in debugging I think.

executor/runtime/docker/docker.go
```diff
@@ -1997,8 +1997,15 @@ func (r *DockerRuntime) Kill(c *runtimeTypes.Container) error { // nolint: gocyc
...
+			_ = c.SetupCommand.Process.Kill() // nolint: gosec
+		})
+		l.Info("Waiting for network teardown to finish")
+		_ = c.SetupCommand.Wait() // nolint: gosec
```
rgulewich: Do we need to call `Stop()` on the timer returned by `time.AfterFunc()` here, or is it fine to let that function run when SIGTERM caused the process to exit?

PR #362, opened 2020-03-19
Prepare VPC Service For Kubernetes / Kubelet

sargun:
This includes two big capabilities:
* Server-side HTB class ID allocation
* Idempotent Assign API


sargun:
It looks like this isn't passing CI, because gpg/gpg-agent was being installed through "recommends"

PR #363, opened 2020-03-24
Add container logviewer proxy

rgulewich:
The proxy sits on the host, and does a `Setns()` into a container's network namespace before connecting to its logviewer.

### DIFF
--- a/logviewer/api/transport_linux.go
+++ b/logviewer/api/transport_linux.go
@@ -5,16 +5,21 @@ package api
 import (
 	"context"
 	"fmt"
-	"golang.org/x/sys/unix"
 	"net"
 	"os"
 	"runtime"
 	"time"
 
+	"golang.org/x/sys/unix"
+
+	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 )
 
-const titusInits = "/var/lib/titus-inits"
+const (
+	titusInits      = "/var/lib/titus-inits"
+	curThreadNsPath = "/proc/thread-self/ns/net"
+)
 
 // nsDialer is a net.Dialer that does a Setns() into a container's network namespace before doing a connect
 type nsDialer struct {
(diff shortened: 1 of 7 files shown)


logviewer/api/transport_linux.go
```diff
@@ -0,0 +1,81 @@
...
+	"runtime"
+	"time"
+
+	log "github.com/sirupsen/logrus"
+)
+
+const titusInits = "/var/lib/titus-inits"
+
+// nsDialer is a net.Dialer that does a Setns() into a container's network namespace before doing a connect
+type nsDialer struct {
```
alekar: Can you extract this into the utility package? This is a pattern that will be required, for example, by metadata service when running on kubelet.

logviewer/api/transport_linux.go
```diff
@@ -0,0 +1,81 @@
...
+func (d *nsDialer) DialContext(ctx context.Context, network, address string) (net.Conn, error) {
+	log.Debugf("nsDialer: DialContext: %s", d.containerNetNsPath)
+	curThreadNsPath := fmt.Sprintf("/proc/%d/task/%d/ns/net", os.Getpid(), unix.Gettid())
+
+	// Make sure that this code executes only in this thread, since `Setns()` changes the thread's namespace
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+
+	// Get the FD for the host namespace, so we can switch back into it after doing the connect
+	prevNsFile, err := os.Open(curThreadNsPath)
```
sargun: Golang x/unix has an Open API that might be slightly nicer to consume: https://godoc.org/golang.org/x/sys/unix#Open

logviewer/api/transport_linux.go
```diff
@@ -0,0 +1,81 @@
...
+
+// nsDialer is a net.Dialer that does a Setns() into a container's network namespace before doing a connect
+type nsDialer struct {
+	containerNetNsPath string
+	systemDialer       net.Dialer
+}
+
+func (d *nsDialer) DialContext(ctx context.Context, network, address string) (net.Conn, error) {
+	log.Debugf("nsDialer: DialContext: %s", d.containerNetNsPath)
+	curThreadNsPath := fmt.Sprintf("/proc/%d/task/%d/ns/net", os.Getpid(), unix.Gettid())
```
sargun: Why not `/proc/self/ns/net`?

logviewer/api/transport_linux.go
```diff
@@ -0,0 +1,81 @@
...
+
+	// Make sure that this code executes only in this thread, since `Setns()` changes the thread's namespace
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+
+	// Get the FD for the host namespace, so we can switch back into it after doing the connect
+	prevNsFile, err := os.Open(curThreadNsPath)
+	if err != nil {
+		return nil, fmt.Errorf("error opening ns file %v: %v", curThreadNsPath, err)
```
sargun: Please use errors.wrapf. It's "go 1.13 convention"

logviewer/api/transport_linux.go
rgulewich: In person conversation: we agreed to leave it as-is, and @alekar will factor it out as part of the metadataservice CNI work.

logviewer/api/transport_linux.go
rgulewich: Paranoia - I wasn't sure if the ns for the entire process could change, if a bunch of its threads were all doing `Setns()`.  Not sure if that's possible or not.  Regardless, if I go the per-thread ns route, `/proc/thread-self` makes more sense.

PR #364, opened 2020-03-25
Respect all disk resource names.

alekar:
This allows us to allocate the correct amount of disk space when the scheduler sets the resource according to the old Titus resource name, or the k8s resource name.

### DIFF
--- a/vk/backend/backend.go
+++ b/vk/backend/backend.go
@@ -18,6 +18,7 @@ import (
 	"github.com/virtual-kubelet/virtual-kubelet/log"
 	"golang.org/x/sys/unix"
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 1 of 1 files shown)


vk/backend/backend.go
```diff
@@ -101,7 +102,16 @@ func RunWithBackend(ctx context.Context, runner *runner.Runner, statuses *os.Fil
 	}
 
 	requests := pod.Spec.Containers[0].Resources.Requests
-	disk := requests["titus/disk"]
+
+	var disk resource.Quantity
+
+	for _, k := range []v1.ResourceName{"titus/disk", v1.ResourceEphemeralStorage, v1.ResourceStorage} {
```
sargun: I think if multiple labels are assigned we should error out, right?

alekar: My guess is they will set both during migration. We simply pick the first one we encounter.

gabrielhartmann: I believe they will not set both. We will advertise both, but they will only pick one at any given time.  Regardless, still looks fine.

PR #368, opened 2020-03-26
Wait for IPv6 default route to be configured 

sargun:
This polls for a default IPv6 route before it considers IPv6 "ready"


sargun:
It also removes DAD, because "we" enforce DAD via our assignment scheme.

PR #369, opened 2020-03-28
IMDS: Delete existing Server header when proxying

hankjacobs:
### Description of the Change

Delete existing Server header when proxying to prevent doubling.


sargun:
lgtm

PR #370, opened 2020-03-29
Metadata Proxy Token Key Salt 

hankjacobs:
Set the token key salt for the metadata proxy to node's hostname. 

As I was testing this, I realized that the running container already knows one part of the key (the task ID) so the salt should be something the container has no knowledge of or else it could mint tokens. I _think_ the hostname of the node satisfies that constraint but I'd love some feedback on that. That being said, I think the worst it could do with minting abilities is to create a token that exceeds the max TTL.

### DIFF
--- a/root/lib/systemd/system/titus-metadata-proxy@.service
+++ b/root/lib/systemd/system/titus-metadata-proxy@.service
@@ -15,7 +15,7 @@ NotifyAccess=all
 WatchdogSec=30
 
 Environment=TITUS_PID_1_DIR=/var/lib/titus-inits/%i
-EnvironmentFile=/run/titus-metadata-proxy.salt
+EnvironmentFile=-/run/titus-metadata-proxy.salt
 EnvironmentFile=/var/lib/titus-environments/%i.env
 ExecStart=/apps/titus-executor/bin/run-titus-metadata-proxy.sh
 LimitNOFILE=65535
--- a/root/lib/systemd/system/titus-metadata-proxy@.service
+++ b/root/lib/systemd/system/titus-metadata-proxy@.service
@@ -6,12 +6,16 @@ ConditionPathIsDirectory=/var/lib/titus-inits/%i/ns
 StartLimitIntervalSec=30
 StartLimitBurst=10
 
+After=titus-salt-generator.service
+Wants=titus-salt-generator.service
+
 [Service]
 Type=notify
 NotifyAccess=all
 WatchdogSec=30
 
 Environment=TITUS_PID_1_DIR=/var/lib/titus-inits/%i
+EnvironmentFile=/run/titus-metadata-proxy.salt
 EnvironmentFile=/var/lib/titus-environments/%i.env
 ExecStart=/apps/titus-executor/bin/run-titus-metadata-proxy.sh
 LimitNOFILE=65535
--- a/root/lib/systemd/system/titus-salt-generator.service
+++ b/root/lib/systemd/system/titus-salt-generator.service
@@ -4,7 +4,7 @@ Description=Write salt for Titus IMDS
 [Service]
 Type=oneshot
 RemainAfterExit=true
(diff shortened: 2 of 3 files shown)


sargun:
I think it's probably better to set the salt to a random value in our boot scripts.

sargun:
What we probably want to do is add into the unit file something like https://github.com/Netflix/titus-executor/blob/master/root/lib/systemd/system/titus-metadata-proxy%40.service:
```
[Unit]
After=titus-salt-generator.service
Wants=titus-salt-generator.service
[Service]
EnvironmentFile=/run/titus-metadata-proxy.salt
```

And then add another unit called titus-salt-generator.service
that writes /run/titus-metdata-proxy.salt, which might as well be:

```bash
[Unit]
Description=Write salt for titus IMDS

[Service]
Type=oneshot
RemainAfterExit=true
ExecStart=/bin/bash -x '/bin/cat /proc/sys/kernel/random/uuid > /run/titus-metdata-proxy.salt
LimitCORE=infinity

[Install]
WantedBy=multi-user.target
```

hankjacobs:
Thanks. I'll tackle it this weekend.

root/lib/systemd/system/titus-metadata-proxy@.service
```diff
@@ -6,12 +6,16 @@ ConditionPathIsDirectory=/var/lib/titus-inits/%i/ns
 StartLimitIntervalSec=30
 StartLimitBurst=10
 
+After=titus-salt-generator.service
```
sargun, line 9: @gabrielhartmann is the systemd expert on the team. Is this right? We have a "soft" dep on this thing starting.

root/lib/systemd/system/titus-salt-generator.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=Write salt for Titus IMDS
+
+[Service]
+Type=oneshot
+RemainAfterExit=true
+ExecStart=/bin/bash -cx 'echo -n TOKEN_KEY_SALT=\" > /run/titus-metadata-proxy.salt && /bin/cat /proc/sys/kernel/random/uuid | tr -d "\n" >> /run/titus-metdata-proxy.salt && echo \" >> /run/titus-metdata-proxy.salt'
```
sargun, line 7: Maybe I'm missing something, but why do we need the quotes around setting the variable, and the tr?

root/lib/systemd/system/titus-salt-generator.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
```
sargun, line 1: Another @gabrielhartmann question. Do we need to explicitly enable this, or will the first time the titus-metatdataproxy is started start this?

root/lib/systemd/system/titus-metadata-proxy@.service
```diff
@@ -6,12 +6,16 @@ ConditionPathIsDirectory=/var/lib/titus-inits/%i/ns
...
+Wants=titus-salt-generator.service
+
 [Service]
 Type=notify
 NotifyAccess=all
 WatchdogSec=30
 
 Environment=TITUS_PID_1_DIR=/var/lib/titus-inits/%i
+EnvironmentFile=/run/titus-metadata-proxy.salt
```
sargun, line 18: We might want to make this a soft-dep, so:
`EnvironmentFile=-/run/titus-metadata-proxy.salt`

(If the dep file is failed to be written we'd rather be "available and insecure" than unavailable)

root/lib/systemd/system/titus-salt-generator.service
hankjacobs: I guess it's technically not needed. I quote out of habit and caution (less potential for breakage). `tr` is necessary since `/proc/sys/kernel/random/uuid` outputs a newline.

hankjacobs: It became an escaping nightmare. I ended up reverting to what you wrote.

root/lib/systemd/system/titus-metadata-proxy@.service
hankjacobs: I believe `Wants` accomplishes the that.

root/lib/systemd/system/titus-salt-generator.service
hankjacobs: From experimentation, it runs the first time the proxy is started.

PR #371, opened 2020-04-01
Implement /latest/meta-data/hostname endpoint in IMDS

sargun:
Implements a new IMDS endpoint


metadataserver/server.go
```diff
@@ -189,6 +190,9 @@ func (ms *MetadataServer) publicIPV4(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+// https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-categories.html
+// The private IPv4 DNS hostname of the instance. In cases where multiple network interfaces are present,
+// this refers to the eth0 device (the device for which the device number is 0).
 func (ms *MetadataServer) localHostname(w http.ResponseWriter, r *http.Request) {
 	metrics.PublishIncrementCounter("handler.localIPV4.count")
```
rgulewich, line 197: I know this isn't really the scope of the fix, but we should probably fix this metric name to `handler.localHostname.count`.

PR #372, opened 2020-04-01
Rotate .out files to S3, so Tomcat apps don't run out of disk space

rgulewich:
Tomcat writes files named like `catalina_20200331_00.out`. We're currently not rotating / uploading these, which causes long-running Java apps to run out of disk space.

### DIFF
--- a/filesystems/watcher.go
+++ b/filesystems/watcher.go
@@ -63,6 +63,7 @@ var defaultUploadRegexpList = []*regexp.Regexp{
 	regexp.MustCompile(`stdout.[\d_\.]+$`),
 	regexp.MustCompile(`stderr.[\d_\.]+$`),
 	regexp.MustCompile(`[\d-_]+.log`),
+	regexp.MustCompile(`[\d-_]+.out`),
 	regexp.MustCompile(`[\S]+\.core\.[\d]+\.[\d]+`),
 }
 
--- a/filesystems/watcher_test.go
+++ b/filesystems/watcher_test.go
@@ -204,51 +204,38 @@ func TestDoubleUpload(t *testing.T) { // nolint: gocyclo
 	}
 }
 
-func TestRotateRegexp(t *testing.T) { // nolint: gocyclo
-	shouldRotate := CheckFileForRotation("stdout", nil)
-	assert.False(t, shouldRotate, "stdout should not rotate")
-
-	shouldRotate = CheckFileForRotation("stdout.2", nil)
-	assert.True(t, shouldRotate, "stdout.2 should rotate")
-
-	shouldRotate = CheckFileForRotation("stout.2", nil)
-	assert.False(t, shouldRotate, "stout.2 should not rotate")
-
-	shouldRotate = CheckFileForRotation("stdout.30092016_151555.073", nil)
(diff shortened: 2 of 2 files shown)


filesystems/watcher_test.go
```diff
@@ -206,73 +206,49 @@ func TestDoubleUpload(t *testing.T) { // nolint: gocyclo
 
 func TestRotateRegexp(t *testing.T) { // nolint: gocyclo
```
sargun, line 207: You should be able to remove the nolint here.

filesystems/watcher.go
```diff
@@ -63,6 +63,7 @@ var defaultUploadRegexpList = []*regexp.Regexp{
 	regexp.MustCompile(`stdout.[\d_\.]+$`),
 	regexp.MustCompile(`stderr.[\d_\.]+$`),
 	regexp.MustCompile(`[\d-_]+.log`),
+	regexp.MustCompile(`[\d-_]+.out`),
```
sargun, line 66: I imagine this is a temporary hack?

filesystems/watcher_test.go
```diff
@@ -206,73 +206,49 @@ func TestDoubleUpload(t *testing.T) { // nolint: gocyclo
...
+	assert.True(t, shouldRotate, "%s should rotate", nqCore)
 
 	nqBadCore := "nodequark.core.13213.abb.3213"
 	shouldRotate = CheckFileForRotation(nqBadCore, nil)
-	if shouldRotate {
-		t.Fatalf("%s should not rotate", nqBadCore)
-	}
+	assert.False(t, shouldRotate, "%s should not rotate", nqBadCore)
+
+	shouldRotate = CheckFileForRotation("catalina_20200330_21.out", nil)
```
sargun, line 250: Does it make sense to follow the pattern above:

```go
catalainaOut := "catalina_20200330_21.out"
shouldRotate = CheckFileForRotation(catalinaOut, nil)
assert.True(t, shouldRotate, "%s should rotate")
```

filesystems/watcher.go
rgulewich: As long as we have Tomcat apps running on Titus, we'll need to support this.

filesystems/watcher_test.go
rgulewich: I started down that road, realized that I should turn this into a table-based test, and then decided that the diff was getting too big.  I'll do the larger change as part of another commit.

filesystems/watcher_test.go
```diff
@@ -204,51 +204,38 @@ func TestDoubleUpload(t *testing.T) { // nolint: gocyclo
...
-	assert.False(t, shouldRotate, "%s should not rotate", nqBadCore)
-
-	shouldRotate = CheckFileForRotation("catalina_20200330_21.out", nil)
-	assert.True(t, shouldRotate, "catalina_20200330_21.out should rotate")
+	var testData = []struct {
+		filename string
+		result   bool
+		regexp   *regexp.Regexp
+	}{
+		{"stdout", false, nil},
```
sargun, line 215: Fancy.

PR #375, opened 2020-04-04
Add `eni-resource-set` command to titus-vpc-tool

rgulewich:
This allows the kubelet poststart script to correctly register the `res` annotation.


rgulewich:
> Doesn’t the config command give you the resource set info?

Nice, didn't know that. Thanks!

PR #377, opened 2020-04-07
Do not make an assignment GCable, until it has existed for 5 minutes

sargun:


vpc/service/gc_v3.go
```diff
@@ -170,6 +170,7 @@ FROM branch_eni_attachments
 JOIN assignments ON branch_eni_attachments.association_id = assignments.branch_eni_association
 WHERE trunk_eni = $1
 AND assignments.assignment_id NOT IN (SELECT unnest($2::text[]))
+AND created_at < now() - INTERVAL '5 minutes'
```
rgulewich, line 173: It would be nice if this was a const at the top of the file, but I'm not too fussed.

sargun: I couldn't figure out how to do this.

sargun: PRs welcome. :P. (No, seriously, https://github.com/lib/pq/issues/78)

PR #381, opened 2020-04-08
K8s imds proxy

alekar:
Enable running metadata proxy under kubernetes. Includes some minor refactorings and removal of dead code.


cmd/titus-metadata-service/main.go
```diff
@@ -235,29 +202,44 @@ func main() {
 			EnvVar:      "X_FORWARDED_FOR_BLOCKING_MODE",
 			Destination: &xFordwardedForBlockingMode,
 		},
+		cli.StringFlag{
+			Name:        "peer-namespace",
```
sargun, line 206: I am not a fan of adding this logic to the metadata proxy. I liked it better when we could have a separation of concerns, and the process which was wiring up the network namespaces was different than the one doing the IMDS work.

The reasoning behind this is a sense of "purity" of the IMDS code, in that it doesn't do any systems or OS-level stuff.

cmd/titus-metadata-service/main.go
```diff
@@ -235,29 +202,44 @@ func main() {
...
-		// for logs to get from the tasks starts by titus tasks to
-		logsutil.MaybeSetupLoggerIfOnJournaldAvailable()
+
+		utils.MaybeSetupLoggerIfOnJournaldAvailable()
+
 		/* Get the requisite configuration from environment variables */
-		listener := getListener(listenPort, listenerFd)
+		var listener net.Listener
+
+		if len(peerNs) > 0 {
```
sargun, line 225: See above.

cmd/titus-metadata-service/main.go
```diff
@@ -235,29 +202,44 @@ func main() {
...
+			// We were launched by a CNI. Bind inside peer namespace.
+			log.Infof("Launched with PEER_NS %s, LISTEN_PORT %d", peerNs, listenPort)
+			nsListener, err := utils.GetNsListener(peerNs, listenPort)
+			if err != nil {
+				log.Fatalf("Error getting listener %s", err)
+			}
+
+			listener = nsListener
+		} else {
+			listener = getListener(listenPort, listenerFd)
```
sargun, line 235: I think if you want to continue this behaviour (logic of getting the listener in this code), then should pass it into getListener, so getListener can validate that listenerPort XOR listener FD are set, and if listenerNS is set, then ensure listener FD is not set.

cmd/titus-metadata-service/main.go
```diff
@@ -235,29 +202,44 @@ func main() {
...
 		/* Get the requisite configuration from environment variables */
-		listener := getListener(listenPort, listenerFd)
+		var listener net.Listener
+
+		if len(peerNs) > 0 {
+			// We were launched by a CNI. Bind inside peer namespace.
+			log.Infof("Launched with PEER_NS %s, LISTEN_PORT %d", peerNs, listenPort)
+			nsListener, err := utils.GetNsListener(peerNs, listenPort)
+			if err != nil {
+				log.Fatalf("Error getting listener %s", err)
```
sargun, line 230: Shouldn't this return an error?

so:
`return errors.Wrap(err, "Could not get listener")`

cmd/titus-reaper/main.go
```diff
@@ -21,7 +22,7 @@ func main() {
 	if debug {
 		log.SetLevel(log.DebugLevel)
 	} else {
-		logsutil.MaybeSetupLoggerIfOnJournaldAvailable()
+		utils.MaybeSetupLoggerIfOnJournaldAvailable()
```
sargun, line 25: It titus-repear still a thing?

go.mod
```diff
@@ -85,6 +85,7 @@ require (
 	github.com/wercker/journalhook v0.0.0-20180428041537-5d0a5ae867b3
 	github.com/wsxiaoys/terminal v0.0.0-20160513160801-0940f3fc43a0 // indirect
 	go.opencensus.io v0.22.3
+	go.uber.org/multierr v1.1.0
```
sargun, line 88: We use `https://github.com/hashicorp/go-multierror` elsewhere in the codebase. Any reason for the switch?

metadataserver/cni/env.go
```diff
@@ -0,0 +1,249 @@
...
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path"
+	"strconv"
+
+	"github.com/Netflix/titus-executor/api/netflix/titus"
+	"github.com/gogo/protobuf/proto"
+
```
sargun, line 17: Formatting seems to have some problems here.

metadataserver/cni/env.go
```diff
@@ -0,0 +1,249 @@
...
+func getPod(args *skel.CmdArgs) (*v1.Pod, error) {
+	var k8sArgs utils.K8sArgs
+	err := types.LoadArgs(args.Args, &k8sArgs)
+	if err != nil {
+		return nil, errors.Wrap(err, "Unable to parse CNI k8s args")
+	}
+	logrus.Debugf("k8s args %#v", k8sArgs)
+
+	ctx := context.Background()
+	defer ctx.Done()
```
sargun, line 80: Why `defer ctx.Done()`?
Should this be:

```go
ctx, cancel := context.WithCancel(context.Background())
defer cancel()
```

metadataserver/cni/env.go
```diff
@@ -0,0 +1,249 @@
...
+
+	return utils.GetPod(ctx, kubeletAPIPodsURL, k8sArgs)
+}
+
+func generateTokenKeySaltIfRequired(pod *v1.Pod) (string, error) {
+	fname := path.Join(tt.TitusEnvironmentsDir, fmt.Sprintf("%s.salt", pod.Name))
```
sargun, line 86: Is this the salt used for IMDSv2? Any reason not to leverage the existing mechanism of setting a system wide salt?

metadataserver/cni/env.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+
+	data, err := ioutil.ReadFile(fname)
+	return string(data), err
+}
+
+func extractEnv(prev *current.Result, pod *v1.Pod) (map[string]string, error) {
+	env := map[string]string{}
+
+	env["LISTEN_PORT"] = "80"
```
sargun, line 115: We should probably prefix these environment variables with `TITUS_`, as that an informal namespace.

metadataserver/cni/env.go
```diff
@@ -0,0 +1,249 @@
...
+		case "4":
+			env["EC2_LOCAL_IPV4"] = ip.Address.String()
+			ok = true
+		case "6":
+			env["EC2_IPV6S"] = ip.Address.String()
+			ok = true
+		}
+	}
+
+	if !ok {
```
sargun, line 133: Can you just make this is `env["EC2_LOCAL_IPV4"] == nil && env["EC2_IPV6S"] == nil` then return this error?

See `section 4.1` of [Out of the Tarpit](https://github.com/papers-we-love/papers-we-love/blob/master/design/out-of-the-tar-pit.pdf), that fuels my reasoning. I think the less state you need to store, and the more state you can derive from your "ground truth", the less likely you are to make errors. 

I think it's currently fine as is, but maybe food-for-thought. (I'll admit myself, that I've been lazy many times and stored state with side effects)

metadataserver/cni/env.go
```diff
@@ -0,0 +1,249 @@
...
+		return nil, errors.New("No IP information in chained CNI result")
+	}
+
+	// pod specific configuration
+	env["TITUS_TASK_INSTANCE_ID"] = pod.Name
```
sargun, line 138: So, I guess we're dropping support for `titusParameter.agent.hostnameStyle`? Do you know if anyone uses this feature?

metadataserver/cni/env.go
```diff
@@ -0,0 +1,249 @@
...
+
+	env["TITUS_IAM_ROLE"] = pod.Annotations[mt.IamRoleArnAnnotation]
+	env["X_FORWARDED_FOR_BLOCKING_MODE"] = pod.Annotations[mt.XForwardedForBlockingModeAnnotation]
+
+	// assume we are on a metatron enabled host if we find /metatron
+	_, err = os.Stat("/metatron")
+	if os.IsNotExist(err) {
+		env[mt.TitusMetatronVariableName] = "false"
+	}
+	env[mt.TitusMetatronVariableName] = "true"
```
sargun, line 156: Doesn't this result in this variable always being set to true?

metadataserver/cni/env.go
```diff
@@ -0,0 +1,249 @@
...
+
+	err = writeContainerInfo(pod)
+	if err != nil {
+		return errors.Wrap(err, "Unable to write containerInfo file")
+	}
+
+	return nil
+}
+
+func writeEnvFile(pod *v1.Pod, env map[string]string) error {
```
sargun, line 190: as I read more and more of this, I am beginning to think that the place where this should be should instead be in a CRI-wrapper. We should create (and delete) state as the pod sandbox is created.

sargun: (as well as the state file).

metadataserver/cni/env.go
```diff
@@ -0,0 +1,249 @@
...
+		_, err := fmt.Fprintf(fd, "%s=%s\n", key, qval)
+		if err != nil {
+			return err
+		}
+	}
+	return fd.Close()
+}
+
+func writeContainerInfo(pod *v1.Pod) error {
+	str, ok := pod.Annotations["containerInfo"]
```
sargun, line 216: I think this should definitely be a CRI wrapper or OCI hook in creation of the pod sandbox

metadataserver/cni/env.go
```diff
@@ -0,0 +1,249 @@
...
+	_, err = os.Stat("/metatron")
+	if os.IsNotExist(err) {
+		env[mt.TitusMetatronVariableName] = "false"
+	}
+	env[mt.TitusMetatronVariableName] = "true"
+
+	return env, nil
+}
+
+func setupEnv(pod *v1.Pod, env map[string]string) error {
```
sargun, line 161: Where does this function get called?

metadataserver/cni/env.go
sargun: that, or everything that relies on this info should just go back to the kubelet API.

metadataserver/cni/unit.go
```diff
@@ -0,0 +1,98 @@
...
+
+func startUnit(podName string) error {
+	conn, err := dbus.New()
+	if err != nil {
+		return err
+	}
+	defer conn.Close()
+
+	unitName := getUnitName(podName)
+	logrus.Infof("start unit %s", unitName)
```
sargun, line 26: use withfield.

metadataserver/types/annotations.go
```diff
@@ -0,0 +1,8 @@
+package types
+
+const (
+	IamRoleArnAnnotation                = "iam.amazonaws.com/role"
```
sargun, line 4: I think it'd be nice to have a package with our k8s annotations across all projects. Maybe titus-api-definitions?

sargun: Or even a single package amongst the executor.

metadataserver/types/annotations.go
```diff
@@ -0,0 +1,8 @@
...
+	RequireTokenAnnotation              = "imds.titus.netflix.com/requireToken" // nolint:gosec
```
sargun, line 5: why does this trigger gosec?

root/lib/systemd/system/titus-imds-proxy@.service
```diff
@@ -0,0 +1,26 @@
...
+
+[Service]
+Type=notify
+NotifyAccess=all
+WatchdogSec=30
+
+Restart=on-failure
+KillMode=mixed
+
+User=www-data
```
sargun, line 16: Why? Can we use https://www.freedesktop.org/software/systemd/man/systemd.exec.html#DynamicUser=

root/lib/systemd/system/titus-imds-proxy@.service
```diff
@@ -0,0 +1,26 @@
...
+Group=www-data
+
+LimitNOFILE=8384
```
sargun, line 19: Why 8384 rather than the original 65k?

utils/kubelet.go
```diff
@@ -0,0 +1,95 @@
...
+
+	"github.com/containernetworking/cni/pkg/types"
+
+	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+)
+
+func PodKey(pod *corev1.Pod) string {
```
sargun, line 18: I'm reluctant to make `utils` an actual grab-bag of random things. I would rather have a separate k8sutils

utils/nsdialer_linux.go
```diff
@@ -0,0 +1,45 @@
...
+	"net"
+
+	"github.com/pkg/errors"
+	"go.uber.org/multierr"
+
+	"github.com/vishvananda/netns"
+)
+
+func (n *NsDialer) DialContext(ctx context.Context, network, address string) (net.Conn, error) {
+	origNs, err := netns.Get()
```
sargun, line 16: Don't we need to lock our runtime thread here?

sargun:
I was a little bit more critical and nit-picky than normal, given this is a big, critical feature, and that this is a big jump into the codebase. Feel free to ignore me on most of the nit-picks. If you disagree with me, I'm happy to change me opinion.

metadataserver/cni/env.go
alekar: We cannot create this file till we have allocated an ENI for the pod. My plan, now that I also looking at setting up netflix environment, was to have this CNI write the file out, and use it to set environment for containers within the pod.

If we use a CRI-wrapper, we will have to make the VPC service calls from the wrapper.

metadataserver/types/annotations.go
alekar: metadataserver/types/annotations.go:5:2: G101: Potential hardcoded credentials (gosec)
	RequireTokenAnnotation              = "imds.titus.netflix.com/requireToken"

metadataserver/cni/env.go
sargun: If we have the CRI write out the environment variable file after the sandbox is created, we can do this. It's more work, but it might be the "better" thing to do.

Then I think we have two sidecars that we add:
1. The (unprivileged) IMDS metadata service that's network namespace is set to host network namespace (but it the FD passed to it is a shared volume, like `/run/imds.sock`)
2. A privileged sidecar (CAP_NET_ADMIN), that runs inside of the container and does unshare and just proxies to `/run/imds.sock`. I know you suggested we do something like this earlier, and you were right. I didn't consider a CRI "interposer" at the time. Without the CRI interposer you cannot do #1, as you cannot have some containers in the pod have the host namespace, and some have the other namespace. 

The "only" thing you need to do it protect the proxy from having its /proc/self/ns/net stolen by another process in the pod. In order to do that, we need to do something like set dumpable to false (https://github.com/containerd/cri/issues/1453).

In addition to that, you need to get "ordering" right, and ensure that #1 and #2 run before user code. 

----
Okay, this is quickly becoming a nightmare of an implementation story. 
----
Can we just intercept RunSandbox, write the environment file, and then start the systemd unit in the interposer?

go.mod
alekar: For `multierr.Combine`. I'll revert this.

metadataserver/cni/env.go
alekar: Yup, thanks.

metadataserver/cni/env.go
alekar: I took away from some conversation that we want per-pod salt. I can revert this.

cmd/titus-metadata-service/main.go
alekar: It's the main method of the metadata proxy.

cmd/titus-metadata-service/main.go
alekar: We do _not_ pass an FD if we are running under k8s. Which is the reason for this structure.

cmd/titus-reaper/main.go
alekar:
```
(nfsuper) ~ $ sudo systemctl list-units | grep reaper
  titus-reaper.service                                                                                                               loaded active running   Titus Reaper: Reaps Titus Containers
```

cmd/titus-metadata-service/main.go
alekar: This process is not accessible from the pod. If it is, we have a host compromise.

root/lib/systemd/system/titus-imds-proxy@.service
alekar: I forgot to push some code here. This service runs as a privileged user on the host. If we feel strongly otherwise, we can muck with some mount locations to move to a more locked down user.

utils/nsdialer_linux.go
alekar: Yup. Thanks.

root/lib/systemd/system/titus-imds-proxy@.service
alekar: Reverted.

alekar:
@sargun I believe I have addressed all of your comments?

sargun:
I thought you were going to implement a CRI which injected pods? :P. Let me take a look tomorrow. Is that okay?

alekar:
:-D A couple of days is fine, there is no rush.

metadataserver/cni/cni_linux.go
```diff
@@ -0,0 +1,420 @@
...
+
+	// forward declare for gotos
+
+	var peerLink, contLink netlink.Link
+	var route *netlink.Route
+
+	var peerNsPath string
+	var errA error
+
+	// move the peer veth into the peer namespace, link up
```
rgulewich, line 148: Thank-you for all of the high-level comments in this file - made following it a lot easier!

metadataserver/types/annotations.go
rgulewich: I think it should be a separate package - not the executor (we already have these names duplicated in the mutating webhook and the agent controllers) or api-definitions (I'd rather not have to vendor a bunch of generated proto files when all I want is a few golang constants).

rgulewich: ... doesn't have to happen as part of this PR, but we should do it soon.

metadataserver/cni/env.go
rgulewich: Does this need to be done via a CRI, or could the CNI launch containers to do this?  They wouldn't be visible to k8s, or to the pod, but that's a feature here, not a bug, I think.

I think the sidecar / socket approach is probably the most secure way of doing this, but doing this via a CRI wrapper seems like overkill.

metadataserver/cni/env.go
alekar: This only seems to be exposed by `/instance-id`.

PR #384, opened 2020-04-11
Add agent parameter to enable requiring of imdsv2 token

hankjacobs:
This PR adds an agent parameter that allows a container to specify if the IMDSv2 token is required to access the metadata service.

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -69,6 +69,7 @@ const (
 	subnets                 = "titusParameter.agent.subnets"
 	elasticIPPool           = "titusParameter.agent.elasticIPPool"
 	elasticIPs              = "titusParameter.agent.elasticIPs"
+	imdsRequireToken        = "titusParameter.agent.imds.requireToken"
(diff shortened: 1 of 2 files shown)


executor/runtime/docker/docker.go
```diff
@@ -574,6 +575,10 @@ func (r *DockerRuntime) dockerConfig(c *runtimeTypes.Container, binds []string,
 		c.Env["EC2_OWNER_ID"] = vpcAccountID
 	}
 
+	if requireToken, ok := c.TitusInfo.GetPassthroughAttributes()[imdsRequireToken]; ok {
```
sargun, line 578: Maybe we should rename REQUIRE_TOKEN to something else. We typically "control" variables under the TITUS_* namespace. So, maybe we should call it like TITUS_IMDS_REQUIRE_TOKEN?

(It'd require a change in two places)

hankjacobs: Sure thing, done. Let me know what you think.

PR #390, opened 2020-04-21
Allow an agent parameter to override default service mesh image.

alekar:
This allows the service mesh team to roll out canaries on select applications by overriding the global service mesh image with an agent parameter.

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -206,12 +206,17 @@ func shouldStartMetatronSync(cfg *config.Config, c *runtimeTypes.Container) bool
 }
 
 func shouldStartServiceMesh(cfg *config.Config, c *runtimeTypes.Container) bool {
-	if !cfg.ContainerServiceMeshEnabled || cfg.ContainerServiceMeshImage == "" {
+	if !cfg.ContainerServiceMeshEnabled {
 		return false
 	}
 
 	enabled, _ := c.GetServiceMeshEnabled()
-	return enabled
+	if !enabled {
+		return false
+	}
+
+	_, err := c.GetServiceMeshImage()
+	return err == nil
 }
 
 // NoEntrypointError indicates that the Titus job does not have an entrypoint, or command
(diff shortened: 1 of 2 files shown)


executor/runtime/docker/docker.go
```diff
@@ -206,12 +206,17 @@ func shouldStartMetatronSync(cfg *config.Config, c *runtimeTypes.Container) bool
...
 		return false
 	}
 
 	enabled, _ := c.GetServiceMeshEnabled()
-	return enabled
+	if !enabled {
+		return false
+	}
+
+	_, err := c.GetServiceMeshImage()
```
rgulewich, line 218: Overall, I like the idea, but I'm wondering if we should only allow setting the version (either by tag or digest), rather than the full image. My concerns are that this could allow users to pull from arbitrary registries, and that the code that gets pulled actually gets supervised from the host, which could potentially allow for interesting DoS opportunities.

snowp: Currently the way we're set up is to publish different images depending on the release stage (snapshot, candidate, final).  Would making just the image name + tag/digest (and not the registry) configurable be an option?

alekar: I propose we check that the attribute points to the right location. My personal take is that it is easier to debug if we make the attribute a fully qualified name.

```
func (c *Container) GetServiceMeshImage() (string, error) {
	defaultImage := c.Config.ContainerServiceMeshImage

	if  defaultImage == "" {
		return "no-image", errors.New("Could not find service mesh image")
	}

	image, ok := c.TitusInfo.GetPassthroughAttributes()[serviceMeshContainerParam]
	if !ok {
		return defaultImage, nil
	}

	// we only allow overrides to the docker registry specified by the fast property
	if strings.Split(image, "/")[0] != strings.Split(defaultImage, "/")[1] {
		return defaultImage, nil
	}

	return image, nil
}
```

rgulewich: OK, as long as it can't be set to pull from an arbitrary registry, that would be fine.  So the property would be in the format of `repo/image:tag`, correct?

executor/runtime/types/types.go
```diff
@@ -28,6 +28,7 @@ const (
 	KvmEnabledParam              = "titusParameter.agent.kvmEnabled"
 	assignIPv6AddressParam       = "titusParameter.agent.assignIPv6Address"
 	serviceMeshEnabledParam      = "titusParameter.agent.service.serviceMesh.enabled"
+	serviceMeshContainerParam    = "titusParameter.agent.service.serviceMesh.container"
```
rgulewich, line 31: Only other minor nit: might want to make this `containerImage`, or `image`, so it's a bit more in your face what this property is for.

alekar:
Re-working this pull request to default to the configured registry for all system services.

PR #391, opened 2020-04-23
Change how we specify system service image.

alekar:
Define new properties, that specify the system service container,
sans the docker registry. The registry defaults to configuration
for the agent.

Also add a passthrough parameter that allows overriding proxyd's
container image. This is to be used by the mesh team to run their
canaries.

### DIFF
--- a/executor/runtime/types/types.go
+++ b/executor/runtime/types/types.go
@@ -3,6 +3,7 @@ package types
 import (
 	"context"
 	"fmt"
+	"path"
 	"path/filepath"
 	"sort"
 	"strconv"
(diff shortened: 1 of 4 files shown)


executor/runtime/types/types.go
```diff
@@ -28,6 +29,7 @@ const (
 	KvmEnabledParam              = "titusParameter.agent.kvmEnabled"
 	assignIPv6AddressParam       = "titusParameter.agent.assignIPv6Address"
 	serviceMeshEnabledParam      = "titusParameter.agent.service.serviceMesh.enabled"
+	serviceMeshContainerParam    = "titusParameter.agent.service.serviceMesh.container"
```
rgulewich, line 32: Maybe `titusParameter.agent.service.serviceMesh.containerImage`?  I figure we should be as explicit as possible about what this is - container seems a bit too generic.

PR #394, opened 2020-05-09
Fix 'deadlock' on context timeout in gc_v3

sargun:
This fixes a situation where the top level GC times out and cancels the context, so the pipeline doesn't make progress but the GC describer aggregator is stuck aggregating enis

### DIFF
--- a/vpc/service/gc_v3.go
+++ b/vpc/service/gc_v3.go
@@ -386,8 +386,12 @@ func (vpcService *vpcService) doGCAttachedENIsLoop(ctx context.Context, protoIte
 }
 
 func (vpcService *vpcService) doGCENIs(ctx context.Context, item *regionAccount) error {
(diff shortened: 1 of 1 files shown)


vpc/service/gc_v3.go
```diff
@@ -508,25 +511,29 @@ ORDER BY RANDOM()
...
 	for {
 		select {
 		case <-ctx.Done():
-			return
+			return errors.Wrap(ctx.Err(), "Context error while waiting on eniWQ")
 		case eni, ok := <-eniWQ:
 			if !ok {
 				// The queue is complete
 				if len(enis) > 0 {
 					describeWQ <- enis
```
fabiokung, line 525: it looks like you also need the conditional channel write here to avoid deadlocking in this case as well

PR #395, opened 2020-05-11
Added a new ListTrunkToBranchENIMappings API

solarkennedy:
This new api provides a mapping from TrunkENI => BranchENI for
internal tools to correlate data about Titus containers
back to the actual EC2 instance with the TrunkENI on it.

----

Example output in json form:
```
{
  "trunkENIMapping": [
    {
      "trunkENI": "eni-0fcfec9dfaeeaa00c",
      "branchENI": "eni-0d80eb46c7984383d"
    },
    {
      "trunkENI": "eni-051c5f6b7ef0251fd",
      "branchENI": "eni-0049c023636cb282e"
    },
...
```

At first you might think that we would want a map of trunkeni -> repeated branchenis.
This seems to be harder than it sounds via protobufs. Also, I think in practice the consumer of this API will actually need the opposite mapping (they need to lookup which trunkeni given a branch). So I left this side of the system "simple" to just return basically what the DB has, and let the consumer do what they want.

This is my first gRPC thing ever, so feedback welcome.
Also I only manually tested this. It looks like there is room for integration tests, would you like me to spend time writing one for this?


vpc/service/branch_enis.go
```diff
@@ -1404,3 +1406,45 @@ func (vpcService *vpcService) createBranchENI(ctx context.Context, tx *sql.Tx, s
...
+
+	tx, err := vpcService.db.BeginTx(ctx, &sql.TxOptions{
+		ReadOnly: true,
+	})
+	if err != nil {
+		err = errors.Wrap(err, "Cannot start database transaction")
+		tracehelpers.SetStatus(err, span)
+		return nil, err
+	}
+	rows, err := tx.QueryContext(ctx, `
```
sargun, line 1424: Aren't you leaking the transaction here? You need to do defer tx.rollback, or tx.commit.

vpc/service/branch_enis.go
```diff
@@ -1404,3 +1406,45 @@ func (vpcService *vpcService) createBranchENI(ctx context.Context, tx *sql.Tx, s
...
+	SELECT
+	branch_eni,
+	trunk_eni
+	FROM assignments
+	JOIN branch_eni_attachments ON assignments.branch_eni_association = branch_eni_attachments.association_id
+`)
+	for rows.Next() {
```
sargun, line 1431: Should probably defer rows.Close.

vpc/service/branch_enis.go
```diff
@@ -1404,3 +1406,45 @@ func (vpcService *vpcService) createBranchENI(ctx context.Context, tx *sql.Tx, s
...
+		var branchENIId, trunkENIId string
+		err = rows.Scan(&branchENIId, &trunkENIId)
+		if err != nil {
+			err = errors.Wrap(err, "Error scanning for all trunk to branch ENI mappings")
+			tracehelpers.SetStatus(err, span)
+			return nil, err
+		}
+		logger.G(ctx).Debug(fmt.Sprintf("Returning trunk to branch eni map: %v => %v", trunkENIId, branchENIId))
```
sargun, line 1439: You should use with field, a la:
```
logger.G(ctx).WithField("trunkENI", trunkENIId).Debug("Returning trunk to branch ENI map")
```
(I think that this may be a little too verbose).

sargun:
Can you comment further on:

> This seems to be harder than it sounds via protobufs. Also, I think in practice the consumer of this API will actually need the opposite mapping (they need to lookup which trunkeni given a branch). So I left this side of the system "simple" to just return basically what the DB has, and let the consumer do what they want.

sargun:
I guess, why can't you do something like this:

```
message BranchENIAssociation {
    string branchENI = 1;
    uint32 idx = 2;
}

message ListTrunkToBranchENIMappingsResponse {
    // Map of trunk ENI ID => branch ENIs associated with it. 
    map<string, BranchENIAssociation> trunkENIMapping = 1;
}
```

solarkennedy:
> I guess, why can't you do something like this:
> 
> ```
> message BranchENIAssociation {
>     string branchENI = 1;
>     uint32 idx = 2;
> }
> 
> message ListTrunkToBranchENIMappingsResponse {
>     // Map of trunk ENI ID => branch ENIs associated with it. 
>     map<string, BranchENIAssociation> trunkENIMapping = 1;
> }
> ```

I was having a hard time wrapping my head around how to best represent this in protobufs. 
I've asked for clarification on which direction they actually want the mapping to go. I guess the idx is a way to avoid repeats?

Seems like maybe it should be this if we want to avoid repeats and make it easy to answer the question of "which trunk eni does this container belong to?":

```
message ListTrunkToBranchENIMappingsResponse {
    // Map of branch ENI ID => trunk ENIs associated with it. 
    map<string, string> trunkENIMapping = 1;
}
```

vpc/service/branch_enis.go
```diff
@@ -1404,3 +1406,45 @@ func (vpcService *vpcService) createBranchENI(ctx context.Context, tx *sql.Tx, s
...
+		err = errors.Wrap(err, "Cannot start database transaction")
+		tracehelpers.SetStatus(err, span)
+		return nil, err
+	}
+	rows, err := tx.QueryContext(ctx, `
+	SELECT
+	branch_eni,
+	trunk_eni
+	FROM assignments
+	JOIN branch_eni_attachments ON assignments.branch_eni_association = branch_eni_attachments.association_id
```
sargun, line 1429: Why isn't this:
```sql
SELECT branch_eni, trunk_eni FROM branch_eni_attachments WHERE state = 'attached'
```

solarkennedy: agreed, much simpler

vpc/service/branch_enis.go
solarkennedy: got it. Also caught in the new test

vpc/service/branch_enis.go
solarkennedy: Hmm, `mock.ExpectRollback()` doesn't seem to complain, even though I don't have those.

For this API, do I even really need a transaction?

sargun: I think you should have a txn to indicate this can be a read only snapshot query.

solarkennedy: It is currently read only, adding snapshot level isloation gives me:
`"Cannot start database transaction: pq: isolation level not supported: 5"`

How about RepeatableRead ?

sargun: err, When I meant snapshot, I didn't mean snapshot isolation, I meant that it was a read only query, so postgres could handle it as using a consistent cut / snapshot

solarkennedy: Ok I think I understand, the the fact that it is in a tx and is readonly is enough.

vpc/service/branch_enis.go
solarkennedy: It still ends up on one line, I've added `WithFields` so that it prints both

vpc/service/branch_enis.go
```diff
@@ -1404,3 +1406,64 @@ func (vpcService *vpcService) createBranchENI(ctx context.Context, tx *sql.Tx, s
...
+		ReadOnly:  true,
+		Isolation: sql.LevelRepeatableRead,
+	})
+	if err != nil {
+		err = errors.Wrap(err, "Cannot start database transaction")
+		tracehelpers.SetStatus(err, span)
+		return nil, err
+	}
+	defer func() {
+		if err != nil {
```
sargun, line 1426: Generally the approach in the code is:
```
defer func() {
   _ = tx.Rollback()
}()

... Codes ...

err = tx.Commit()
if err != nil {
   tracehelper...
   return err
}
```

If you tx rollback a txn that's already been committed it doesn't cause a problem (throws an error saying not in txn).

vpc/service/branch_enis.go
```diff
@@ -1404,3 +1406,64 @@ func (vpcService *vpcService) createBranchENI(ctx context.Context, tx *sql.Tx, s
...
+		err = tx.Commit()
+	}()
+
+	rows, err := tx.QueryContext(ctx, `
+	SELECT
+	branch_eni,
+	trunk_eni
+	FROM branch_eni_attachments WHERE state = 'attached'
+`)
+	defer rows.Close()
```
sargun, line 1438: This should be after the err check

solarkennedy:
> > I guess, why can't you do something like this:
> > ```
> > message BranchENIAssociation {
> >     string branchENI = 1;
> >     uint32 idx = 2;
> > }
> > 
> > message ListTrunkToBranchENIMappingsResponse {
> >     // Map of trunk ENI ID => branch ENIs associated with it. 
> >     map<string, BranchENIAssociation> trunkENIMapping = 1;
> > }
> > ```
> 
> I was having a hard time wrapping my head around how to best represent this in protobufs.
> I've asked for clarification on which direction they actually want the mapping to go. I guess the idx is a way to avoid repeats?
> 
> Seems like maybe it should be this if we want to avoid repeats and make it easy to answer the question of "which trunk eni does this container belong to?":
> 
> ```
> message ListTrunkToBranchENIMappingsResponse {
>     // Map of branch ENI ID => trunk ENIs associated with it. 
>     map<string, string> trunkENIMapping = 1;
> }
> ```

I've re-done this to just be a simple map after confirming exactly which direction we need this in to do the mapping in sonar.

vpc/service/branch_enis.go
solarkennedy: Why? Wouldn't we want to always close the rows object even on error? or is the concern that the close method will crash somehow if the rows object isn't legit in the first place?

sargun: QueryContext return (*Row, error). Given this, there's no guarantee that Row will be non-nil. See code:
https://github.com/lib/pq/blob/f3b22b2cad9e4567803b7ffd9853b8acda021438/conn.go#L830-L832

Often times when you have (interface{}, error) or (pointer,error) in golang, the pointer / interface is nil'd fi error != nil

vpc/proto/vpc.proto
```diff
@@ -356,6 +356,13 @@ message DetachBranchNetworkInterfaceResponse {
     string associationID = 3;
 }
 
+message ListBranchToTrunkENIMappingRequest {
+}
+
+message ListBranchToTrunkENIMappingResponse {
+    map<string, string> branchENIMapping = 1;
```
sargun, line 363: Might be work adding a comment saying that this is key of branch ENI -> trunk ENI.

vpc/service/branch_enis.go
solarkennedy: got it.

PR #396, opened 2020-05-21
Updated the ListBranchToTrunkENIMapping api to use the bigger titus api

solarkennedy:
This promotes the ListBranchToTrunkENIMapping to be "public" as part of the bigger titus-api-definitions (instead of the private vpcapi). After this change the new service.method is `com.netflix.titus.TitusAgentVPCInformationService.ListBranchToTrunkENIMapping` (tested manually with grpcurl).

To do this I've updated the go modules and vendor folder (not 100% I'm sure I did this correctly?).
I don't know why go.mod has `v0.0.1-rc9.0.20200520235959-0ab6f1129886`. The git sha is right, but the "tag" is wrong.

Otherwise the rest of the change is updating the references to point at the new imports / structs.


sargun:
You need to run:
```
go mod vendor
make protogen
make fmt
# This step is for validation:
make lint
```

solarkennedy:
> You need to run:
> 
> ```
> go mod vendor
> make protogen
> make fmt
> # This step is for validation:
> make lint
> ```

Yep. The only thing I didn't do was git-commit *all* the changes that `go mod vendor` did. (about 370 files on about a dozen modules, some added some removed). It seemed "wrong", but maybe it is collecting committed updates? Would you like me to go ahead and commit everything in this PR?

sargun:
Please update / commit `vendor/`

solarkennedy:
> Please update / commit `vendor/`

Added.

sargun:
lgtm

PR #398, opened 2020-05-29
Removed v2 versions of the titusvpcservice apis

solarkennedy:


sargun:
I think ProvisionInstanceRequestV2 may still be in use. 

Also, I think that there may be DB tables we can drop.

solarkennedy:
I think I've confirmed via logs that these APIs are no longer in use (as of a couple days ago).

PR #399, opened 2020-06-11
Titus oci wrapper

alekar:
Intercept OCI bundles and inject Netflix environment.


sargun:
Why isn't all of this logic server-side?

alekar:
It's in my notes to move this, along with the ENI allocation server side.
Q3.

alekar:
We're going to try and use a web hook on "pods/bind" before we go down this road.

PR #401, opened 2020-06-17
Check for user.stdio xattr for all filenames.

alekar:
Logerator now uses this attribute to indicate the active file it is
writing to. In order for the traditional rotate loop to recognize
this, it needs to check for the attribute on all files, not just
the file names written by tini.

### DIFF
--- a/filesystems/watcher.go
+++ b/filesystems/watcher.go
@@ -155,12 +155,14 @@ and then pessimistic expansion.
 
 // CheckFileForStdio determines whether the file at the path below is one written by tini as a stdio rotator
 func CheckFileForStdio(fileName string) bool {
-	if _, err := xattr.GetXattr(fileName, StdioAttr); err == xattr.ENOATTR {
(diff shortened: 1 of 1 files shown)


filesystems/watcher.go
```diff
@@ -155,12 +155,14 @@ and then pessimistic expansion.
...
 func CheckFileForStdio(fileName string) bool {
-	if _, err := xattr.GetXattr(fileName, StdioAttr); err == xattr.ENOATTR {
+	_, err := xattr.GetXattr(fileName, StdioAttr)
+
+	if err == xattr.ENOATTR {
 		return false
 	} else if os.IsNotExist(err) {
 		return false
 	} else if err != nil {
-		log.Errorf("Unable to fetch stdio attr from file %s because: %v", fileName, err)
```
sargun, line 163: Wont this spam errors?

alekar:
Most files should be `ENOATTR`, and not errors.

PR #402, opened 2020-06-25
Fix bug in call to CheckFileForStdio.

alekar:
When we build a recursive list of files for `traditionalRotate`, we need to pass the dirname to `shouldIgnore` so calls to `CheckFileForStdio` use the correct file name.

### DIFF
--- a/filesystems/watcher.go
+++ b/filesystems/watcher.go
@@ -649,28 +649,29 @@ func buildFileListInDir2(dirName string, fileList []string, checkModifiedTimeThr
 
 	for _, fileInfo := range fileInfos {
 		fqName := path.Join(dirName, fileInfo.Name())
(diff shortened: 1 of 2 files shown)


filesystems/watcher.go
```diff
@@ -688,7 +687,9 @@ func shouldIgnoreFile(fileInfo os.FileInfo, uploadThreshold time.Duration, check
 	}
 
 	// Do not "traditonally" "rotate" (really upload, and delete) stdio files
-	if CheckFileForStdio(fileInfo.Name()) {
+	qualifiedFileName := path.Join(dirName, fileInfo.Name())
+	if CheckFileForStdio(qualifiedFileName) {
+		log.Infof("ignoring %s, %s is set", qualifiedFileName, StdioAttr)
```
sargun, line 692: USe WithField. Also, wont this get noisy?

PR #403, opened 2020-07-02
Always set utf-8 encoding on plain text uploads to s3

solarkennedy:
I think this is a good idea?

Sometimes the stdout/err/logs of our tasks have non-ascii characters. Specifically, the titus web interface allows you to inserts things, and it encodes things as utf-8.

This change should make it so plain text files get uploaded with utf-8 encodings, so that 👍shows up correctly instead of `đ` or something.

### DIFF
--- a/uploader/s3.go
+++ b/uploader/s3.go
@@ -153,15 +153,24 @@ func (u *S3Backend) uploadFile(ctx context.Context, local io.Reader, remote stri
 		contentType = defaultS3ContentType
 	}
 
+	contentEncoding := ""
+	if contentType == 'text/plain' {
+		// On plain text files (default encoding), without setting this, it will be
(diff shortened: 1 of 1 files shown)


uploader/s3.go
```diff
@@ -153,15 +153,24 @@ func (u *S3Backend) uploadFile(ctx context.Context, local io.Reader, remote stri
 		contentType = defaultS3ContentType
 	}
 
+	contentEncoding := ""
+	if contentType == defaultS3ContentType {
```
sargun, line 157: Shouldn't this just be == "text/plain"?

solarkennedy: Yea I wasn't sure if I should reuse that constant or not.
But yea I do think that uft-8 does only make sense with text/plain, regardless of what the default content type is in the file.

PR #406, opened 2020-07-09
Remove CAP_NET_ADMIN capability

gabrielhartmann:
We need CAP_NET_ADMIN for setting up networking for VMs.  But we're not actively pursuing enabling VMs at the moment.  So let's remove the capability until we need or figure out a better way.


solarkennedy:
I vote no-ship, I don't think having this makes us too vulnerable.

PR #408, opened 2020-07-21
Remove Mesos Support

sargun:


vpc/tool/gc3/gc.go
```diff
@@ -52,8 +50,6 @@ func GC(ctx context.Context, timeout time.Duration, instanceIdentityProvider ide
 	switch args.SourceOfTruth {
```
solarkennedy, line 50: What do you think about removing switch statements like this? Second pass maybe, or keep in place for future nomad support?

sargun: I was thinking for leaving it when we add openstack support.

[Kidding, will remove in a second pass]

PR #409, opened 2020-07-24
add ccache for getSubnet

jigish:
### Description of the Change

Add ccache for `getSubnet` to reduce database load during large ENI attach load.

### DIFF
--- a/vpc/service/assign_addresses_v3.go
+++ b/vpc/service/assign_addresses_v3.go
@@ -33,8 +33,8 @@ const (
 	invalidAssociationIDNotFound = "InvalidAssociationID.NotFound"
 	assignTimeout                = 5 * time.Minute
 
-	getSubnetTimeout     = 1 * time.Minute
-	subnetExpirationTime = 1 * time.Minute
+	// The P99 for get subnet is 3 seconds
+	getSubnetTimeout = 10 * time.Second
 )
 
 var (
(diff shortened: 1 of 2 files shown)


vpc/service/assign_addresses_v3.go
```diff
@@ -104,6 +105,20 @@ func (vpcService *vpcService) getSubnet(ctx context.Context, az, accountID strin
 	ctx, span := trace.StartSpan(ctx, "getSubnet")
 	defer span.End()
 
+	item := vpcService.subnetCache.Get(fmt.Sprintf("az:%s accountID:%s subnetIDs:%s", az, accountID, strings.Join(subnetIDs, ",")))
```
sargun, line 108: How does the cache ever get populated?

jigish: yah sorry, forgot that -- added now

vpc/service/assign_addresses_v3.go
```diff
@@ -98,12 +103,30 @@ type branchENI struct {
...
+	item := vpcService.subnetCache.Get(subnetCacheKey(az, accountID, subnetIDs))
+	if item != nil {
+		span.AddAttributes(trace.BoolAttribute("cached", true))
+		if !item.Expired() {
+			span.AddAttributes(trace.BoolAttribute("expired", false))
+
+			val := item.Value().(*subnet)
+			return val, nil
+		}
+		span.AddAttributes(trace.BoolAttribute("expired", true))
```
sargun, line 125: If it's expired you should dispatch an ask to refresh the cache.

vpc/service/assign_addresses_v3.go
```diff
@@ -98,12 +103,30 @@ type branchENI struct {
 	idx           int
 }
 
+func subnetCacheKey(az, accountID string, subnetIDs []string) string {
+	return fmt.Sprintf("az:%s accountID:%s subnetIDs:%s", az, accountID, strings.Join(subnetIDs, ","))
```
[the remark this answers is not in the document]
jigish: fair point. sorted.

vpc/service/assign_addresses_v3.go
```diff
@@ -31,6 +33,9 @@ const (
 	invalidParameterValue        = "InvalidParameterValue"
 	invalidAssociationIDNotFound = "InvalidAssociationID.NotFound"
 	assignTimeout                = 5 * time.Minute
+
+	minSubnetExpirationTime = 5 * time.Minute
+	maxSubnetExpirationTime = 10 * time.Minute
```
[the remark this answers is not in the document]
sargun: I think 1 minute is better. and then what we should do is after 1 minute, asynchronously refresh the cache. After 10 minutes, and we've been unable to refresh the cache, then start throwing errors/ blocking.

vpc/service/assign_addresses_v3.go
```diff
@@ -98,27 +102,56 @@ type branchENI struct {
...
+		var staleSubnet *subnet
+		if item != nil {
+			span.AddAttributes(trace.BoolAttribute("cached", true))
+			span.AddAttributes(trace.BoolAttribute("expired", item.Expired()))
+			span.AddAttributes(trace.StringAttribute("expires", item.Expires().String()))
+
+			val := item.Value().(*subnet)
+			if !item.Expired() {
+				return val, nil
+			} else if item.Expires().Add(subnetStaleValidTime).After(time.Now()) {
```
solarkennedy, line 129: Can you double check this?

If we get an item and cache it at T=0 (minutes)
It expires at T=1
It is stale up to T=11 (expires.add)

If it is now T=15, should we execute line 130? Am I understanding "After" correctly? Oh, no, yea, because 11.After(15) is false.

Wait, I lost it. I'm not sure why the word doesn't work so well in my brain.  I got it again, whew. ! before(now) doesn't read any better to me, idk.

👍

jigish: `time.Now().Before(...)` would read better maybe?

solarkennedy: Oh yea :) 
All good though, I'm not that worried.

jigish:
@sargun @alekar this is ready for re-review now.

vpc/service/assign_addresses_v3.go
```diff
@@ -98,27 +102,56 @@ type branchENI struct {
...
+
 func (vpcService *vpcService) getSubnet(ctx context.Context, az, accountID string, subnetIDs []string) (*subnet, error) {
-	ctx, cancel := context.WithCancel(ctx)
-	defer cancel()
-	ctx, span := trace.StartSpan(ctx, "getSubnet")
-	defer span.End()
+	// NOTE[jigish] caching using the full set of subnetIDs will cause the same subnet to be returned for all hits
+	cacheKey := subnetCacheKey(az, accountID, subnetIDs)
+	r, err, _ := vpcService.getSubnetLock.Do(cacheKey, func() (interface{}, error) {
+		ctx, cancel := context.WithCancel(ctx)
```
sargun, line 114: I think you should have a span on the outside as well to indicate if multiple raced.

vpc/service/assign_addresses_v3.go
```diff
@@ -98,27 +102,56 @@ type branchENI struct {
...
+		if item != nil {
+			span.AddAttributes(trace.BoolAttribute("cached", true))
+			span.AddAttributes(trace.BoolAttribute("expired", item.Expired()))
+			span.AddAttributes(trace.StringAttribute("expires", item.Expires().String()))
+
+			val := item.Value().(*subnet)
+			if !item.Expired() {
+				return val, nil
+			} else if time.Now().Before(item.Expires().Add(subnetStaleValidTime)) {
+				staleSubnet = val
```
sargun, line 130: Thinking back to it now, I do not see any reason not to always serve stale subnets.

vpc/service/assign_addresses_v3.go
```diff
@@ -98,27 +102,56 @@ type branchENI struct {
...
-	})
-	if err != nil {
-		err = status.Error(codes.Unknown, errors.Wrap(err, "Could not start database transaction").Error())
-		span.SetStatus(traceStatusFromError(err))
-		return nil, err
-	}
-	defer func() {
-		_ = tx.Rollback()
-	}()
+		tx, err := vpcService.db.BeginTx(ctx, &sql.TxOptions{
```
sargun, line 136: Do we want to setup a timeout here?

vpc/service/assign_addresses_v3.go
```diff
@@ -98,27 +102,56 @@ type branchENI struct {
...
+}
+
 func (vpcService *vpcService) getSubnet(ctx context.Context, az, accountID string, subnetIDs []string) (*subnet, error) {
-	ctx, cancel := context.WithCancel(ctx)
-	defer cancel()
-	ctx, span := trace.StartSpan(ctx, "getSubnet")
-	defer span.End()
+	// NOTE[jigish] caching using the full set of subnetIDs will cause the same subnet to be returned for all hits
+	cacheKey := subnetCacheKey(az, accountID, subnetIDs)
+	r, err, _ := vpcService.getSubnetLock.Do(cacheKey, func() (interface{}, error) {
```
sargun, line 113: I'm thinking that the steps should be:
1. Check if the item is in the cache (and not expired) -- if it is not expired, return immediately
2. Move into a singleflight operation
3. In the single flight operation, check the cache _again_, and see if the subnet is in there (and not expired).
4. Create a (shorter) timeout and try to fetch / renew the subnets
5. If we failed to fetch / renew the subnets, then try to serve the expired version (always, and log a message + mark the span)

vpc/service/assign_addresses_v3.go
jigish: deleting subnets?

sargun: If we're unable to query the database, a subnet has been deleted, and we cannot restart the service, it's a bad day and we probably have other problems.

vpc/service/assign_addresses_v3.go
jigish: what do you feel like a better shorter timeout would be? the request timeout is 5 minutes -- so 1 minute?

vpc/service/assign_addresses_v3.go
jigish: added

vpc/service/assign_addresses_v3.go
sargun: Sure. There's no reason the process of fetching the subnet should be cancelled if the request gets cancelled.

vpc/service/assign_addresses_v3.go
jigish: ok removed the valid time check

vpc/service/assign_addresses_v3.go
jigish: timeout added as 1 minute

PR #411, opened 2020-07-24
Do not close readyCh multiple times

sargun:


vpc/service/branch_enis.go
```diff
@@ -1042,7 +1042,8 @@ type actionWorker struct {
 
 	pendingState string
 
-	readyCh chan struct{}
+	readyCh       chan struct{}
+	readyChClosed bool
```
jigish, line 1046: i'm assuming there are no concurrency concerns that would require a lock here?

sargun: Shouldn't.

PR #412, opened 2020-07-28
Move imds-cni to cmd/.

alekar:
Move imds cni under /cmd.


root/lib/systemd/system/titus-imds-proxy@.service
```diff
@@ -23,7 +23,7 @@ LimitNOFILE=65536
 PrivateTmp=yes
 
 EnvironmentFile=-/run/titus-metadata-proxy.salt
-EnvironmentFile=/var/lib/titus-environments/%i/imds-proxy.env
+EnvironmentFile=/var/lib/titus-environments/%i/titus-imds-proxy.env
```
solarkennedy, line 26: Will this be safe to deploy in a mutable way?

alekar: Should be, it is launched once-per-container.

PR #413, opened 2020-07-28
OCI wrapper

alekar:
This is intended to wrap runc, so we can inject environment into pods. It is modeled on oci-add-hooks.


solarkennedy:
I'm not sure if this is a great idea or awesome idea, but we could make this wrapper be the *new* titus-executor, with any magic entrypoint we need to get things done.

These env variables are the first one, but injecting ssh might be the second thing this could be useful for.

PR #414, opened 2020-07-29
Increase ulimit -n inside proxyd container.

alekar:
Allow service proxy more file descriptors.


andrew-leung:
I'm assuming there isn't a runc flag to set this and it needs to be done in the systemd unit?

PR #418, opened 2020-08-05
Fix situation where reader re-election can occur

sargun:


vpc/service/action_worker.go
```diff
@@ -122,10 +127,16 @@ func (actionWorker *actionWorker) loop(ctx context.Context, item keyedItem) erro
 					err = errors.Wrap(err, "Could not retrieve all work items")
 					return err
 				}
-				if !actionWorker.readyChClosed {
-					close(actionWorker.readyCh)
-					actionWorker.readyChClosed = true
-				}
+				actionWorker.readyCond.L.Lock()
+				actionWorker.ready = true
+				actionWorker.readyCond.L.Unlock()
```
solarkennedy, line 132: I don't really understand, shouldn't this stay Locked? Why unlock right away?

sargun: I added a comment in the struct trying to make this make more sense.

vpc/service/action_worker.go
```diff
@@ -173,7 +173,7 @@ func (actionWorker *actionWorker) worker(ctx context.Context, wq workqueue.RateL
 			err = errors.Wrap(err, "Cannot set lock timeout to 1000 milliseconds")
 			tracehelpers.SetStatus(err, span)
 			logger.G(ctx).WithError(err).Error()
-			return nil
+			return err
```
solarkennedy, line 195: Is this and the other `return err` the *actual* fix in this PR then?

sargun: And the Go upgrade. Where we would get into a rapid leader-reelection cycle because it'd keep picking up bad txns.

solarkennedy: Oh right right.

solarkennedy:
Nice!

PR #420, opened 2020-08-08
Consider subnets not found error as non-fatal, NotFound

sargun:
This error can happen in one of three conditions:
1. The agent's account / task's account / az being asked for has
   no subnets configured for it. This would be either caused
   by a misconfiguration on the Titus team's part, or the
   user's part
2. The data we fetch from AWS around subnets is wrong,
   and we've lost track of the subnets which we launch
   in.
3. Misconfiguration - Either the control plane, or the
   agent have misconfigured the default account / AZ / subnets
   for tasks.

Recently, we've seen an uptick in cases where the inputs
-Agent account
-Task requested account [optional]
-Task requested subnets [optional]

Have not aligned. We haven't seen grave misconfiguration of
our services, or the EC2 APIs return bad data lately.

We can alert on notfound as a % of total launches at a
much higher percentage if we want to keep track of
these errors still.

### Description of the Change

### DIFF
--- a/vpc/service/assign_addresses_v3.go
+++ b/vpc/service/assign_addresses_v3.go
@@ -158,11 +158,11 @@ LIMIT 1
 	err = row.Scan(&ret.az, &ret.vpcID, &ret.accountID, &ret.subnetID, &ret.cidr, &ret.region)
 	if err == sql.ErrNoRows {
 		if len(subnetIDs) == 0 {
-			err = fmt.Errorf("No subnet found matching IDs %s in az %s", subnetIDs, az)
+			err = newNotFoundError(fmt.Errorf("No subnet found matching IDs %s in az %s", subnetIDs, az))
 		} else {
-			err = fmt.Errorf("No subnet found in account %s in az %s", accountID, az)
+			err = newNotFoundError(fmt.Errorf("No subnet found in account %s in az %s", accountID, az))
 		}
-		span.SetStatus(traceStatusFromError(err))
+		tracehelpers.SetStatus(err, span)
 		// explicitly not returning stale subnet here
 		return nil, err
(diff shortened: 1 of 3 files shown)


gabrielhartmann:
unrelated: @sargun We never respect code coverage warnings.  Should we keep it on?  It's adding more noise than usual now.

vpc/service/assign_addresses_v3.go
```diff
@@ -158,11 +158,11 @@ LIMIT 1
 	err = row.Scan(&ret.az, &ret.vpcID, &ret.accountID, &ret.subnetID, &ret.cidr, &ret.region)
 	if err == sql.ErrNoRows {
 		if len(subnetIDs) == 0 {
-			err = fmt.Errorf("No subnet found matching IDs %s in az %s", subnetIDs, az)
+			err = newNotFoundError(fmt.Errorf("No subnet found matching IDs %s in az %s", subnetIDs, az))
```
alekar, line 161: Isn't subnetIDs empty? I suppose this is a bad request?

vpc/service/assign_addresses_v3.go
```diff
@@ -158,11 +158,11 @@ LIMIT 1
...
 		} else {
-			err = fmt.Errorf("No subnet found in account %s in az %s", accountID, az)
+			err = newNotFoundError(fmt.Errorf("No subnet found in account %s in az %s", accountID, az))
```
joshi-keyur, line 163: If we also log the subnetIDs as part of the error message, it could be useful for aggregation queries, etc?

sargun: Yea,  this is just a classification thing.

PR #421, opened 2020-08-11
CNI fixes

alekar:
Use containerInfo: we plan to deprecate admission webhook for now.


vpc/tool/cni/cni.go
```diff
@@ -2,18 +2,20 @@ package cni
...
 	"context"
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"net"
 	"os"
 	"strings"
 
+	"github.com/Netflix/titus-executor/api/netflix/titus"
+	"github.com/gogo/protobuf/proto"
```
sargun, line 13: is this the right protobuf library to use?

PR #427, opened 2020-08-21
Added cert-based authz to the titusvpcservice

solarkennedy:
The titusvpcservice currently does authn via certificates.

This change adds authz to validate the CN or DNS SAN from and
incoming cert matches titusagent. Currently it is a hard-coded
regex, but can expand to something more flexible in the future.


solarkennedy:
@sargun bump

vpc/service/auth.go
```diff
@@ -57,12 +58,37 @@ func (vpcService *titusVPCAgentServiceAuthFuncOverride) AuthFuncOverride(ctx con
...
 	if err != nil {
-		return nil, errors.Wrap(err, "Unable to verify client cert")
+		return err
 	}
-	return ctx, nil
+	// The cert.Verify function is relativly limited in what it can verify.
+	// It only can take in a single string and match it against DNS/CN.
+	// We know we will need to match against a variety of incoming names,
+	// so we check everything in a loop and return an error if nothing matched.
+	validCNs := []string{`titusagent\..*`}
```
sargun, line 81: Please make this a configuration value.

vpc/service/auth.go
```diff
@@ -57,12 +58,37 @@ func (vpcService *titusVPCAgentServiceAuthFuncOverride) AuthFuncOverride(ctx con
...
+	for _, validCn := range validCNs {
```
sargun, line 82: I think we should make it a config, and the regex(ex) should be compiled at startup time to validate them.

sargun (changes requested):

vpc/service/auth.go
solarkennedy: Added `SSL_VALID_CNS` and the flag for it.

I feel a little weird doing a csv? But in the end strings are kinda all we have for input config this way. Hopefully we will very rarely need to change this away from the default.

vpc/service/auth.go
solarkennedy: Oh yea good idea. Added.

vpc/service/auth.go
sargun: A CSV? What do you mean a CSV?

solarkennedy: comma separated value?

I know it is lame. Do you think it would be better to just have 1 regex only and leave it as an exercise to the reader to craft a regex that meets all their needs?

Writing this now it seems kinda obvious we should just have 1 regex and not an array of them...

sargun: Just add a command to combine regexes:

https://github.com/spadgos/regex-combiner#:~:text=Combines%20an%20array%20of%20regexes%20into%20a%20single%20regex.

sargun: (yes, there should be 1 regex)

cmd/titus-vpc-service/main.go
```diff
@@ -55,6 +55,7 @@ const (
 	sslPrivateKeyFlagName   = "ssl-private-key"
 	sslCAFlagName           = "ssl-ca"
 	sslTitusAgentCAFlagName = "ssl-titusagent-ca"
+	sslValidCNsFlagName     = "ssl-valid-cns"
```
sargun, line 58: `ssl-valid-cn-regex` probably makes more sense.

vpc/service/auth.go
```diff
@@ -57,12 +57,34 @@ func (vpcService *titusVPCAgentServiceAuthFuncOverride) AuthFuncOverride(ctx con
 	if err != nil {
 		return ctx, errors.Wrap(err, "Cannot parse remote certificate")
 	}
-	_, err = cert.Verify(x509.VerifyOptions{
+	err = vpcService.validateCert(cert)
+	if err != nil {
+		return nil, errors.Wrap(err, "Unable to verify client cert")
```
sargun, line 62: you can use fmt.errorf now we're on go1.13

vpc/service/auth.go
```diff
@@ -57,12 +57,34 @@ func (vpcService *titusVPCAgentServiceAuthFuncOverride) AuthFuncOverride(ctx con
...
+	if vpcService.ValidCNRegex.MatchString(cert.Subject.CommonName) {
+		return nil
+	}
+	for _, san := range cert.DNSNames {
+		if vpcService.ValidCNRegex.MatchString(san) {
+			return nil
+		}
+	}
+
+	return errors.Errorf("Certificate's CN: %s and SANS: %s failed to match our allow list: %s", cert.Subject.CommonName, cert.DNSNames, vpcService.ValidCNs)
```
sargun, line 89: Maybe %q would be better. Also, prefer fmt.errorf

vpc/service/service.go
```diff
@@ -248,6 +253,12 @@ func Run(ctx context.Context, config *Config) error {
 		return errors.New("The worker role must be non-empty")
 	}
 
+	validCNRegex, err := regexp.Compile(config.ValidCNs)
+	if err != nil {
+		panic(fmt.Errorf("Error parsing ValidCN Regexes: %w", err))
```
sargun, line 258: Any reason not to just return this error from here?

vpc/service/auth.go
solarkennedy: Hah gosh, I have trouble remembering what is best practice between wrap, `%w`, and fmt.error

vpc/service/service.go
solarkennedy: Nope

solarkennedy:
@sargun one more time?

vpc/service/auth.go
```diff
@@ -57,12 +57,34 @@ func (vpcService *titusVPCAgentServiceAuthFuncOverride) AuthFuncOverride(ctx con
...
+	if vpcService.ValidCNRegex.MatchString(cert.Subject.CommonName) {
+		return nil
+	}
+	for _, san := range cert.DNSNames {
+		if vpcService.ValidCNRegex.MatchString(san) {
+			return nil
+		}
+	}
+
+	return fmt.Errorf("Client certificate's CN: %q and SANS: %q failed to match our allow list: %v", cert.Subject.CommonName, cert.DNSNames, &vpcService.ValidCNRegex)
```
sargun, line 89: Dumb question, why is this a pointer?

sargun: Also, an aside, does this error get logged?

sargun (approved):

vpc/service/auth.go
solarkennedy: regex.compile returns a pointer to a struct.
When dereferenced like this it prints the actual regex instead of a hard to read struct representation.
I'll change it to just be `String()` to be more obvious.

It appears both client and server side. From vpctool:
```
FATA[0000] Failed                                        command=setup-instance error="rpc error: code = Unknown desc = Unable to verify client cert: Client certificate's CN: \"\" and SANS: [\"main01cell001.titusagent.dyntest.netflix.net\" \"main01cell001.titusagent.us-west-2.dyntest.netflix.net\" \"titusagent.dyntest.netflix.net\" \"titusagent.test.netflix.net\" \"titusagent.us-west-2.dyntest.netflix.net\" \"titusagent.us-west-2.test.netflix.net\" \"main01cell001.titusagent.us-west-2.test.netflix.net\" \"main01cell001.titusagent.test.netflix.net\" \"titusagent.app.us-west-2.test.cloud.netflix.net\"] failed to match our allow list: titusagentWRONG\\..*"
```

And from the vpcservice:
```
ERRO[0002] finished unary call with code Unknown         error="Unable to verify client cert: Client certificate's CN: \"\" and SANS: [\"main01cell001.titusagent.dyntest.netflix.net\" \"main01cell001.titusagent.us-west-2.dyntest.netflix.net\" \"titusagent.dyntest.netflix.net\" \"titusagent.test.netflix.net\" \"titusagent.us-west-2.dyntest.netflix.net\" \"titusagent.us-west-2.test.netflix.net\" \"main01cell001.titusagent.us-west-2.test.netflix.net\" \"main01cell001.titusagent.test.netflix.net\" \"titusagent.app.us-west-2.test.cloud.netflix.net\"] failed to match our allow list: titusagentWRONG\\..*" grpc.code=Unknown grpc.method=ProvisionInstanceV3 grpc.request.deadline="2020-10-07T18:25:47Z" grpc.service=com.netflix.titus.executor.vpc.TitusAgentVPCService grpc.start_time="2020-10-07T18:23:47Z" grpc.time_ms=0.302 origin=grpc peer.address="100.78.155.109:34696" span.kind=server system=grpc
```

Should be enough data to figure out who was connecting and why it failed.

solarkennedy:
Just waiting a bit on merging this in as there are some higher priority things I'm working on.

solarkennedy:
Rebased. I haven't forgotten about this PR. I'll be merging in the new year after the initial flood of instability settles down.

sargun:
Dumb question: In the 4 months since starting this, have any of the grpc proxies matured?

solarkennedy:
:) Maybe. But in this particular circumstance I don't think adding a proxy layer on top *just* for auth is worth it, given how simple our needs are (Today), with a simple regex.

cmd/titus-vpc-service/main.go
```diff
@@ -290,6 +292,7 @@ func main() {
 	rootCmd.Flags().String(sslCertFlagName, "", "The SSL Certificate")
 	rootCmd.Flags().String(sslCAFlagName, "", "General SSL CA")
 	rootCmd.Flags().String(sslTitusAgentCAFlagName, "", "Titus Agent CA")
+	rootCmd.Flags().String(sslValidCNRegexFlagName, `titusagent\..*`, "A regex representing valid CNs/SANS for incoming client certs")
```
JackOfMostTrades, line 295: Not sure if there's a different regex you were going to use rather than this default, but keep in mind that with this default anyone can deploy their own app with stack=titusagent to pass this regex since `${stack}.${appname}....` is one of the SANs the web certs gets.

vpc/service/auth.go
```diff
@@ -57,12 +57,34 @@ func (vpcService *titusVPCAgentServiceAuthFuncOverride) AuthFuncOverride(ctx con
...
 	})
 	if err != nil {
-		return nil, errors.Wrap(err, "Unable to verify client cert")
+		return err
 	}
-	return ctx, nil
+	// The cert.Verify function is relativly limited in what it can verify.
+	// It only can take in a single string and match it against DNS/CN.
+	// We know we will need to match against a variety of incoming names,
+	// so we check everything in a loop and return an error if nothing matched.
```
JackOfMostTrades, line 79: If you had a set of strings (rather than regexs) you could do cert.VerifyHostname(...) in a loop. Regexs are just dangerous because it can be really easy to footgun with them (see above comment about the `titusagent.` prefix) which I expect is why you don't see APIs for hostname verification that takes a regex.

vpc/service/test_certs/Makefile
```diff
@@ -0,0 +1,19 @@
...
+
+test-leaf-expired.crt: test-leaf-expired.csr test-CA.key test-CA.pem
+	# Note on OSX you must have a local copy of the openssl binary, not in /usr/bin for this to work
+	faketime '-356d' /bin/bash -c 'openssl x509 -req -in test-leaf-expired.csr -CA test-CA.pem -CAkey test-CA.key -CAcreateserial -out $@ -days 1 -sha256'
+
+
+test-dnsname.csr: dnsname.csr.txt test-key.key
+	openssl req -new -nodes -key test-key.key -out $@ -config dnsname.csr.txt
+test-dnsname.crt: test-dnsname.csr test-CA.key test-CA.pem
+	openssl x509 -req -in test-dnsname.csr -CA test-CA.pem -CAkey test-CA.key -CAcreateserial -out $@ -days 100000 -sha256 -extensions 'v3_req' -extfile dnsname.csr.txt
```
JackOfMostTrades, line 19: No real problem doing it this way, but FWIW I like to generate certs on the fly for tests. Saves me the panic of seeing private keys checked in. ;) Also saves me the heartache of discovering tests that randomly start failing 1 year later (though it looks like you got the -days 100000 covered). Example: https://stash.corp.netflix.com/projects/PS/repos/metatron-family/browse/bootstrap-go/certificateserver-test/mock_certificate_server.go#107

cmd/titus-vpc-service/main.go
solarkennedy: Heh, I was actually aware of that when I wrote it, I just figured it was close enough

solarkennedy:
I'm going to close this for now. It may be a future improvement for the HN team to implement.

PR #430, opened 2020-08-27
Regenerated titus-vpc-service migration files

solarkennedy:
Also added a make target to make it easier to remember how to do that.


Makefile
```diff
@@ -138,3 +138,7 @@ clean-proto-defs: | $(clean)
 $(GOBIN_TOOL):
 	go get github.com/myitcv/gobin
 	go install github.com/myitcv/gobin
+
+vpc/service/db/migrations/bindata.go: vpc/service/db/migrations/generate.go vpc/service/db/migrations/*.sql
+	go generate ./vpc/service/db/migrations/
+	go fmt ./vpc/service/db/migrations/*.go
```
sargun, line 144: goimports -w

Makefile
```diff
@@ -138,3 +138,7 @@ clean-proto-defs: | $(clean)
 $(GOBIN_TOOL):
 	go get github.com/myitcv/gobin
 	go install github.com/myitcv/gobin
+
+vpc/service/db/migrations/bindata.go: vpc/service/db/migrations/generate.go vpc/service/db/migrations/*.sql
```
sargun, line 142: Do you want to make `build` rely on this target?

solarkennedy: Wasn't super sure. It doesn't change that often, but if it isn't in the build target, then who is going to "remember" to run it? :) 

Make should do its thing. I'll add it.

PR #431, opened 2020-08-27
Jumbo frames

sargun:
[Begin to] Add jumbo frames support


vpc/service/assign_addresses_v3.go
```diff
@@ -607,6 +608,26 @@ func (vpcService *vpcService) AssignIPV3(ctx context.Context, req *vpcapi.Assign
 	return vpcService.assignIPsToENI(ctx, req, ass, maxIPAddresses, instance, trunkENI)
 }
 
+func addJumboRoute(ctx context.Context, req *vpcapi.AssignIPRequestV3, ass *vpcapi.AssignIPResponseV3) {
```
solarkennedy, line 611: how about just `addDefaultRoute` ? (the name of the function implies that it always does jumbo

vpc/tool/container2/setup_container_linux.go
```diff
@@ -145,35 +145,73 @@ func addIPv4AddressAndRoute(ctx context.Context, nsHandle *netlink.Handle, link
...
-		newRoute.MTU = *mtu
-	}
-	err = nsHandle.RouteAdd(&newRoute)
-	if err != nil {
-		logger.G(ctx).WithError(err).Error("Unable to add route to link")
-		return errors.Wrap(err, "Unable to add default route to link")
+	var defaultRouteFound bool
+	if len(routes) > 0 {
+		for _, route := range routes {
+			if route.Family != vpcapi.AssignIPResponseV3_Route_IPv4 {
```
solarkennedy, line 151: What else could this possibly be? V6?

vpc/tool/container2/setup_container_linux.go
```diff
@@ -145,35 +145,73 @@ func addIPv4AddressAndRoute(ctx context.Context, nsHandle *netlink.Handle, link
...
+				MTU:       int(route.Mtu),
+			}
+			err = nsHandle.RouteAdd(&newRoute)
+			if err != nil {
+				logger.G(ctx).WithField("route", route).WithError(err).Error("Unable to add route to link")
+				return fmt.Errorf("Unable to add route %v to link due to: %w", route, err)
+			}
+		}
+		if !defaultRouteFound {
+			logger.G(ctx).Warn("Did not install default route. Suspicious.")
```
solarkennedy, line 177: Did not install ipv4 default route right?

vpc/tool/container2/setup_container_linux.go
sargun: In theory, we could add v6 support, but I haven't added custom MTU routes to V6 yet

vpc/tool/container2/setup_container_linux.go
```diff
@@ -127,7 +127,7 @@ func DoSetupContainer(ctx context.Context, netnsfd int, bandwidth, ceil uint64,
 	return configureLink(ctx, nsHandle, newLink, bandwidth, ceil, mtu, allocation)
 }
 
-func addIPv4AddressAndRoute(ctx context.Context, nsHandle *netlink.Handle, link netlink.Link, mtu *int, address *vpcapi.UsableAddress) error {
+func addIPv4AddressAndRoute(ctx context.Context, nsHandle *netlink.Handle, link netlink.Link, mtu *int, address *vpcapi.UsableAddress, routes []*vpcapi.AssignIPResponseV3_Route) error {
```
solarkennedy, line 130: We already have the (link) mtu here, do we need the whole routes[] complexity, or can we get away with a routeMtu int to use here?

In fact, it looked like previously if `mtu != nil` we *were* setting the mtu on the default route. Could we just start actually passing the mtu into this function if jumbo?


I guess it looks like this PR is 2 things: jumbo support and extending the api to handle multiple additional routes. Is the extra routes something we will eventually need?

vpc/tool/container2/setup_container_linux.go
sargun: Yeah. We never install an IPv6 default route. We rely on SLAAC for that

vpc/tool/container2/setup_container_linux.go
sargun: I guess this PR is really moving the jumbo frame logic to the server, from the individual agents.

sargun: If the server side doesn't send down Jumbo routes, it'll fall back to the local settings. Otherwise, (hopefully soon) we can add logic to have a greenlist on the titus vpc service about what other VPCs / subnets to send jumbo traffic to.

vpc/service/assign_addresses_v3.go
sargun: Done

PR #433, opened 2020-08-28
Optimize SQL query in assignarbitraryipv4address

sargun:
This should make our median assignprivate IP time go down from 1.1 seconds to...much less


solarkennedy:
Neat!

PR #435, opened 2020-09-02
Fix max session length

sargun:


solarkennedy (approved):
Can you give some more context on this change? Did something change/break?

sargun:
@solarkennedy it's in the commit comments:
```
 * Change the session length limit to 64, see:
   https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRole.html
 * Change the regexp to not unnecessarily replace - with _
```

They changed the limit from 32 to 64, and changed the allowable regexp.

PR #437, opened 2020-09-04
Optimize query 2

sargun:
It turns out testing on your laptop is never quite as good as testing in production.

[Please see commit messages for more information]

### DIFF
--- a/vpc/service/assign_addresses_v3.go
+++ b/vpc/service/assign_addresses_v3.go
@@ -1064,7 +1064,7 @@ func assignArbitraryIPv6AddressV3(ctx context.Context, tx *sql.Tx, branchENI *ec
 	if unusedIPAddresses.Len() > 0 {
 		unusedIPv6AddressesList := unusedIPAddresses.List()
 
-		rows, err := tx.QueryContext(ctx, "SELECT ip_address, last_seen FROM ip_last_used_v3 WHERE host(ip_address) = any($1) AND vpc_id = $2", pq.Array(unusedIPv6AddressesList), aws.StringValue(branchENI.VpcId))
+		rows, err := tx.QueryContext(ctx, "SELECT ip_address, last_seen FROM ip_last_used_v3 WHERE ip_address = any($1::inet[]) AND vpc_id = $2", pq.Array(unusedIPv6AddressesList), aws.StringValue(branchENI.VpcId))
 		if err != nil {
 			err = status.Error(codes.Unknown, errors.Wrap(err, "Could not fetch utilized IPv4 addresses from the database").Error())
 			span.SetStatus(traceStatusFromError(err))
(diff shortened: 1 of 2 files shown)


vpc/service/assign_addresses_v3.go
```diff
@@ -1061,40 +1060,24 @@ func assignArbitraryIPv6AddressV3(ctx context.Context, tx *sql.Tx, branchENI *ec
...
 
-	if unusedIPAddresses.Len() > 0 {
+	if l := unusedIPAddresses.Len(); l > 0 {
 		unusedIPv6AddressesList := unusedIPAddresses.List()
 
-		rows, err := tx.QueryContext(ctx, "SELECT ip_address, last_seen FROM ip_last_used_v3 WHERE host(ip_address) = any($1) AND vpc_id = $2", pq.Array(unusedIPv6AddressesList), aws.StringValue(branchENI.VpcId))
-		if err != nil {
-			err = status.Error(codes.Unknown, errors.Wrap(err, "Could not fetch utilized IPv4 addresses from the database").Error())
-			span.SetStatus(traceStatusFromError(err))
+		row := tx.QueryRowContext(ctx, "SELECT ip_address FROM ip_last_used_v3 WHERE ip_address = any($1::inet[]) AND vpc_id = $2 ORDER BY last_seen ASC LIMIT 1", pq.Array(unusedIPv6AddressesList), aws.StringValue(branchENI.VpcId))
```
solarkennedy, line 1066: Just to confirm, this query picks the oldest ipv6 out of the database that has been last_seen, that happens to be already assigned to this agent, just not used by any task?

vpc/service/assign_addresses_v3.go
```diff
@@ -1061,40 +1060,24 @@ func assignArbitraryIPv6AddressV3(ctx context.Context, tx *sql.Tx, branchENI *ec
...
+		var ipAddress string
+		err = row.Scan(&ipAddress)
+		if err == sql.ErrNoRows {
+			// Effectively choose a random one.
+			ipAddress = unusedIPv6AddressesList[rand.Intn(l)]
```
solarkennedy, line 1071: And if there are no rows, that means it is safe to pick one from the list we already have?

vpc/service/assign_addresses_v3.go
```diff
@@ -1061,40 +1060,24 @@ func assignArbitraryIPv6AddressV3(ctx context.Context, tx *sql.Tx, branchENI *ec
...
+		} else if err != nil {
+			err = errors.Wrap(err, "Could not fetch utilized IPv6 addresses from the database")
+			tracehelpers.SetStatus(err, span)
 			return nil, err
 		}
```
solarkennedy, line 1076: And I guess there is the unspoken else, that there *is* an old ipv6 that was in our local unused list that was also previously seen, and that was already assigned to `ipAddress` during the scan.

solarkennedy (approved):
If my comments are correct in my understanding (I think I already get the `host()` optimization) then shipit!

Great job!

vpc/service/assign_addresses_v3.go
sargun: Correct.

vpc/service/assign_addresses_v3.go
sargun: Yes

sargun: Well, one at random.

vpc/service/assign_addresses_v3.go
sargun: Well, if err == nil, then the scan was successful, and we successfully read the right IP.

PR #438, opened 2020-09-07
add abmetrix system service

nabilschear:
### Description of the Change

using adminlogs and service mesh as an example, adding the capability to launch the abmetrix testing service.  I can cut and paste like the best of em.  But, I don't have the ability to test this out.  Also, need to figure out where the container image gets specified.


sargun:
@nabilschear I'm trying to understand. Is there a case where we would want to enable / disable abmetrix on an image-by-image basis? If not, it might be better to take the approach the SSH image has where it's always enabled.

nabilschear:
I saw that metatron also had some flags to turn on and off.  So, I figured that abmetrix was in a similar state.  Also, there's a lot of layers configs and overrides.  I'm not exactly sure where it is different than SSHD/metatron.  The only place where they seem to diverge is in types.go where there's extra params.

solarkennedy:
It looks like there is a big toggle as a command line flag for the executor, then a task attribute toggle. 

I don't mind having containers require to opt in at first, gain confidence that the thing works, then make it opt out (like sshd).

solarkennedy (approved):

solarkennedy:
I'm also curious about the failure mode if this container fails to start up. Does that halt a task?
Is there any way we can make this container "nice to have"? and never be able to crash a task?

nabilschear:
@solarkennedy wouldn't systemd try a few times and then give up?

executor/runtime/docker/docker_linux.go
```diff
@@ -97,6 +97,12 @@ var systemServices = []serviceOpts{
 		required:     true,
 		enabledCheck: shouldStartServiceMesh,
 	},
+	{
+		humanName:    "abmetrix",
+		unitName:     "titus-abmetrix",
+		required:     true,
```
solarkennedy, line 103: Can you make this `false`, so that if the systemd unit doesn't work for *any* reason, we never fail a task?
Esp in these early days. I don't think any failed launch of abmetric should fail a task.

solarkennedy (changes requested):

executor/runtime/docker/docker_linux.go
nabilschear: done

nabilschear:
what's the next step for testing this out?

sargun:
@nabilschear So, I think that you should get a dev stack and test it there. Nonetheless, I would make a couple changes to not make this controlled by container params and just make an executor-wide configurable.

nabilschear:
i got into some git trouble, but I think I've got it implemented like sshd now.

executor/runtime/docker/docker_linux.go
nabilschear: i changed this back to true to match what was going on in sshd. thoughts @solarkennedy @sargun ?

solarkennedy: I still say that this is a nice to have, and should never be in the critical path of a task launching.
I don't want a bad push to this container to cause tasks to fail, so therefore I think it should be `required: false`.

(I dare say sshd should be the same! But not on this PR)

nabilschear:
i'm still going to experiment with this a little and see whether we can make it more optional.

root/lib/systemd/system/titus-abmetrix@.service
```diff
@@ -0,0 +1,16 @@
...
+ConditionPathIsDirectory=/var/lib/titus-inits/%i/ns
+
+# If the service restarts more than 10 times in 30 seconds, let it die
+StartLimitIntervalSec=30
+StartLimitBurst=10
+
+[Service]
+EnvironmentFile=/var/lib/titus-environments/%i.env
+# Run as root (UID 0, GID 0) and with CAP_DAC_OVERRIDE so that containers with a `USER` instruction work
+ExecStart=/usr/bin/runc --root /var/run/docker/runtime-${TITUS_OCI_RUNTIME}/moby exec --user 0:0 --cap CAP_DAC_OVERRIDE --cwd /titus/abmetrix ${TITUS_CONTAINER_ID} /titus/abmetrix/start
```
solarkennedy, line 12: Can you confirm this process is actually long-lived, and doesn't fork?

solarkennedy (approved):

root/lib/systemd/system/titus-abmetrix@.service
nabilschear: yeah, it keeps running to refresh data.  Doesn't fork.

config/config.go
```diff
@@ -42,6 +42,10 @@ type Config struct {
 	ContainerLogViewer    bool
 	LogViewerServiceImage string
 
+	// Enable abmetrix service
+	ContainerAbmetrixEnabled bool
+	AbmetrixServiceImage     string
```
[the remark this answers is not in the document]
sargun: This may simplify things a little.

config/config.go
```diff
@@ -156,6 +160,16 @@ func NewConfig() (*Config, []cli.Flag) {
 			Destination: &cfg.LogViewerServiceImage,
 			EnvVar:      "LOGVIEWER_SERVICE_IMAGE",
 		},
+		cli.BoolTFlag{
+			Name:        "container-abmetrix",
+			Destination: &cfg.ContainerAbmetrixEnabled,
+			EnvVar:      "CONTAINER_ABMETRIX",
+		},
+		cli.StringFlag{
```
sargun, line 168: See suggestion above.

executor/runtime/docker/docker_linux.go
```diff
@@ -97,6 +97,14 @@ var systemServices = []serviceOpts{
 		required:     true,
 		enabledCheck: shouldStartServiceMesh,
 	},
+	{
+		humanName: "abmetrix",
+		unitName:  "titus-abmetrix",
+		required:  true,
```
sargun, line 103: Let's make required: false for now? And we can look at logs to see if there are any problems?

sargun (changes requested):

executor/runtime/docker/docker_linux.go
nabilschear: i'm not entirely sure what setting required=false means here.  But I did it =)

config/config.go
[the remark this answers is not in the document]
nabilschear: ok, added a shouldstart function that checks the enabled var and if the image var is not null.

executor/runtime/types/types.go
```diff
@@ -30,6 +30,8 @@ const (
 	assignIPv6AddressParam       = "titusParameter.agent.assignIPv6Address"
 	serviceMeshEnabledParam      = "titusParameter.agent.service.serviceMesh.enabled"
 	serviceMeshContainerParam    = "titusParameter.agent.service.serviceMesh.container"
+	abmetrixEnabledParam         = "titusParameter.agent.service.abmetrix.enabled"
```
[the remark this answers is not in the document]
nabilschear: no more container params

config/config.go
nabilschear: this is still here as an agent option

executor/runtime/docker/docker_linux.go
nabilschear: i set to false.  not entirely sure what that means, but the other logic should turn it off if it doesn't have sufficient config to start up

executor/runtime/docker/docker_linux.go
solarkennedy: When `required: false` is here, when the executor does `systemctl start titus-abmetrix`, and it doesn't work, the task continues to launch. That way, if we make a mistake in the image or the binary fails to start for whatever reason, customers are not impacted.

nabilschear:
ok @sargun and @solarkennedy, i think this is ready to go now.  I've tested the following conditions that all do what we want with this current build:

- if there's no abmetrix container image in config.sh, task will still launch without abmetrix
- if there is a container image but CONTAINER_ABMETRIX env var is set to false, task will still launch without abmetrix
- if there is a container image but the abmetrix crashes on startup, task will still launch with dead systems abmetrix service
- and obviously, if there is a container image (and optionally CONTAINER_ABMETRIX is true too), then abmetrix will start up and run correctly.

sargun (approved):

sargun (approved):

sargun:
LGTM.

solarkennedy (approved):

PR #439, opened 2020-09-08
Move to relying on Titus VPC Service to send us routes

sargun:
This finishes wiring up jumbo frames end-to-end, but it means that
it relies on the Titus VPC Service to send down routes to the
agent with MTU-per-route.

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -64,7 +64,6 @@ const (
 	defaultRunTmpFsSize     = "134217728" // 128 MiB
 	defaultRunLockTmpFsSize = "5242880"   // 5 MiB: the default setting on Ubuntu Xenial
 	trueString              = "true"
-	jumboFrameParam         = "titusParameter.agent.allowNetworkJumbo"
 	accountID               = "titusParameter.agent.accountId"
 	subnets                 = "titusParameter.agent.subnets"
 	elasticIPPool           = "titusParameter.agent.elasticIPPool"
@@ -749,6 +748,12 @@ func prepareNetworkDriver(parentCtx context.Context, cfg Config, c *runtimeTypes
 	}
 	if assignIPv6Address {
 		args = append(args, "--assign-ipv6-address=true")
+	}
+
+	if jumbo, err := c.Jumbo(); err != nil {
+		log.WithError(err).Error("Could not parse value for jumbo frame parameter")
+	} else if jumbo {
+		args = append(args, "--jumbo=true")
 	}
 
 	// We intentionally don't use context here, because context only KILLs.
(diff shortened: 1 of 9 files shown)


vpc/tool/container2/setup_container_linux.go
```diff
@@ -225,13 +208,13 @@ func addIPv4AddressAndRoute(ctx context.Context, nsHandle *netlink.Handle, link
 	err = nsHandle.RouteDel(&oldLinkRoute)
 	if err != nil {
 		logger.G(ctx).WithError(err).Error("Unable to delete route on link")
-		return errors.Wrap(err, "Unable to add delete existing 'link' route to link")
+		return 0, errors.Wrap(err, "Unable to add delete existing 'link' route to link")
 	}
 
-	return nil
+	return mtu, nil
```
solarkennedy, line 214: This return of mtu from this function confuses me a little bit.
I get it though, this is exactly where the actual default route mtu is calculated, and needs to be consumed later on.

But would a function called `getDefaultRouteMtu()` (to be injected into setupHTBClasses) that does this make the code any simpler?

executor/runtime/docker/docker.go
```diff
@@ -1881,13 +1886,12 @@ func setupNetworkingArgs(burst bool, c *runtimeTypes.Container) []string {
...
-	if jumbo, ok := c.TitusInfo.GetPassthroughAttributes()[jumboFrameParam]; ok {
-		if val, err := strconv.ParseBool(jumbo); err != nil {
-			log.Error("Could not parse value for "+jumboFrameParam+": ", err)
-		} else if val {
-			args = append(args, "--jumbo=true")
-		}
+	if jumbo, err := c.Jumbo(); err != nil {
+		log.WithError(err).Error("Could not parse value for jumbo frame parameter")
+	} else if jumbo {
+		args = append(args, "--jumbo=true")
```
solarkennedy, line 1892: So we have to have both things in place here:
* task attribute to allow jumbo
* New version of vpc service that sends routes with mtus

Anything else?

I'm trying to wrap my head around deployment strategy and potential failure modes between the components.

solarkennedy (approved):

vpc/tool/container2/setup_container_linux.go
sargun: I feel like then we're writing the code to see if a default route exists & returning the MTU in two different places, and that seems a little weird to me?

executor/runtime/docker/docker.go
sargun: The task attribute to allow jumbo has been a thing for a long time, but yeah, the thing that goes out "first" now is the VPC service that sends down routes. Earlier this was written with the idea that the agent would go out first before the VPC Service was taught to send down any routes.

PR #440, opened 2020-09-08
Remove github annotations from code coverage checks

sargun:


sargun:
@gabrielhartmann This quiets down the excess noise from code coverage

solarkennedy:
Geeze, thank you!

PR #441, opened 2020-09-08
Consult resource limits for cgroup construction

gabrielhartmann:
Additionally consult the pod not containerinfo for network limit
specification


sargun:
What is the canonical value for networkMbps? The annotation, or the resources value? It looks like we set the environment variables based on the annotation:

```
func getResourceEnv(pod *corev1.Pod) map[string]string {
        networkAnnotation := pod.GetAnnotations()[IngressBandwidthAnnotation]
        network, _ := resource.ParseQuantity(networkAnnotation)
        firstConainer := pod.Spec.Containers[0]
        numCPUs := firstConainer.Resources.Limits.Cpu()
        memory := firstConainer.Resources.Limits.Memory()
        disk := firstConainer.Resources.Limits["titus/disk"] // NOTE(manas) this is specified in MB
        return map[string]string{
                "TITUS_NUM_CPU":               fmt.Sprintf("%d", numCPUs.Value()),
                "TITUS_NUM_MEM":               fmt.Sprintf("%d", memory.Value()/(1024*1024)),
                "TITUS_NUM_DISK":              fmt.Sprintf("%d", disk.Value()),
                "TITUS_NUM_NETWORK_BANDWIDTH": fmt.Sprintf("%d", network.Value()/(1e6)),
        }
}
```

gabrielhartmann:
> What is the canonical value for networkMbps? The annotation, or the resources value? It looks like we set the environment variables based on the annotation:
> 
> ```
> func getResourceEnv(pod *corev1.Pod) map[string]string {
>         networkAnnotation := pod.GetAnnotations()[IngressBandwidthAnnotation]
>         network, _ := resource.ParseQuantity(networkAnnotation)
>         firstConainer := pod.Spec.Containers[0]
>         numCPUs := firstConainer.Resources.Limits.Cpu()
>         memory := firstConainer.Resources.Limits.Memory()
>         disk := firstConainer.Resources.Limits["titus/disk"] // NOTE(manas) this is specified in MB
>         return map[string]string{
>                 "TITUS_NUM_CPU":               fmt.Sprintf("%d", numCPUs.Value()),
>                 "TITUS_NUM_MEM":               fmt.Sprintf("%d", memory.Value()/(1024*1024)),
>                 "TITUS_NUM_DISK":              fmt.Sprintf("%d", disk.Value()),
>                 "TITUS_NUM_NETWORK_BANDWIDTH": fmt.Sprintf("%d", network.Value()/(1e6)),
>         }
> }
> ```
I don't wish to address the code above that is in a separate project.  If you look closely it is dividing the Network value by 1,000,000.  Elsewhere in that code, the network value is scaled up by 1,000,000.  All this takes places in annotations and doesn't affect the specification of limits, requested when it comes to resources.

Today the unit of interest from the `titus-executor` perspective is in Mbps.
https://github.com/Netflix/titus-executor/blob/master/api/netflix/titus/agent.pb.go#L1196

If a user requests for example 700 "network thingies" in the JobDescriptor, they get a value of "700" in the `titus/network` request/limit resource fields of the container.  To reiterate, no transformation of the user specified network resource value occurs today where its specification in native pod fields, or Job Descriptor fields are concerned.

rostyboost:
Thank you @gabrielhartmann !!

executor/runner/runner.go
```diff
@@ -104,7 +105,7 @@ func WithRuntime(ctx context.Context, m metrics.Reporter, rp RuntimeProvider, cf
 }
 
 // StartTask can be called once to start a task, by a given Runner
-func (r *Runner) StartTask(taskID string, titusInfo *titus.ContainerInfo, mem int64, cpu int64, gpu int64, disk uint64) error {
+func (r *Runner) StartTask(taskID string, titusInfo *titus.ContainerInfo, mem int64, cpu int64, gpu int64, disk uint64, network uint64) error {
```
sargun, line 108: If we intend to keep this around much longer, IMHO we should replace ContainerInfo with v1.Pod, and the bits to extract mem / cpu / gpu / disk / network should happen *in* here. Nothing to do with this PR specifically, but there's a bunch of stuff that could be extracted at the pod level that's more straightforward, helping us remove stuff from containerinfo.

executor/runtime/types/types.go
```diff
@@ -532,6 +532,7 @@ type Resources struct {
 	CPU       int64
 	GPU       int64
 	Disk      uint64
+	Network   uint64
 	HostPorts []uint16
```
sargun, line 536: O.O

sargun (approved):
LGTM. yay for ripping stuff out of containerinfo.

sargun:
I think this was slightly confusing to me because it said "cgroup construction" -- what it really is is consult pod's resource limits in order to determine the resources granted by the executor.

PR #442, opened 2020-09-08
Added a scheduled task to prune old unused ips

solarkennedy:
---

I guess I didn't ask before starting, but is simply removing old entries in this table all this needs to do?

I don't know that much about postgres transactions and consistency. Are there any locking concerns that I should be aware of? (with other things that might be trying to get stuff from this table. No right? We are using a real db here).

I didn't test this on any live infra (only SELECT to validate the query), as it is destructive. Would you like me to do more with regards to testing?

### DIFF
--- a/vpc/service/prune_unused_ips.go
+++ b/vpc/service/prune_unused_ips.go
@@ -1,0 +1,56 @@
+package service
+
+import (
+	"context"
+	"database/sql"
+
+	"github.com/Netflix/titus-executor/logger"
+	"github.com/Netflix/titus-executor/vpc/tracehelpers"
+	"github.com/pkg/errors"
+	"go.opencensus.io/trace"
+)
+
+const (
+	//  Each Titus task requires an IPv4 address. Each VPC has 3 x /16s for Titus to operate in.
+	// That means that the maximum number of "interesting" IPv4 rows per table are 3*(2**16) = 196608
+	howManyUnusedIPsToKeep = 200000
+)
+
+func (vpcService *vpcService) pruneLastUsedIPAddresses(ctx context.Context, null keyedItem, tx *sql.Tx) (retErr error) {
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+	ctx, span := trace.StartSpan(ctx, "pruneLastUsedIPAddresses")
+	defer span.End()
+
+	logger.G(ctx).Debug("Beginning purge of last used ip addresses")
+	result, err := tx.ExecContext(ctx,
+		`WITH addresses_by_family AS
+		    (SELECT id,
+				last_seen,
+				RANK() OVER (PARTITION BY family(ip_address), vpc_id
+							 ORDER BY last_seen DESC) AS rank
+		    FROM ip_last_used_v3
+		    WHERE last_seen != TIMESTAMP 'EPOCH')
+	    DELETE
+	    FROM ip_last_used_v3
+	    WHERE id IN
+		    (SELECT id
+		    FROM addresses_by_family
+		    WHERE rank > $1 AND last_seen < now() - INTERVAL '1 DAY')`,
+		howManyUnusedIPsToKeep,
+	)
+	if err != nil {
+		err = errors.Wrap(err, "Failed to delete unused ip adddres from ip_last_used_v3s")
+		tracehelpers.SetStatus(err, span)
+		return err
+	}
+
+	rows, err := result.RowsAffected()
+	if err != nil {
+		err = errors.Wrap(err, "Failed to load the result from the DELETE query to ip_last_used_v3s")
+		tracehelpers.SetStatus(err, span)
+		return err
+	}
+	logger.G(ctx).WithField("deleted_ips", rows).Infof("Deleted ip addresses (left the oldest %d)", howManyUnusedIPsToKeep)
+	return nil
+}
--- a/vpc/service/service.go
+++ b/vpc/service/service.go
@@ -477,6 +477,11 @@ func (vpcService *vpcService) getTaskLoops() []taskLoop {
 			itemLister: vpcService.getRegionAccounts,
 			workFunc:   vpcService.reconcileAvailabilityZonesRegionAccount,
 		},
+		{
+			taskName:   "last_used_ip_pruner",
+			itemLister: nilItemEnumerator,
+			workFunc:   vpcService.pruneLastUsedIPAddresses,
+		},
 	}
 }
 


solarkennedy:
@sargun Ok I think I got the sql the way you want. I still made it select by vpc because the tasks are split out per-vpc anyway. It makes the sql more complex, but it makes it easier to reason about what each task is operating over?

Or maybe we don't need to fan out the tasks per vpc at all?

vpc/service/prune_unused_ips.go
```diff
@@ -0,0 +1,109 @@
...
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+	ctx, span := trace.StartSpan(ctx, "pruneLastUsedIPAddresses")
+	defer span.End()
+
+	logger.G(ctx).WithFields(map[string]interface{}{
+		"vpcID": vpcID.String,
+	}).Info("Beginning purge of last used ip addresses")
+
+	result, err := tx.ExecContext(ctx,
```
sargun, line 70: Is there a reason to have this scheduled task run on a per-vpc basis, versus using the nilitem and running on scheduled task periodically and doing the delete with a slightly more complex SQL query?

solarkennedy: I don't know. I was mostly going off of your original advice to make the tasks fan out on a per-vpc basis.

vpc/service/prune_unused_ips.go
```diff
@@ -0,0 +1,109 @@
...
+		if err != nil {
+			return nil, err
+		}
+		ret = append(ret, &vid)
+	}
+	_ = tx.Commit()
+	return ret, nil
+}
+
+type vpcID struct {
```
sargun, line 43: Any reason this isn't just type vpcId string?

vpc/service/prune_unused_ips.go
```diff
@@ -0,0 +1,109 @@
...
+
+func (v *vpcID) String() string {
+	return fmt.Sprintf(v.vpcID)
+}
+
+func (vpcService *vpcService) pruneLastUsedIPAddresses(ctx context.Context, vpcID keyedItem, tx *sql.Tx) (retErr error) {
+
+	//  Each Titus task requires an IPv4 address. Each VPC has 3 x /16s for Titus to operate in.
+	// That means that the maximum number of "interesting" IPv4 rows per table are 3*(2**16) = 196608
+	howManyToKeep := 200000
```
sargun, line 59: this should probably be a const somewhere

vpc/service/prune_unused_ips.go
```diff
@@ -0,0 +1,109 @@
...
+				last_seen,
+				RANK() OVER (PARTITION BY family(ip_address)
+							 ORDER BY last_seen DESC) AS rank
+		FROM ip_last_used_v3 WHERE AND vpc_id = '$2;)
+	    DELETE
+	    FROM ip_last_used_v3
+	    WHERE id IN
+		  (SELECT id
+		   FROM addresses_by_family_in_vpc
+		   WHERE rank > $1 AND last_seen > now() - INTERVAL '30 DAY')
```
sargun, line 82: I think you can probably crank this down to something like 1 day.

vpc/service/prune_unused_ips.go
```diff
@@ -0,0 +1,109 @@
...
+		"vpcID": vpcID.String,
+	}).Info("Beginning purge of last used ip addresses")
+
+	result, err := tx.ExecContext(ctx,
+		`WITH addresses_by_family_in_vpc AS
+		(SELECT id,
+				last_seen,
+				RANK() OVER (PARTITION BY family(ip_address)
+							 ORDER BY last_seen DESC) AS rank
+		FROM ip_last_used_v3 WHERE AND vpc_id = '$2;)
```
sargun, line 76: no seimicolon

vpc/service/prune_unused_ips.go
```diff
@@ -0,0 +1,109 @@
...
+	defer cancel()
+	ctx, span := trace.StartSpan(ctx, "pruneLastUsedIPAddresses")
+	defer span.End()
+
+	logger.G(ctx).WithFields(map[string]interface{}{
+		"vpcID": vpcID.String,
+	}).Info("Beginning purge of last used ip addresses")
+
+	result, err := tx.ExecContext(ctx,
+		`WITH addresses_by_family_in_vpc AS
```
sargun, line 71: btw, I find https://sqlformat.org/ helpful

vpc/service/prune_unused_ips.go
sargun, line 76: is that quoting correct?

sargun, line 76: It should be something like:
```
WITH addresses_by_family AS
  (SELECT id,
          last_seen,
          RANK() OVER (PARTITION BY family(ip_address)                       
                       ORDER BY last_seen DESC) AS rank
   FROM ip_last_used_v3 
WHERE vpc_id = $1 AND last_seen != TIMESTAMP 'EPOCH'
)
DELETE
FROM ip_last_used_v3
WHERE id IN
    (SELECT id
     FROM addresses_by_family
     WHERE rank > $2 AND last_seen > now() - INTERVAL '1 DAY');
```

I think?

sargun (changes requested):

vpc/service/prune_unused_ips.go
```diff
@@ -0,0 +1,109 @@
...
+		return err
+	}
+
+	rows, err := result.RowsAffected()
+	if err != nil {
+		err = errors.Wrap(err, "Failed to load the result")
+		return err
+	}
+
+	logger.G(ctx).WithFields(map[string]interface{}{
```
sargun, line 98: If rows > 0; you may want to add an analyze statement here.

solarkennedy: Why though? Wouldn't it be pretty normal to remove rows?

sargun: Eh, I guess we can look at it later.

vpc/service/prune_unused_ips.go
solarkennedy: I'm going with nilitem

solarkennedy:
This has some updates from last night and now doesn't do per-vpc. Ready for second round of review.

vpc/service/prune_unused_ips.go
```diff
@@ -0,0 +1,64 @@
...
+				RANK() OVER (PARTITION BY family(ip_address), vpc_id
+							 ORDER BY last_seen DESC) AS rank
+		    FROM ip_last_used_v3
+		    WHERE last_seen != TIMESTAMP 'EPOCH')
+	    DELETE
+	    FROM ip_last_used_v3
+	    WHERE id IN
+		    (SELECT id
+		    FROM addresses_by_family
+		    WHERE rank > $1 AND last_seen > now() - INTERVAL '1 DAY')`,
```
sargun, line 41: shouldn't this be:
`last_seen < now() - INTERVAL '1 DAY'`?

vpc/service/prune_unused_ips.go
```diff
@@ -0,0 +1,64 @@
...
+		howManyUnusedIPsToKeep,
+	)
+	if err != nil {
+		err = errors.Wrap(err, "Failed to delete unused ip adddress")
+		return err
+	}
+
+	rows, err := result.RowsAffected()
+	if err != nil {
+		err = errors.Wrap(err, "Failed to load the result")
```
sargun, line 51: Please user tracehelpers to set the trace status on error.

sargun (changes requested):

vpc/service/prune_unused_ips.go
solarkennedy: Whoops! I updated your jira comment but didn't update the code... :(

vpc/service/prune_unused_ips.go
```diff
@@ -0,0 +1,56 @@
...
+		return err
+	}
+
+	rows, err := result.RowsAffected()
+	if err != nil {
+		err = errors.Wrap(err, "Failed to load the result from the DELETE query to ip_last_used_v3s")
+		tracehelpers.SetStatus(err, span)
+		return err
+	}
+	logger.G(ctx).Infof("Deleted %d ip addresses (left the oldest %d)", rows, howManyUnusedIPsToKeep)
```
sargun, line 54: Please use withfields here.

sargun: or WithField() on the # or rows.

sargun (approved):
One small nit, and then we're good to go!

vpc/service/prune_unused_ips.go
solarkennedy: added

sargun (approved):

PR #444, opened 2020-09-11
Make it so that an ENI must not already have an existing cross account permission

sargun:
This is added because now in a given region, we have two pools of agents being managed. When CreateNetworkInterfacePermission is used, it cannot be used to give multiple accounts permissions. This helps us deal with that...

### DIFF
--- a/vpc/service/generate_assignment_id.go
+++ b/vpc/service/generate_assignment_id.go
@@ -691,13 +691,27 @@ func (vpcService *vpcService) getUnattachedBranchENIV3(ctx context.Context, fast
 	var eni branchENI
 
 	row := fastTx.QueryRowContext(ctx, `
-SELECT branch_enis.branch_eni, az, account_id  
-FROM branch_enis 
-WHERE 
-subnet_id = $1 AND 
-(SELECT state FROM branch_eni_attachments WHERE branch_eni = branch_enis.branch_eni AND state IN ('attaching', 'attached', 'unattaching')) IS NULL
+SELECT branch_enis.branch_eni,
+       branch_enis.az,
+       branch_enis.account_id
+FROM branch_enis
+LEFT JOIN eni_permissions ON branch_enis.branch_eni = eni_permissions.branch_eni
+WHERE branch_enis.subnet_id = $1
+ -- Either the branch eni and trunk eni must be in the same account, or there must be an existing eni permission for
+-- the account the trunk eni is in, or there must be no other network permissions
+
+  AND (eni_permissions.account_id IS NULL
+       OR eni_permissions.account_id = $2
+       OR branch_enis.account_id = $2)
+  AND
+    (SELECT state
+     FROM branch_eni_attachments
+     WHERE branch_eni = branch_enis.branch_eni
+       AND state IN ('attaching',
+                     'attached',
+                     'unattaching')) IS NULL
 ORDER BY RANDOM()
-LIMIT 1`, req.subnet.subnetID)
+LIMIT 1`, req.subnet.subnetID, req.trunkENIAccount)
 	err := row.Scan(&eni.id, &eni.az, &eni.accountID)
 	if err == nil {
 		span.AddAttributes(
--- a/vpc/service/branch_enis.go
+++ b/vpc/service/branch_enis.go
@@ -550,6 +550,11 @@ func (vpcService *vpcService) ensureBranchENIPermissionV3(ctx context.Context, t
 	ctx, span := trace.StartSpan(ctx, "ensureBranchENIPermissionV3")
 	defer span.End()
 
+	span.AddAttributes(
+		trace.StringAttribute("branch", eni.id),
+		trace.StringAttribute("branchAccount", eni.accountID),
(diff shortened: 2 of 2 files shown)


jigish (approved):

vpc/service/generate_assignment_id.go
```diff
@@ -691,13 +691,27 @@ func (vpcService *vpcService) getUnattachedBranchENIV3(ctx context.Context, fast
...
+-- the account the trunk eni is in, or there must be no other network permissions
+
+  AND (eni_permissions.account_id IS NULL
+       OR eni_permissions.account_id = $2
+       OR branch_enis.account_id = $2)
+  AND
+    (SELECT state
+     FROM branch_eni_attachments
+     WHERE branch_eni = branch_enis.branch_eni
+       AND state IN ('attaching',
```
solarkennedy, line 710: I realize this was the same before, but why are we looking at these states? I thought we were looking for *unattached* enis? Or is this what makes them "warm" ?

solarkennedy (approved):
I think I understand, but this query is a tad over my head

vpc/service/generate_assignment_id.go
sargun: We do this because it makes life easier for the query planner. This is the table structure:
```
                                                Table "public.branch_eni_attachments"
          Column           |            Type             | Collation | Nullable |                      Default                       
---------------------------+-----------------------------+-----------+----------+----------------------------------------------------
 id                        | integer                     |           | not null | nextval('branch_eni_attachments_id_seq'::regclass)
 branch_eni                | text                        |           | not null | 
 trunk_eni                 | text                        |           |          | 
 idx                       | integer                     |           |          | 
 association_id            | text                        |           |          | 
 created_at                | timestamp without time zone |           | not null | CURRENT_TIMESTAMP
 attachment_generation     | integer                     |           |          | 
 association_token         | text                        |           |          | 
 disassociation_token      | text                        |           |          | 
 attachment_created_by     | text                        |           |          | 
 attachment_created_at     | timestamp without time zone |           |          | now()
 attachment_completed_by   | text                        |           |          | 
 attachment_completed_at   | timestamp without time zone |           |          | 
 unattachment_created_by   | text                        |           |          | 
 unattachment_created_at   | timestamp without time zone |           |          | 
 unattachment_completed_by | text                        |           |          | 
 unattachment_completed_at | timestamp without time zone |           |          | 
 state                     | attachment_state            |           | not null | 'attached'::attachment_state
 force                     | boolean                     |           |          | false
 error_code                | text                        |           |          | 
 error_message             | text                        |           |          | 
Indexes:
    "branch_eni_attachments_pk" PRIMARY KEY, btree (id)
    "branch_eni_attachments_association_id_uindex" UNIQUE, btree (association_id)
    "branch_eni_attachments_association_token_uindex" UNIQUE, btree (association_token)
    "branch_eni_attachments_branch_eni_trunk_eni_idx_uindex" UNIQUE, btree (branch_eni, trunk_eni, idx) WHERE state = 'attaching'::attachment_state OR state = 'attached'::attachment_state OR state = 'unattaching'::attachment_state
    "branch_eni_attachments_branch_eni_uindex" UNIQUE, btree (branch_eni) WHERE state = 'attaching'::attachment_state OR state = 'attached'::attachment_state OR state = 'unattaching'::attachment_state
    "branch_eni_attachments_disassociation_token_uindex" UNIQUE, btree (disassociation_token)
    "branch_eni_attachments_trunk_eni_idx_uindex" UNIQUE, btree (trunk_eni, idx) WHERE state = 'attaching'::attachment_state OR state = 'attached'::attachment_state OR state = 'unattaching'::attachment_state
    "branch_eni_attachments_branch_eni_state_index" btree (branch_eni, state)
    "branch_eni_attachments_state_index" btree (state)
    "branch_eni_attachments_trunk_eni_index" btree (trunk_eni)
Check constraints:
    "branch_eni_attachments_association_id_check" CHECK (association_id IS NULL OR association_id ~~ 'trunk-assoc-%'::text)
    "branch_eni_attachments_branch_eni_check" CHECK (branch_eni ~~ 'eni-%'::text)
    "branch_eni_attachments_trunk_eni_check" CHECK (trunk_eni ~~ 'eni-%'::text)
Foreign-key constraints:
    "branch_eni_attachments_branch_enis_branch_eni_fk" FOREIGN KEY (branch_eni) REFERENCES branch_enis(branch_eni) ON DELETE CASCADE
Referenced by:
    TABLE "assignments" CONSTRAINT "assignments_branch_eni_attachments_association_id_fk" FOREIGN KEY (branch_eni_association) REFERENCES branch_eni_attachments(association_id) ON DELETE CASCADE
```


If you look at the branch_eni_attachments_trunk_eni_idx_uindex, branch_eni_attachments_branch_eni_uindex, and branch_eni_attachments_branch_eni_trunk_eni_idx_uindex indices they are these states. This allows the query planner to make the assertion to perform an index only scan as opposed to having to look directly at the values.

sargun: The change is to basically make sure there's not an existing permission on the branch ENI.

PR #445, opened 2020-09-11
Do dns resolution lookups for NFS mounts at mount time

solarkennedy:
Previously, the titus-executor would do the dns lookup
for an NFS mount on behalf of a container, and pass
that addr into the mount syscall.

However, sometimes the dns name for an EFS (Amazon NFS)
endpoint won't even resolve unless you are in the
correct network namespace.

This change defers the lookup of the IP for NFS
until the last second, while we are inside the network
namespace of the container.

To get around the limitations of statically linked binaries
and glibc+libnss, we switch to use musl-gcc, which is fine
for such a small binary.

----

I would like to test this end-to-end on a real stack and real mount to verify it actually does what we expect.
I copy/pasted getaddr functions from mount.nfs4 directly, so they should match the status quo behavior of what normal tools would do if you just had this in fstab or whatever.

### DIFF
--- a/mount/Makefile
+++ b/mount/Makefile
@@ -1,3 +1,2 @@
 titus-mount: mount.c
-	gcc -D_GNU_SOURCE=1 -std=gnu11 -Wall -Werror -static -g -o titus-mount mount.c
-
+	musl-gcc -std=gnu11 -Wall -static -g -o titus-mount mount.c
--- a/mount/mount.c
+++ b/mount/mount.c
@@ -1,6 +1,13 @@
+#define _GNU_SOURCE
+
 #include <stdlib.h>
 #include <stdio.h>
 #include <errno.h>
+
+/* getaddrinfo */
+#include <arpa/inet.h>
+#include <netdb.h>
+#include <string.h>
 
 /* setns */
 #include <sched.h>
@@ -23,6 +30,30 @@ static char* get_fs_type() {
 	return fs_type;
 }
 
+static int dns_lookup(const char *hostname, struct sockaddr_in *addr)
+{
+	struct hostent *hp;
+	addr->sin_family = AF_INET;
+
+	if (inet_aton(hostname, &addr->sin_addr)) {
+		fprintf(stderr, "titus-mount: %s is already an IP. Not doing a DNS lookup\n", hostname);
+		return 0;
+	}
+	fprintf(stderr, "titus-mount: Decoding and resolving dns hostname for %s\n", hostname);
+	hp = gethostbyname(hostname);
+	if (hp == NULL) {
+		int err_ret = h_errno;
+		fprintf(stderr, "titus-mount: can't get address for %s: %s\n", hostname, hstrerror(err_ret));
+		return -1;
+	}
+	if (hp->h_length > (int)sizeof(struct in_addr)) {
+		fprintf(stderr, "titus-mount: got bad hp->h_length");
+		return -1;
+	}
+	memcpy(&addr->sin_addr, hp->h_addr, hp->h_length);
+	return 0;
+}
+
 int main() {
 	int mnt_ns_fd, net_ns_fd;
 	unsigned long flags_ul;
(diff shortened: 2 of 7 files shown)


mount/mount.c
```diff
@@ -23,6 +30,42 @@ static char* get_fs_type() {
...
 }
 
+static int fill_ipv4_sockaddr(const char *hostname, struct sockaddr_in *addr)
+{
+	struct hostent *hp;
+	addr->sin_family = AF_INET;
+
+	if (inet_aton(hostname, &addr->sin_addr))
+		return 0;
+	if ((hp = gethostbyname(hostname)) == NULL) {
```
sargun, line 40: Please put assignment on a different line than comparison.

mount/mount.c
```diff
@@ -23,6 +30,42 @@ static char* get_fs_type() {
 	return fs_type;
 }
 
+static int fill_ipv4_sockaddr(const char *hostname, struct sockaddr_in *addr)
```
sargun, line 33: I think either a comment, or maybe renaming the function that this is supposed to resolve the destination address would be valuable.

mount/mount.c
```diff
@@ -23,6 +30,42 @@ static char* get_fs_type() {
...
+	memcpy(&addr->sin_addr, hp->h_addr, hp->h_length);
+	return 0;
+}
+
+static int get_my_ipv4addr(char *ip_addr, int len)
+{
+	char myname[1024];
+	struct sockaddr_in myaddr;
+
+	if (gethostname(myname, sizeof(myname))) {
```
sargun, line 58: I think the original approach / implementation of using a UDP socket was better. It took routing into account.

mount/mount.c
```diff
@@ -98,8 +145,25 @@ int main() {
 		}
 	}
 
+	/* For NFS, we must do the dns resolution *here* while we are inside net ns */
+	if (nfs_mount_hostname) {
+		printf("Decoding and resolving dns hostname");
+		static struct sockaddr_in server_addr;
+		static char ip_addr[16] = "127.0.0.1";
```
sargun, line 152: Should this be initialized at all?

mount/Makefile
```diff
@@ -1,3 +1,2 @@
 titus-mount: mount.c
-	gcc -D_GNU_SOURCE=1 -std=gnu11 -Wall -Werror -static -g -o titus-mount mount.c
-
+	musl-gcc -std=gnu11 -Wall -static -g -o titus-mount mount.c
```
sargun, line 2: Does musl have the same behaviour of glibc as to reading /etc/resolv.conf on the first query?

sargun (changes requested):

mount/mount.c
solarkennedy: Doing the UDP socket to establish the source local ip?
Why do we really need that? Won't the kernel pick the right ip (take routing into account) when establishing the connection anyway?

mount/mount.c
solarkennedy: I guess it is a little defensive against something going wrong and having it crash.
I guess in our case crashing would probably be better than weird behavior where it tries to connect to localhost.

mount/Makefile
solarkennedy: I don't know, but this is the only dns query this code does, should be fine right?
There are some documented differences, but nothing about resolv.conf: https://wiki.musl-libc.org/functional-differences-from-glibc.html#Name-Resolver/DNS

You did mention there was something I could add that I could test, what was that again?

mount/mount.c
solarkennedy: Yea honestly I think this code is more complex than we need.
I'll try to simplify this.

solarkennedy:
I've confirmed that this code (`0a2b8ee`) does make our efs integration tests pass on real servers and real nfs mounts!

mount/Makefile
sargun: musl doesn't implement the low level libresolv API. Reading the musl code, it looks like the behaviour is the same, where they load resolv.conf on the first query

mount/mount.c
```diff
@@ -35,11 +66,15 @@ int main() {
...
 	const char *net_ns = getenv("NET_NS");
 	const char *source = getenv("MOUNT_SOURCE");
+	const char *nfs_mount_hostname = getenv("MOUNT_NFS_HOSTNAME");
 	const char *target = getenv("MOUNT_TARGET");
 	const char *flags = getenv("MOUNT_FLAGS");
 	const char *options = getenv("MOUNT_OPTIONS");
 	const char *fs_type = get_fs_type();
 
+	char final_options [200];
+	strcpy(final_options, options);
```
sargun, line 76: Maybe use strncpy and add an assertion, or make this buffer something impossibly large, like 4096 (PAGE_SIZE)?

sargun (approved):
Some minor comments, but lgtm.

PR #446, opened 2020-09-14
Use apt to install titus executor

sargun:


hack/test-images/metatron/Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM ubuntu:bionic as builder
 
 COPY --from=golang:1.13-stretch /usr/local/go /usr/local/go
-RUN apt-get update && apt-get --no-install-recommends install -y busybox-static
+RUN apt-get update && apt-get install busybox-static
```
solarkennedy, line 4: I'm surprised this works without -y ?

PR #448, opened 2020-09-15
Fix fetching assignment

sargun:
This fixes a small error / typo with fetchAssignment. It didn't happen until we had machines in different subnets than the ENI itself.


rgulewich (approved):

solarkennedy:
Do you prefer the fmt.error over wrapping?

solarkennedy (approved):

sargun:
@solarkennedy fmt.error is the "new" way in Go 1.13. and the "Right" way in go 1.13+

PR #449, opened 2020-09-16
Bump tinit to include new seccomp fd handler

solarkennedy:
First time doing a submodule bump here. Seems fine?

The `tini/src` make target does it anyway I think.

### DIFF
--- a/hack/builder/Dockerfile
+++ b/hack/builder/Dockerfile
@@ -1,8 +1,16 @@
 FROM ubuntu:bionic
 
+# titus-agent-ci includes debs for ci-related stuff
+# Specifically the linux-libc-dev package was downloaded from
+# https://packages.debian.org/experimental/linux-libc-dev
+# and then manually uploaded to https://packagecloud.io/netflix/titus-agent-ci
+RUN DEBIAN_FRONTEND=noninteractive apt-get update && apt-get -y install curl && \
+    curl https://packagecloud.io/install/repositories/netflix/titus-agent-ci/script.deb.sh | bash
+
 RUN export DEBIAN_FRONTEND=noninteractive && apt-get update && \
     apt-get install -y build-essential make cmake g++ gcc libc6-dev pkg-config \
-        libattr1-dev git curl wget jq ruby ruby-dev rubygems lintian unzip bison flex clang llvm musl-tools && \
+        libattr1-dev git curl wget jq ruby ruby-dev rubygems lintian unzip bison flex clang llvm musl-tools \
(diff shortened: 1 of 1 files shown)


sargun:
I think you need newer kernel headers in the builder containers.

solarkennedy:
Ah. Added.

sargun:
@solarkennedy Can we keep this on a branch, and set this branch to upload to packagecloud? And then when v5.9 comes out, we can pickup the headers from v5.9?

sargun:
or if this only needs 5.8 headers, please source them from the PPA: https://kernel.ubuntu.com/~kernel-ppa/mainline/v5.8/

solarkennedy:
> or if this only needs 5.8 headers, please source them from the PPA: https://kernel.ubuntu.com/~kernel-ppa/mainline/v5.8/

Is your concern using debs that have no lineage? Or that they are not mainline?
That PPA above isn't "real" in the sense that it isn't a legit apt repo. I would have to scrape and fetch the deb.

Or I could save the debs and put them in our package cloud? Then we could depend on them. That would mean we are using mainline debs, but still from our package cloud. Which do you prefer?

sargun:
@solarkennedy Err, I have no problem with using debs. I think my (primary) concern comes from using debs made by us vs. debs made by canonical inc. IIRC, the "Kernel Mainline" project is already a thing: https://wiki.ubuntu.com/Kernel/MainlineBuilds

My concern is that "Who will make them the next time around". Versus, if we use the kernel mainline project, we can just track that.

solarkennedy:
> @solarkennedy Err, I have no problem with using debs. I think my (primary) concern comes from using debs made by us vs. debs made by canonical inc. IIRC, the "Kernel Mainline" project is already a thing: https://wiki.ubuntu.com/Kernel/MainlineBuilds
> 
> My concern is that "Who will make them the next time around". Versus, if we use the kernel mainline project, we can just track that.

Got it. How about I save the mainline kernel debs in our packagecloud ci repo and use them?

sargun:
That works for me. 

One alternative:

```
ppa:canonical-kernel-team/ppa 
```

And then whenever an HWE kernel drops (which should be in the next ~1-2 months), we move to that?

solarkennedy:
Got it. I originally thought this was the linux headers we needed, but it is actually `linux-libc-dev`.
That wasn't pre-built anywhere convenient except for in debian's repos for 5.9.

solarkennedy:
Bumped now to include the 3fd version, ready for real testing.

sargun (approved):

PR #450, opened 2020-09-21
WIP: allow mounting arbitrary NFS hosts

rgulewich:
This is *not* intended to be merged - it's just a hack to allow for testing.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -8,6 +8,7 @@ steps:
     branches:
       - master
       - multi-account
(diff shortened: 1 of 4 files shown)


solarkennedy:
I think we should totally make it so if it is an efs id, we turn it into an efs hostname (not ip) and otherwise we let it pass as a normal hostname!

solarkennedy:
@rgulewich closing as this code is now in master, no need for this branch anymore (you can specify the NFS hostname directly in the EFS ID)

PR #452, opened 2020-09-23
Fix seccomp profile

sargun:
This is to fix the seccomp profile rule that was put in place for the CVE.


solarkennedy:
I see the new test, but I don't see any actual change to the profile?

sargun:
@solarkennedy WIP.

PR #453, opened 2020-09-24
Accept resources passed from the control plane in bytes

rgulewich:
The standard kubernetes way of representing resources is in bytes, as per the pod spec. Our kubelet implementation already uses bytes for some resources, but not others. In order to ease the migration from VK to K with kubescheduler, we've decided to do this migration now and ensure all components (control plane, VK, and K) are consistent.

See also: Netflix/titus-control-plane#928

### DIFF
--- a/vk/backend/backend.go
+++ b/vk/backend/backend.go
@@ -7,12 +7,14 @@ import (
 	"fmt"
 	"os"
 	"os/signal"
+	"strconv"
 	"sync"
 	"time"
 
 	"github.com/Netflix/titus-executor/api/netflix/titus"
 	titusdriver "github.com/Netflix/titus-executor/executor/drivers"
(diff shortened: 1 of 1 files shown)


vk/backend/backend.go
```diff
@@ -101,6 +139,11 @@ func RunWithBackend(ctx context.Context, runner *runner.Runner, statuses *os.Fil
 		return errors.Wrap(err, "Could not deserialize protobuf")
 	}
 
+	useBytes, err := useByteUnits(pod)
```
alekar, line 142: Won't this error if the annotation is missing? Do you want this call to return `numBytes, useBytes, err`?

alekar: Doh. My bad.

alekar (approved):

PR #455, opened 2020-09-28
Removed references to TITUS_REGISTRY

solarkennedy:
I'm trying to remove all references to `TITUS_REGISTRY`, I think it is confusing and we don't need it anymore.

We already set `DOCKER_REGISTRY` in `/etc/titus-executor/config.sh`, I don't want it overridden.


sargun:
Do we even need this file? It seems like it was for running the executor as a Mesos executor.

solarkennedy:
> Do we even need this file? It seems like it was for running the executor as a Mesos executor.

Sounds legit.

sargun (approved):

PR #456, opened 2020-09-30
WIP

solarkennedy:
WIP


mount/mount.c
```diff
@@ -54,6 +58,26 @@ static int dns_lookup(const char *hostname, struct sockaddr_in *addr)
 	return 0;
 }
 
+static int fsopen(const char *fsname, unsigned int flags) {
```
sargun, line 61: this should be long.

mount/mount.c
```diff
@@ -149,8 +173,14 @@ int main() {
 		fprintf(stderr, "titus-mount: using these nfs mount options: %s\n", final_options);
 	}
 
-	/* We don't check for overflow */
-	rc = mount(source, target, fs_type, flags_ul, final_options);
+	if (nfs_mount_hostname && check_new_mount_api_available()) {
```
sargun, line 176: why check nfs_mount_hostname?

mount/mount.c
```diff
@@ -54,6 +58,26 @@ static int dns_lookup(const char *hostname, struct sockaddr_in *addr)
 	return 0;
 }
 
+static int fsopen(const char *fsname, unsigned int flags) {
+	return syscall(__NR_fsopen, fsname, flags);
+}
+
+bool check_new_mount_api_available() {
+	fsopen("", 0);
+	return errno != ENOSYS;
```
sargun, line 67: neat.

mount/mount.c
solarkennedy: Hmm. I guess I thought that titus-mount did more than nfs, and that I only wanted the new behavior to be on that.

Looking at the executor code, it looks like we only invoke titus-mount on efs mounts, so this gate won't do anything in practice.

mount/mount.c
solarkennedy: Why? They all seem to be ints here:

https://github.com/torvalds/linux/blob/60e720931556fc1034d0981460164dcf02697679/samples/vfs/test-fsmount.c#L72

PR #458, opened 2020-10-02
Refactor docker runtime to not be built around Mesos

sargun:
Refactor docker runtime to have an API which isn't designed around legacy mesos
    
    This change is mostly just cleanup. It's basically moving the Docker runtime
    logic to be purely Prepare/Start/Stop, and not getting into the idea that
    one given runner, or runtime can run multiple containers. This design
    originally stemmed from Mesos, where one executor would actuate multiple
    containers.
    
    We don't have Mesos anymore. :).


sargun:
@solarkennedy @alekar -- This is a 0 logic (Except a tiny bit with how context work) change. It's just shuffling around code. Please take a look.

solarkennedy:
The diff is pretty big in a part of this codebase I haven't looked at before. I don't think I'll be able to say much.
I also don't understand the context canceling stuff (were we canceling before?)

PR #459, opened 2020-10-05
Rename .codecov -> codecov

sargun:


solarkennedy:
Why? 
https://docs.codecov.io/docs/codecov-yaml#can-i-name-the-file-codecovyml

Even with this change we dropped coverage somehow on random files?

I feel like we must be missing something else about how codecov works... Oh I see, our yaml isn't valid. Let me try a thing.

sargun:
I hate yaml

solarkennedy (approved):

PR #462, opened 2020-10-05
Plumb through pods

sargun:


executor/runtime/types/container.go
```diff
@@ -40,35 +43,43 @@ const (
 	BurstWorkloadType  WorkloadType = "burst"
 )
 
+func itoa(i int64) string {
```
rgulewich, line 46: https://golang.org/pkg/strconv/#Itoa ?

rgulewich: Oh, for int64.  I see.

rgulewich (approved):

PR #465, opened 2020-10-05
Refactor docker runtime to have an API which isn't designed around mesos

sargun:
This change is mostly just cleanup. It's basically moving the Docker runtime
logic to be purely Prepare/Start/Stop, and not getting into the idea that
one given runner, or runtime can run multiple containers. This design
originally stemmed from Mesos, where one executor would actuate multiple
containers.

We don't have Mesos anymore. :).

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -18,6 +18,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/Netflix/metrics-client-go/metrics"
@@ -39,6 +40,10 @@ import (
 	log "github.com/sirupsen/logrus"
 	"golang.org/x/sync/errgroup"
 	"golang.org/x/sys/unix"
+)
+
+var (
+	_ runtimeTypes.Runtime = (*DockerRuntime)(nil)
 )
 
 // units
@@ -71,6 +76,9 @@ const (
 	systemdImageLabel       = "com.netflix.titus.systemd"
 )
 
+// cleanupFunc can be registered to be called on container teardown, errors are reported, but not acted upon
+type cleanupFunc func() error
+
 // NoEntrypointError indicates that the Titus job does not have an entrypoint, or command
 var NoEntrypointError = &runtimeTypes.BadEntryPointError{Reason: errors.New("Image, and job have no entrypoint, or command")}
 
@@ -95,6 +103,7 @@ type ucred struct {
 
 // DockerRuntime implements the Runtime interface calling Docker Engine APIs
 type DockerRuntime struct { // nolint: golint
+	// Following fields to be set when NewDockerRuntime is called
 	metrics           metrics.Reporter
 	registryAuthCfg   *types.AuthConfig
 	client            *docker.Client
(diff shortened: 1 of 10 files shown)


sargun:
I accidentally fubar'd the old PR. I think @rgulewich has the most familiarity for now with this part of the code base, and it heavily overlaps with his work to "plumb through pods"

executor/runtime/docker/docker.go
```diff
@@ -1935,17 +1968,17 @@ func (r *DockerRuntime) setupGPU(ctx context.Context, c *runtimeTypes.Container,
...
 	var errs *multierror.Error
 
-	containerStopTimeout := time.Second * time.Duration(c.TitusInfo.GetKillWaitSeconds())
+	containerStopTimeout := time.Second * time.Duration(r.c.TitusInfo.GetKillWaitSeconds())
 	if containerStopTimeout == 0 {
 		containerStopTimeout = defaultKillWait
 	}
 
-	if containerJSON, err := r.client.ContainerInspect(context.TODO(), c.ID); docker.IsErrNotFound(err) {
+	if containerJSON, err := r.client.ContainerInspect(context.TODO(), r.c.ID); docker.IsErrNotFound(err) {
```
rgulewich, line 1981: Can probably change `context.TODO()` to `ctx` here now?

executor/runtime/docker/docker.go
```diff
@@ -1955,32 +1988,32 @@ func (r *DockerRuntime) Kill(c *runtimeTypes.Container) error { // nolint: gocyc
 		goto stopped
 	}
 
-	if err := r.client.ContainerStop(context.TODO(), c.ID, &containerStopTimeout); err != nil {
+	if err := r.client.ContainerStop(context.TODO(), r.c.ID, &containerStopTimeout); err != nil {
```
rgulewich, line 1991: Same as above

executor/runtime/docker/docker.go
```diff
@@ -1955,32 +1988,32 @@ func (r *DockerRuntime) Kill(c *runtimeTypes.Container) error { // nolint: gocyc
...
 		r.metrics.Counter("titus.executor.dockerStopContainerError", 1, nil)
-		log.Errorf("container %s : stop %v", c.TaskID, err)
+		log.Errorf("container %s : stop %v", r.c.TaskID, err)
 		errs = multierror.Append(errs, err)
 	} else {
 		goto stopped
 	}
 
-	if err := r.client.ContainerKill(context.TODO(), c.ID, "SIGKILL"); err != nil {
+	if err := r.client.ContainerKill(context.TODO(), r.c.ID, "SIGKILL"); err != nil {
```
rgulewich, line 1999: Same as above

rgulewich (approved):

executor/runtime/docker/docker.go
sargun: No, we can't, because by the time that this part (Kill) gets called, the parent context may already have been cancelled (hence why we're killing. I don't yet know how to solve this problem. Suggestions welcome.

executor/runtime/docker/docker.go
sargun: No, we can't, because by the time that this part (Kill) gets called, the parent context may already have been cancelled (hence why we're killing. I don't yet know how to solve this problem. Suggestions welcome.

executor/runtime/docker/docker.go
sargun: No, we can't, because by the time that this part (Kill) gets called, the parent context may already have been cancelled (hence why we're killing. I don't yet know how to solve this problem. Suggestions welcome.

executor/runtime/docker/docker.go
rgulewich: OK, I see.  The only thing I can think of is doing a `context.WithTimeout()`, so at least this entire block has some timeout associated with it.  Not the greatest, though.

PR #476, opened 2020-10-12
Add more logging for tasks and referer

nabilschear:
### Description of the Change

Trying to get parity between IMDS logging that we had in baseami with titus.  Just adding a couple of fields to make correlation easier.

### DIFF
--- a/metadataserver/server.go
+++ b/metadataserver/server.go
@@ -227,7 +227,7 @@ func (ms *MetadataServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	r2 := r.WithContext(ctx)
 	defer cancel()
 	defer func() {
-		log.WithFields(logging.Entry(ctx)).Infof("Request %s %s '%s'", r2.Method, r2.RequestURI, r2.UserAgent())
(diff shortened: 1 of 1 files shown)


metadataserver/server.go
```diff
@@ -235,6 +235,8 @@ func (ms *MetadataServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		"uri":           r.RequestURI,
 		"path":          r.URL.Path,
 		"token-present": len(r.Header.Get("X-aws-ec2-metadata-token")) > 0,
+		"referer":       r.Header.Get("Referer"),
+		"taskid":        ms.titusTaskInstanceID,
```
sargun, line 239: So, task ID is already present in the logs in the systemd unit name. Do you want duplication?

nabilschear: yeah, i saw that.  it's a pain to get out of there in elastic or bpd.  requires tokenizing the unit name and what not.

sargun: I know that it's a pain to get from elastic, but it shouldn't be from BDP. Let me know if you need help with that.

sargun (approved):

nabilschear:
Hold off on merging this.  i'm gonna use this to build some other metadata service changes

sargun:
lgtm.

PR #477, opened 2020-10-13
Fix pod env fallback

sargun:
We have code to fallback to the environment provided in
containerinfo is the environment in the pod is not populated.
The check previously was just seeing if the pod env was empty,
but VK automatically adds some environment variables, like:
 * KUBERNETES_SERVICE_HOST
 * KUBERNETES_SERVICE_PORT

In order to properly handle fallback, now we look for the
presence of an environment variable TITUS_TASK_INSTANCE_ID.

### DIFF
--- a/executor/runtime/types/types.go
+++ b/executor/runtime/types/types.go
@@ -244,9 +244,11 @@ func (c *Container) GetEnv() map[string]string {
 	env["NETFLIX_AUTO_SCALE_GROUP"] = asgName
 	env["TITUS_IAM_ROLE"] = c.iamRole
 
-	// User provided environment
-	if c.pod == nil || len(c.pod.Spec.Containers[0].Env) == 0 {
-		// titus provided can override user provided
+	// passed environment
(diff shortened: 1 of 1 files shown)


rgulewich (approved):

executor/runtime/types/types.go
```diff
@@ -255,12 +257,38 @@ func (c *Container) GetEnv() map[string]string {
...
+			"TITUS_ENV_FROM": "pod",
+		}
 		for _, val := range c.pod.Spec.Containers[0].Env {
 			if val.Value != "" {
-				env[val.Name] = val.Value
+				podEnv[val.Name] = val.Value
 			}
 		}
+		if val, ok := podEnv[titusTaskInstanceIDKey]; !ok {
+			// We need to have the pod env have this variable
```
rgulewich, line 282: Maybe change this comment here - if `titusTaskInstanceIDKey` is set, then the control plane is passing all of the environment in the pod, so we're good to use it.  Otherwise, use containerInfo.

PR #478, opened 2020-10-15
Create a standalone binary for titus-log-uploader

cconger:
### Description of the Change
Creates a standalone binary version of the previously embedded log uploader from the virutal kubelet.

The uploader also runs a simple log-viewer webserver which expects to be proxied to from the log-viewer running on the agent.

I include a simple builder/alpine wrapped container for injecting into end-user Pods via the mutating web hook.  Should this be its own repository? or do you think I should extend the build pipeline to generate this image?

All configuration is expected to be provided as Environment Variables.


logviewer/kubelet/handlers.go
```diff
@@ -0,0 +1,116 @@
+package kubelet
```
solarkennedy, line 1: This is a lot of new code for the logviewer. Do we even need it?

On the stock log viewer, is the only thing we need to change the base path for where the logs are?

Everything else should be the same.

I guess this is complicated by the fact that there are two modes, proxy and non :(

cmd/titus-log-uploader/config.go
```diff
@@ -0,0 +1,164 @@
...
+
+func mustReadRegexpFromEnv(envKey string) *regexp.Regexp {
+	if val, ok := os.LookupEnv(envKey); ok {
+		r, err := regexp.Compile(val)
+		if err != nil {
+			logrus.WithFields(logrus.Fields{
+				"key":   envKey,
+				"value": val,
+				"err":   err,
+			}).Error("Unable to parse valid regular expression from environment variable")
```
solarkennedy, line 73: I haven't seen this `mustFoo` pattern before.

Is there a particular reason for handling errors like this and continuing in the face of these errors?

cconger: Will fix this error to be a `Fatal()` and on line 94 as well.  Is your preference to Panic?  I was modeling off of the `template.Must` from the stdlib: https://golang.org/pkg/html/template/#Must

I've seen it elsewhere but examples are, of course, eluding me.

logviewer/kubelet/handlers.go
cconger: Fair point, I'll see if I can try to convince the default handlers to perform the way I want.  I may have overreacted when I saw all the string building in the base handlers.  Might have a followup PR to swap them to templates.

alekar: As a note, this is a replacement for the logviewer we inject, not the one in titus-executor, right? I like the new, idiomatic handler. I'd prefer we add some tests to make sure it behaves 'exactly' like the log viewer we inject today, because there are platforms which might depend on its idiosyncrasies.

cmd/titus-log-uploader/config.go
```diff
@@ -0,0 +1,164 @@
...
+	"regexp"
+	"strconv"
+	"time"
+
+	"github.com/sirupsen/logrus"
+
+	"github.com/Netflix/titus-executor/filesystems"
+)
+
+const (
```
alekar, line 14: Awesome, thanks for doing this.

cmd/titus-log-uploader/main.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/Netflix/titus-executor/filesystems"
+	kubeviewer "github.com/Netflix/titus-executor/logviewer/kubelet"
+	"github.com/Netflix/titus-executor/tag"
+	"github.com/Netflix/titus-executor/uploader"
+)
+
+/*
+	This binary handles log persistence and access from within a pod.
+
+	It accomplishes this by having an injected shared volume and a webserver.  The binary monitors this shared volume and publishes to s3 on a specified cadience.  Additionally it provides a simple web server for stating and listing these logs in accordance with the logviewer API from the titus-api server.
```
alekar, line 23: Nit; long line.

cmd/titus-log-uploader/main.go
```diff
@@ -0,0 +1,113 @@
...
+	if err != nil {
+		logrus.Fatal(err)
+	}
+
+	err = watcher.Watch(ctx)
+	if err != nil {
+		logrus.Fatal(err)
+	}
+
+	// Start the logviewer webserver
```
alekar, line 79: My preference is to make this a different binary, and inject separate uploading and viewing sidecars.

alekar (commented):
Left a few comments. LGTM otherwise.

cmd/titus-log-uploader/main.go
solarkennedy: https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line

alekar: https://www.kernel.org/doc/html/v4.10/process/coding-style.html#breaking-long-lines-and-strings

cmd/titus-log-uploader/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM golang:1-alpine as builder
```
sargun, line 1: Why is this a separate builder / Dockerfile than the other ones?

cmd/titus-log-uploader/config.go
```diff
@@ -0,0 +1,164 @@
+package main
+
+import (
+	"os"
+	"regexp"
+	"strconv"
+	"time"
+
+	"github.com/sirupsen/logrus"
+
```
sargun, line 10: Extra newline

cmd/titus-log-uploader/main.go
```diff
@@ -0,0 +1,113 @@
...
+	"context"
+	"net/http"
+	"os"
+	"os/signal"
+	"syscall"
+	"time"
+
+	"github.com/Netflix/metrics-client-go/metrics"
+	"github.com/sirupsen/logrus"
+
```
sargun, line 13: extra newline.

cmd/titus-log-uploader/main.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/Netflix/titus-executor/uploader"
+)
+
+/*
+	This binary handles log persistence and access from within a pod.
+
+	It accomplishes this by having an injected shared volume and a webserver.  The binary monitors this shared volume and publishes to s3 on a specified cadience.  Additionally it provides a simple web server for stating and listing these logs in accordance with the logviewer API from the titus-api server.
+*/
+
+func main() {
```
sargun, line 26: We need to make it so that the logrus is connected to journald

cmd/titus-log-uploader/main.go
```diff
@@ -0,0 +1,113 @@
...
+		}
+	}()
+
+	sigs := make(chan os.Signal, 1)
+	signal.Notify(sigs, syscall.SIGTERM, syscall.SIGINT)
+
+	// Block on a signal or toplevel context cancel
+	select {
+	case sig := <-sigs:
+		logrus.WithFields(logrus.Fields{
```
sargun, line 98: Is using one field, please just use WithField

logviewer/api/pod.go
```diff
@@ -0,0 +1,100 @@
...
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	IP := pod.Status.PodIP
+	if IP == "" {
+		http.Error(w, "No IP found for the targetted pod", http.StatusNotFound)
+		return
+	}
+	url, err := url.Parse("http://" + IP + ":8004")
```
sargun, line 62: How will you make sure you have connectivity to the pod IP, and the titus agent isn't filtered from reaching it? For example, for the studio accounts, it is this way. @rgulewich wrote a reverse proxy to solve this issue that goes across namespaces.

logviewer/kubelet/handlers.go
sargun: Secnded, there's a lot of duplication.

cmd/titus-log-uploader/main.go
cconger: I can do that, I'd appreciate you fleshing out your thinking on that, would you ever want one without the other?

cmd/titus-log-uploader/Dockerfile
cconger: Temporary build solution until I got an answer to the standard practice of releasing sidecars.  I'm thinking it might make sense to have this be a separate repo and just pull in the packages I need from this, but was waiting until I got feedback from Rob.

logviewer/api/pod.go
cconger: Thanks for the heads up, I will reach out to @rgulewich about that use case.  I ran this past @alekar but am open to your thoughts on how we should route internally.  An alternatives could be a host accessible mount with named pipe created by the log-viewer at startup.  I think this would work, but would have to try before I was confident but maybe worth investigation.

cmd/titus-log-uploader/main.go
alekar: Yes, exactly. We already have customers who run filebeat inside their containers (and some day their pods).

sargun: Didn't @rgulewich already build that?

logviewer/api/pod.go
sargun: I would prefer unix socket over pipe. But yeah, either approach works.

cmd/titus-log-uploader/Dockerfile
rgulewich: If the log uploader is going to live in this repo, then let's move it to `hack/images/titus-log-uploader` (or similar) - that way it matches what we do with sshd.

cmd/titus-log-uploader/main.go
rgulewich: Only if systemd's actually running in the container, though - it should still fallback to writing to stdout / stderr otherwise.

cmd/titus-log-uploader/main.go
rgulewich: We use the same logviewer that BaseOS installs on all the EC2 instances, packaged up as a container.  It implements all of the old Titus "legacy" endpoints for backward compatibility.

This gets us out of the business of building a second logviewer, and keeps it consistent between EC2 and Titus.

rgulewich: The implementation in `cmd/titus-logviewer/...` is the old logviewer. It's kept around so that the integration tests in this repo have something to test against, but isn't used otherwise.

logviewer/api/pod.go
rgulewich: The socket approach is a good idea.  What do you think about preserving the current way of proxying, and writing an OCI hook to do the bind mounting of the container's pid 1, like we do here - https://github.com/Netflix/titus-executor/blob/master/executor/runtime/docker/docker_linux.go#L146

cconger:
Closing in favor of #513

PR #479, opened 2020-10-17
titus-mount: Use new mount apis for newer kernels

solarkennedy:
This changes the titus-mount command to use the newer
kernel mounting APIs on newer linux kernels:
https://lwn.net/Articles/753473/

On the plus side, this removes the need to have a patched
kernel to get NFS mounts to work with user namespaces
(only a tiny patch is required).

On the down side, the titus-mount command needs to be more
complex, initializing a fsfd in the user namespace, and then
passing that back for the parent titus-mount command to
configure and actually mount.

----

I had considered splitting the command up into parts:
1. titus-resolve (just to do the dns resolution in the net ns
2. titus-create-fsfd (just to get in the namespace to create the fsfd and ... get it back somehow
3. Let the titus executor go code handle the rest

In the end I decided that the 'monolithic' C binary was going to be easier
to live with in the longer term, because you can run it standalone,
at the cost of having to do string parsing in C.

----

Once we upgrade this, we cannot go back. This must be run on a >5.4 kernel only.
So it will need to be deployed in lock-step with a kernel upgrade.

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -1522,57 +1522,29 @@ func validateMessage(c *runtimeTypes.Container, message events.Message) {
 const (
 	// MS_RDONLY indicates that mount is read-only
 	MS_RDONLY = 1 // nolint: golint
-	// MS_MGC_VAL does nothing, it's just a non-0 value that used to be legacy in the kernel to indicate mount options
-	MS_MGC_VAL = 0xC0ED0000 // nolint: golint
 )
 
 func (r *DockerRuntime) setupEFSMounts(parentCtx context.Context, c *runtimeTypes.Container, rootFile *os.File, cred *ucred, efsMountInfos []efsMountInfo) error {
-	baseMountOptions := []string{"vers=4.1,nosharecache,rsize=1048576,wsize=1048576,timeo=600,retrans=2,noresvport"}
-
-	mntNSPath := filepath.Join("/proc", strconv.Itoa(int(cred.pid)), "ns", "mnt")
-	mntNSFile, err := os.OpenFile(mntNSPath, os.O_RDONLY, 0444)
-	if err != nil {
-		return err
-	}
-	defer shouldClose(mntNSFile)
-
-	netNSPath := filepath.Join("/proc", strconv.Itoa(int(cred.pid)), "ns", "net")
-	netNSFile, err := os.OpenFile(netNSPath, os.O_RDONLY, 0444)
-	if err != nil {
-		return err
-	}
-	defer shouldClose(netNSFile)
-
+	baseMountOptions := []string{"vers=4.1,rsize=1048576,wsize=1048576,timeo=600,retrans=2"}
 	for _, efsMountInfo := range efsMountInfos {
 		// Todo: Make into a const
-		// TODO: Run this under the container's PID namespace
 		// Although 5 minutes is probably far too much here, this window is okay to be large
 		// because the parent window should be greater
 		ctx, cancel := context.WithTimeout(parentCtx, 5*time.Minute)
 		defer cancel()
-		cmd := exec.CommandContext(ctx, "/apps/titus-executor/bin/titus-mount") // nolint: gosec
-		// mntNSFD = 3+0 = 3
-		// userNSFD = 3+1 = 4
-		flags := MS_MGC_VAL
+		cmd := exec.CommandContext(ctx, "/apps/titus-executor/bin/titus-mount", strconv.Itoa(int(cred.pid))) // nolint: gosec
+		flags := 0
 		if efsMountInfo.readWriteFlags == ro {
 			flags = flags | MS_RDONLY
 		}
-
-		cmd.ExtraFiles = []*os.File{mntNSFile, netNSFile}
-
 		mountOptions := append(
 			baseMountOptions,
 			fmt.Sprintf("fsc=%s", c.TaskID),
+			fmt.Sprintf("source=%s:%s", efsMountInfo.hostname, efsMountInfo.cleanEfsFsRelativeMntPoint),
 		)
 		cmd.Env = []string{
-			// Go-ism
-			// If you pass file descriptors over os/cmd, it will be 3+n where N is the index of the file descriptor in the slice you pass.
-			// See above for "math"
-			"MOUNT_NS=3",
-			"NET_NS=4",
 			fmt.Sprintf("MOUNT_TARGET=%s", efsMountInfo.cleanMountPoint),
 			fmt.Sprintf("MOUNT_NFS_HOSTNAME=%s", efsMountInfo.hostname),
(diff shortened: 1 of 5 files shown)


mount/mount.c
```diff
@@ -54,30 +45,223 @@ static int dns_lookup(const char *hostname, struct sockaddr_in *addr)
...
+		return 1;
+	}
+	if (worker == 0) {
+		close(sk_pair[0]);
+		ret = setup_fsfd_in_namespaces(sk_pair[1], pidfd);
+		close(sk_pair[1]);
+		exit(ret);
+	}
+	close(sk_pair[1]);
+	fsfd = recv_fd(sk_pair[0]);
```
sargun, line 203: Does sk_pair[0] ever get closed?

mount/mount.c
```diff
@@ -54,30 +45,223 @@ static int dns_lookup(const char *hostname, struct sockaddr_in *addr)
...
+void switch_namespaces(int pidfd) {
+	int ret;
+	ret = setns(pidfd, CLONE_NEWNET | CLONE_NEWNS);
+	if (ret == -1) {
+		perror("setns net / mount");
+		exit(1);
+	}
+}
+
+void mount_and_move(int fsfd, const char* target, int pidfd) {
```
sargun, line 225: static

mount/mount.c
```diff
@@ -54,30 +45,223 @@ static int dns_lookup(const char *hostname, struct sockaddr_in *addr)
...
+		return -1;
+	}
+	if (!WIFEXITED(status) || WEXITSTATUS(status)) {
+		fprintf(stderr, "worker exited nonzero\n");
+		return -1;
+	}
+	return fsfd;
+}
+
+void switch_namespaces(int pidfd) {
```
sargun, line 216: static

mount/mount.c
```diff
@@ -54,30 +45,223 @@ static int dns_lookup(const char *hostname, struct sockaddr_in *addr)
...
+	close(sk_pair[1]);
+	fsfd = recv_fd(sk_pair[0]);
+	assert(fsfd >= 0);
+	if (waitpid(worker, &status, 0) != worker) {
+		perror("waitpid");
+		return -1;
+	}
+	if (!WIFEXITED(status) || WEXITSTATUS(status)) {
+		fprintf(stderr, "worker exited nonzero\n");
+		return -1;
```
sargun, line 211: why -1 in this case?

executor/runtime/docker/docker.go
```diff
@@ -1527,52 +1527,26 @@ const (
...
 		// Todo: Make into a const
-		// TODO: Run this under the container's PID namespace
 		// Although 5 minutes is probably far too much here, this window is okay to be large
 		// because the parent window should be greater
 		ctx, cancel := context.WithTimeout(parentCtx, 5*time.Minute)
 		defer cancel()
-		cmd := exec.CommandContext(ctx, "/apps/titus-executor/bin/titus-mount") // nolint: gosec
-		// mntNSFD = 3+0 = 3
-		// userNSFD = 3+1 = 4
+		cmd := exec.CommandContext(ctx, "/apps/titus-executor/bin/titus-mount", strconv.Itoa(int(cred.pid))) // nolint: gosec
```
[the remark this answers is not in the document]
solarkennedy: Don't you think it is simpler to pass the pid like this? (I assumed so because your version of the code accepted the pid as the first arg, and I thought it was more straightforward than passing the FDs like before)

mount/mount.c
solarkennedy: Seems like if something went wrong, the fsfd we got may not be valid.
But it looks like we don't check for -1, eh?

Is your concern that we are returning sentinel values for no reason, or that the parent process shouldn't care?

mount/mount.c
```diff
@@ -90,71 +275,25 @@ int main() {
...
-		if (dns_lookup(nfs_mount_hostname, &server_addr)) {
-			fprintf(stderr, "titus-mount: DNS lookup failed for %s. Exiting 1.\n", nfs_mount_hostname);
-			return 1;
-		}
-		ip_string = inet_ntoa(server_addr.sin_addr);
-		strcat(final_options, ",addr=");
-		strcat(final_options, ip_string);
-		fprintf(stderr, "titus-mount: using these nfs mount options: %s\n", final_options);
-	}
+	/* Now we can switch namespaces so we can lookup the ip for the addr */
```
rgulewich, line 287: Might be worth clarifying why it's switching both net and mount?

rgulewich (approved):

mount/mount.c
sargun: I have no idea what exit(-1) will do. Exit codes are supposed to be positive.

executor/runtime/docker/docker.go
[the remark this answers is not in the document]
sargun: It's simpler, but passing by FD is more "secure" and "reliable" (in that the PID cannot be recycled while in transition). If it's a faff, don't bother, because we're still in tini controlled time here.

sargun (approved):

mount/mount.c
solarkennedy: Hmm, we were not exiting -1 from here, we were returning -1 as the fsfd.
I've changed it to exit 1, these are all fatal errors.

mount/Makefile
```diff
@@ -1,2 +1,4 @@
-titus-mount: mount.c
-	musl-gcc -std=gnu11 -Wall -static -g -o titus-mount mount.c
+titus-mount: mount.c scm_rights.c
+	# musl needs this extra path here
+	# so it can pick up our linux headers for syscalls
+	C_INCLUDE_PATH=/usr/include/x86_64-linux-gnu/:/usr/include/:. musl-gcc -std=gnu11 -Wall -static -g -o titus-mount mount.c scm_rights.c
```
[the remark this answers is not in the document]
solarkennedy: It does, but that just doesn't work or behave the same way.
I don't fully understand include path semantics, but with -I it seems to include lots more than I want, or the order is wrong? idk, lots of weird macro expansion errors like:
```musl-gcc -I /usr/include/x86_64-linux-gnu/ -I /usr/include/ -I. -std=gnu11 -Wall -static -g -o titus-mount mount.c scm_rights.c
In file included from /usr/include/x86_64-linux-gnu/bits/libio.h:35:0,
                 from /usr/include/stdio.h:41,
                 from mount.c:3:
/usr/include/x86_64-linux-gnu/bits/_G_config.h:51:20: error: unknown type name ‘__gnuc_va_list’; did you mean ‘__NEED_va_list’?
 #define _G_va_list __gnuc_va_list
                    ^```

executor/runtime/docker/docker.go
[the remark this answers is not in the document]
solarkennedy: Yea I would rather not give up the ability to run this command by itself without emulating what titus-executor does.
The pid recycling is real, but should be super rare in this tiny span of time here.

solarkennedy:
Ok, all functional tests are green with this code.
I've squashed and rebased in preparation to merge on monday.

PR #480, opened 2020-10-20
Add instance identity document support

sargun:


solarkennedy (approved):
How did we get away without this doc for so long?

cmd/titus-metadata-service/main.go
```diff
@@ -158,6 +159,12 @@ func main() {
 			Value:       "",
 			EnvVar:      "EC2_REGION",
 		},
+		cli.StringFlag{
+			Name:        "account-id",
+			Usage:       "The 'network' account ID the container is running in",
+			Destination: &accountID,
+			EnvVar:      "NETFLIX_ACCOUNT_ID",
```
solarkennedy, line 166: We are going to use the variable the agents already have right?

cmd/titus-metadata-service/main.go
```diff
@@ -118,6 +118,7 @@ func main() {
 		backingMetadataServer      string
 		metatronEnabled            bool
 		region                     string
+		accountID                  string
```
rgulewich, line 121: Where does `accountID` get passed into `types.MetadataServerConfiguration` as `NetflixAccountID`?  Should it be around L246 below?

rgulewich (approved):

PR #484, opened 2020-10-28
Move pod constants to titus-kube-common

rgulewich:
See: https://github.com/Netflix/titus-kube-common

This also fixes a bunch of linting errors turned up due to a new version of golangci-lint getting installed.


sargun:
Cool. thanks for upgrading golang ci lint. We upgraded a few others deps, but the big one I see is github.com/golang/protobuf v1.4.2. Does that have any effect, should we run `make protogen`?

sargun (approved):

rgulewich:
> Cool. thanks for upgrading golang ci lint. We upgraded a few others deps, but the big one I see is github.com/golang/protobuf v1.4.2. Does that have any effect, should we run `make protogen`?

Yeah, good point - will update.

cmd/titus-imds-cni/env.go
```diff
@@ -13,7 +13,7 @@ import (
 
 	"github.com/Netflix/titus-executor/utils/k8s"
 
-	"github.com/golang/protobuf/proto"
+	"github.com/golang/protobuf/proto" // nolint: staticcheck
```
rgulewich, line 16: As a heads up: these were due to the new version of `golangci-lint` not being happy that this package is marked as deprecated.  It's actually a major version update, which I didn't want to take on as part of this change.  For more info, see here: https://blog.golang.org/protobuf-apiv2

rgulewich:
I tried updating the generated proto files, and it turns out that the new v1 version of github.com/golang/protobuf is actually implemented using the v2 version, and it's not actually backwards compatible: https://github.com/golang/protobuf/issues/1014

(It also caused tests to start OOMing, which added a bonus level of concern)

This seems risky enough that I'm pinning to v1.3.3 for the time being, until the upstream compatibility issues get fixed.

PR #485, opened 2020-10-29
Detect gpu support based on device presence, not instance type

rgulewich:
We used to use the Metadata service to get the instance type and then had a regex to determine GPU instance types. This resulted in a code change for every GPU instance type.  Instead, let's use the presence of an nvidia device, to avoid changing this regex over and over.


nvidia/nvidia.go
```diff
@@ -104,19 +97,14 @@ func NewNvidiaInfo(ctx context.Context, runtime string) (types.GPUManager, error
...
 
 func isGPUInstance(ctx context.Context) (bool, error) {
-	// Only check if we're on a GPU instance type
-	sess := session.Must(session.NewSession())
-	metadatasvc := ec2metadata.New(sess)
-	instanceType, err := metadatasvc.GetMetadata("instance-type")
-	if err != nil {
+	if _, err := os.Stat(nvidiaDev); err != nil {
+		if os.IsNotExist(err) {
+			logger.G(ctx).Info("Not on a GPU instance type. No GPU info available.")
```
solarkennedy, line 102: I know this was there before, but I don't think we need to log this. This is the "normal" path.

nvidia/nvidia.go
```diff
@@ -104,19 +97,14 @@ func NewNvidiaInfo(ctx context.Context, runtime string) (types.GPUManager, error
...
+		}
 		return false, err
 	}
 
-	if !r.MatchString(instanceType) {
-		logger.G(ctx).Info("Not on a GPU instance type. No GPU info available.")
-		return false, nil
-	}
-
 	return true, nil
```
solarkennedy, line 108: Optional: Can you rewrite this to be more "positive", in the sense that it only returns true if it really exists?
I think that is a better long term design, where we only say something is a gpu if we detect the device, no error needed to pass back.

solarkennedy (approved):

nvidia/nvidia.go
rgulewich: I'm not sure I follow - if the code makes it to this point, then the device exists.  I'm a little leery about hiding errors here, since if we can't stat files, other things on this box are likely messed up. :)

solarkennedy: Gotcha.
I agree with logging the error here if it isn't taken care of by the caller.

nvidia/nvidia.go
rgulewich: Good call - moved so that it logs in the "abnormal" case.

PR #486, opened 2020-10-29
Fix blocked describe channel

sargun:
getGCableENIs now checks context, and exits when it's done
    
    getGCableENIs previously would iterate over all ENIs, and then and
    write them to a channel. This was a buffered channel that had a
    depth of 10,000. Simultaneously, there was a context that was used
    by the describeENIsFoGCWorker that read off this channel to
    abort reading off the channel. If the context was cancelled,
    the describeENIsFoGCWorker would stop reading off of the
    (describe) channel.
    
    This was previously never a problem because the size of the buffer
    was 10,000, and we never had more than 10,000 ENIs. Now that we have
    more than 10,000 ENIs, it means that this describe loop can stall
    out.
    
    I wish that Go channels were more like Erlang, and that they were
    unbounded, and if they made memory run out, they would panic,
    or you could get an error on write. That seems far more sane.

### DIFF
--- a/vpc/service/gc_v3.go
+++ b/vpc/service/gc_v3.go
@@ -390,6 +390,8 @@ func (vpcService *vpcService) doGCENIs(ctx context.Context, item *regionAccount)
 	defer cancel()
 
 	defer time.AfterFunc(11*time.Minute, func() {
+		// TODO: Consider panicing if we hit this condition
+		// TODO: Consider adding such a watchdog to all of these functions
 		logger.G(ctx).Warning("Function running too long")
 	}).Stop()
 
--- a/vpc/service/gc_v3.go
+++ b/vpc/service/gc_v3.go
@@ -498,7 +498,13 @@ ORDER BY RANDOM()
 			span.SetStatus(traceStatusFromError(err))
 			return err
 		}
-		ch <- eni
+		select {
+		case ch <- eni:
+		case <-ctx.Done():
+			err = fmt.Errorf("Context done while trying to write to gc-able ENIs: %w", ctx.Err())
+			tracehelpers.SetStatus(err, span)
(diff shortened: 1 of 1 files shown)


solarkennedy:
For my own context, this bug was found from a hint from a goroutine dump. Channel blocked for... 807 minutes!?:
```
goroutine 4970862331 [chan send, 807 minutes]:
github.com/Netflix/titus-executor/vpc/service.(*vpcService).getGCableENIs(0xc000143560, 0x147c600, 0xc0600bbfb0, 0xc000729010, 0x9, 0xc000729020, 0xc, 0xc05fcfd500, 0x0, 0x0)
         /var/lib/buildkite-agent/builds/ip-192-168-2-43-1/netflix/titus-executor/vpc/service/gc_v3.go:501 +0x32d
 github.com/Netflix/titus-executor/vpc/service.(*vpcService).doGCENIs.func2(0xc000cd5768, 0x0)
         /var/lib/buildkite-agent/builds/ip-192-168-2-43-1/netflix/titus-executor/vpc/service/gc_v3.go:421 +0x6f
 golang.org/x/sync/errgroup.(*Group).Go.func1(0xc00222bf50, 0xc00222bf80)
         /go/pkg/mod/golang.org/x/sync@v0.0.0-20190911185100-cd5d95a43a6e/errgroup/errgroup.go:57 +0x64
 created by golang.org/x/sync/errgroup.(*Group).Go
         /go/pkg/mod/golang.org/x/sync@v0.0.0-20190911185100-cd5d95a43a6e/errgroup/errgroup.go:54 +0x66
```

solarkennedy:
I'll try to take this and run with it. Starting with this broken test (which seems unrelated to this change)...

PR #487, opened 2020-11-03
Use a more generic EC2_ACCOUNT_ID variable for ec2-account-id

solarkennedy:
The `EC2_OWNER_ID` variable is a netflixism that represents the aws account id that controls the underlying ec2 instance.

This number is changing as we change accounts, which is breaking ssh (the only thing that uses this config setting.

To fix it, I would like to being consuming the very generic `EC2_OWNER_ID` variable, and we can set that "correctly" in our netflix init scripts.
(specifically, I'll set it to the expected prod/test settings)

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -52,7 +52,7 @@ type Config struct {
 
 	ContainerSSHDCAFile string
 	ContainerSSHDUsers  cli.StringSlice
-	EC2AccountID        string
+	SSHAccountID        string
 
 	// CopiedFromHost indicates which environment variables to lift from the current config
 	copiedFromHostEnv cli.StringSlice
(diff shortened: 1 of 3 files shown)


JackOfMostTrades:
Since this _isn't_ going to end up being the underlying aws/ec2 account id, would it make sense to use a name that's more accurate? E.g. `JOB_ACCOUNT_ID`?

solarkennedy:
Haha, we already have an iam account id for a job, `JOB_ACCOUNT_ID` also doesn't reflect what this concept is.

JackOfMostTrades:
D'oh! Maybe `SSH_ACCOUNT_ID` if that's really the only thing this is used for. Doesn't matter too much to me as long as the right thing gets plumbed through.

JackOfMostTrades (approved):

solarkennedy:
Hmm. @sargun what do you think of `SSH_ACCOUNT_ID` ?
This config setting is only used for ssh currently, and with a name like that it is unlikely someone will use it accidentally thinking it is something else.

sargun:
@solarkennedy I like SSH_ACCOUNT_ID.

sargun (approved):

JackOfMostTrades (approved):
LGTM

PR #493, opened 2020-11-18
Service mesh is enabled if an error is *not* returned for image name

rgulewich:
Rather than the opposite.


gabrielhartmann (approved):

solarkennedy (approved):

rgulewich:
Note: there weren't any tests for this functionality, which is why this bug got through.  I'll add them in a follow-on commit.

PR #495, opened 2020-11-19
Add more tests around service mesh enabling

rgulewich:
As a follow-on to the last commit.

### DIFF
--- a/executor/runtime/types/container_test.go
+++ b/executor/runtime/types/container_test.go
@@ -668,3 +668,41 @@ func TestServiceMeshEnabled(t *testing.T) {
 	assert.NotNil(t, svcMeshConf)
 	assert.Equal(t, svcMeshConf.Image, imgName)
 }
+
+func TestServiceMeshEnabledWithConfig(t *testing.T) {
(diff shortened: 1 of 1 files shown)


executor/runtime/types/container_test.go
```diff
@@ -668,3 +668,41 @@ func TestServiceMeshEnabled(t *testing.T) {
 	assert.NotNil(t, svcMeshConf)
 	assert.Equal(t, svcMeshConf.Image, imgName)
 }
+
+func TestServiceMeshEnabledWithConfig(t *testing.T) {
+	// If service mesh is set to enabled, but neither the `ProxydServiceImage` config value
+	// or the passhtrough property are set, service mesh should end up disabled
+	config := config.Config{
+		ContainerServiceMeshEnabled: true,
```
solarkennedy, line 676: Can you add a negative/absent test, where we have an empty config and we assert that there is not a servicemesh sidecar?

rgulewich: Yeah, I added that in the last commit.  It's part of `TestNewContainer()` above.

solarkennedy (approved):

alekar (approved):

PR #497, opened 2020-11-24
IMDS: log response time and lock acquisition time

rgulewich:
Hopefully helpful in tracking down timeouts.

### DIFF
--- a/metadataserver/iam.go
+++ b/metadataserver/iam.go
@@ -12,9 +12,9 @@ import (
 
 	"sync"
 
-	"github.com/Netflix/titus-executor/metadataserver/types"
-
+	"github.com/Netflix/titus-executor/metadataserver/logging"
 	"github.com/Netflix/titus-executor/metadataserver/metrics"
+	"github.com/Netflix/titus-executor/metadataserver/types"
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/arn"
 	"github.com/aws/aws-sdk-go/aws/awserr"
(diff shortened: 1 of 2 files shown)


metadataserver/iam.go
```diff
@@ -347,10 +347,12 @@ func (proxy *iamProxy) doAssumeRole(sessionLifetime time.Duration) {
 }
 
 func (proxy *iamProxy) getRoleAssumptionState(ctx context.Context) *roleAssumptionState {
+	startTime := time.Now()
 	// So this can potentially block for up to the upper bound of the request timeout
 	proxy.startRoleAssumer()
 	proxy.roleAssumptionStateLock.RLock()
 	defer proxy.roleAssumptionStateLock.RUnlock()
+	log.WithFields(logging.Entry(ctx)).WithField("acquire-time", time.Since(startTime).Milliseconds()).Info("got role assumption state")
```
sargun, line 355: This is going to log _a lot_. What do you think about adding tracing here instead?

rgulewich: It will log one extra line whenever `/info` and `/security-credentials/{id}` are called.  Is that actually going to be a lot?  I thought about adding tracing, but didn't think 2 extra lines per request were going to be _that_ much.

sargun: I don't know, I guess I'd rather see if we can add it to the request context and in the middleware that does the request logging, if we can have it passed back in there.

rgulewich:
Closing in favour of https://github.com/Netflix/titus-executor/pull/498

PR #498, opened 2020-11-24
Don't start two IMDS assumerole goroutines

rgulewich:
Two fixes here:

1. Add times to request log lines: total request time, and time waiting for the assumerole lock
2. Only start one goroutine to do AssumeRoles in the background - the previous code was starting one per IAM router.

I also added a test for getting specific security credentials while I was in here.

### DIFF
--- a/metadataserver/iam.go
+++ b/metadataserver/iam.go
@@ -18,6 +18,8 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/arn"
 	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/aws/credentials/ec2rolecreds"
+	"github.com/aws/aws-sdk-go/aws/ec2metadata"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/sts"
 	"github.com/gorilla/mux"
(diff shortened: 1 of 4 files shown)


metadataserver/iam.go
```diff
@@ -75,9 +75,19 @@ func newIamProxy(ctx context.Context, router *mux.Router, config types.MetadataS
...
+		if config.STSEndpoint != "" {
+			endpoint = config.STSEndpoint
+		}
+
 		stsAwsCfg := aws.NewConfig().
 			WithRegion(config.Region).
-			WithEndpoint(fmt.Sprintf("sts.%s.amazonaws.com", config.Region))
+			WithEndpoint(endpoint)
+
+		if config.DisableSTSEndpointSSL {
```
sargun, line 87: Why would you ever want this?

metadataserver/iam.go
```diff
@@ -91,6 +101,13 @@ func newIamProxy(ctx context.Context, router *mux.Router, config types.MetadataS
 	}
 	proxy.roleName = splitRole[1]
 
+	// No need to block here
+	go proxy.startRoleAssumer()
```
sargun, line 105: Do we even need `startRoleAssumer`? If it's always started, why not have:
```
log.Info("Starting role assumer")
		proxy.doAssumeRole(defaultSessionLifetime)
		log.Info("Ran first assume role")

		go proxy.roleAssumer()
```
here?

metadataserver/iam.go
rgulewich: For testing - otherwise STS fails to connect to the mock endpoints.  I suppose I could put it behind a `STSTestMode` flag or something.

metadataserver/iam.go
rgulewich: I didn't write this piece of code - I only moved it around.  No idea why.

sargun: originally, every time we got a request that needed an assumerole, we would call the `startRoleAssumer` function to start the assume role in the background. You should be able to hoist this code to here. That should give you two side effects:

1. Simplicity
2. The IMDS will block on its first attempt to assume role.

metadataserver/iam.go
sargun: Can you use the Golang httptest TLS functionality? https://golang.org/pkg/net/http/httptest/#NewTLSServer

metadataserver/iam.go
rgulewich: Sure, that's fine.  PTAL.

sargun (approved):

PR #499, opened 2020-11-28
IMDS CNI: fully populate config file with all fields required for task identity

rgulewich:
Before, it was missing some fields that would make it incomplete.

### DIFF
--- a/cmd/titus-imds-cni/env.go
+++ b/cmd/titus-imds-cni/env.go
@@ -4,20 +4,18 @@ package main
 
 import (
 	"context"
-	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"os"
 	"path"
 	"strconv"
+	"time"
 
 	"github.com/Netflix/titus-executor/utils/k8s"
 
-	"github.com/golang/protobuf/proto" // nolint: staticcheck
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 
-	"github.com/Netflix/titus-executor/api/netflix/titus"
 	tt "github.com/Netflix/titus-executor/executor/runtime/types"
 	mt "github.com/Netflix/titus-executor/metadataserver/types"
 	"github.com/containernetworking/cni/pkg/skel"
@@ -91,7 +89,7 @@ func extractEnv(prev *current.Result, pod *v1.Pod) (map[string]string, error) {
 	for _, ip := range prev.IPs {
 		switch ip.Version {
 		case "4":
-			env["EC2_LOCAL_IPV4"] = ip.Address.String()
+			env[mt.EC2IPv4EnvVarName] = ip.Address.String()
 			ok = true
 		case "6":
 			env["EC2_IPV6S"] = ip.Address.String()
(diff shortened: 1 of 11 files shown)


executor/runtime/types/container.go
```diff
@@ -378,7 +385,7 @@ func addProcessLabels(containerInfo *titus.ContainerInfo, labels map[string]stri
 
 		command := process.GetCommand()
 		if command != nil {
-			commandStr := strings.Join(entryPoint[:], " ")
+			commandStr := strings.Join(command[:], " ")
```
rgulewich, line 388: @gabrielhartmann - turns out this has been a bug in this code all along.  Will fixing what gets put into the command label mess up something in titus-isolate or the training models?

solarkennedy: Huh. We have this internal ticket TITUS-4150 (deleted?) that sounds super similar.

rgulewich: CC: @rostyboost, in case you can weigh in here.

gabrielhartmann: LGTM.  I don't think it will screw any training severely, but we should confirm with @rostyboost

cmd/titus-imds-cni/env.go
```diff
@@ -109,9 +107,36 @@ func extractEnv(prev *current.Result, pod *v1.Pod) (map[string]string, error) {
...
+func getUserContainer(pod *v1.Pod) *v1.Container {
+	firstContainer := pod.Spec.Containers[0]
+	for _, c := range pod.Spec.Containers {
+		if c.Name == pod.Name {
+			ctr := c
+			return &ctr
+		}
+	}
+
+	return &firstContainer
```
gabrielhartmann, line 137: Why the fallback assumption that the first container is the user container if we don't find a podName == containerName match?  Is it better to fail?

executor/runtime/types/container.go
rostyboost: This should be ok.

cmd/titus-imds-cni/env.go
```diff
@@ -188,36 +214,37 @@ func openFile(dname, fname string) (*os.File, error) {
 	return fd, nil
 }
 
-func writeContainerInfo(pod *v1.Pod) error {
-	str, ok := pod.Annotations["containerInfo"]
+func writeContainerInfo(pod *v1.Pod, env map[string]string) error {
+	ipv4Addr, ok := env["EC2_LOCAL_IPV4"]
```
gabrielhartmann, line 218: I count 4 references in non-test code to this string.  Maybe time for a constant, but a different PR might be better.

executor/runtime/types/pod.go
```diff
@@ -0,0 +1,330 @@
...
+
+	"github.com/Netflix/titus-executor/api/netflix/titus"
+	"github.com/Netflix/titus-executor/uploader"
+	vpcTypes "github.com/Netflix/titus-executor/vpc/types"
+	"github.com/golang/protobuf/proto" // nolint: staticcheck
+	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+)
+
+// PodContainer is an implementation of Container backed only by a kubernetes pod.
```
gabrielhartmann, line 16: Can we add the force interface implementation trick here?

executor/runtime/types/pod.go
```diff
@@ -0,0 +1,330 @@
...
+	command       []string
+	entrypoint    []string
+	hostnameStyle string
+	ipv4Address   *string
+	pod           *corev1.Pod
+	titusInfo     *titus.ContainerInfo
+}
+
+func NewPodContainer(pod *corev1.Pod, ipv4Address *string) (*PodContainer, error) {
+	if pod == nil {
```
sargun, line 28: If pod is nil, that seems..bad, maybe panic?

executor/runtime/types/pod.go
```diff
@@ -0,0 +1,330 @@
...
+	err = proto.Unmarshal(data, &cInfo)
+	if err != nil {
+		return errors.Wrap(err, "unable to decode containerInfo protobuf")
+	}
+
+	c.titusInfo = &cInfo
+	return nil
+}
+
+func (c *PodContainer) AllowCPUBursting() bool {
```
gabrielhartmann, line 82: Why is this feature and others below hard-coded to false or empty string or nil?

cmd/titus-imds-cni/env.go
rgulewich: Picking the first container has been the behaviour so far in all of our code, so I'm preserving that for backwards compatibility.

executor/runtime/types/pod.go
rgulewich: In the case of the executor, shouldn't we return a nice-ish message?

executor/runtime/types/pod.go
rgulewich: This is only a partial implementation - the rest will be done as part of the "native pods" work.

gabrielhartmann: ok

sargun (approved):

PR #500, opened 2020-12-07
Added titus-oci-hook, with prestart hook in place for more stuff

solarkennedy:
This scaffolding is for my intermediate project: Getting sshd working on kubelet again.

To do this, I want to re-use as much as the titus-executor code as I can.
Previously I didn't do that, and it broke, and and it drifted in behavior (see https://github.com/Netflix/titus-executor/pull/487).

So my new goal is to implement ssh in such a way that, as we improve/change ssh on VK, it automatically applies to RK, and we don't have to continuously play catchup.

To do that, I propose this oci hook binary to use in https://github.com/Netflix-Skunkworks/titus-kubelet/pull/39. A single hook for all the things, to reduce the burden of making different binaries for all the things, and in theory no additional changes to titus-containerd or any other hook configuration to implement new stuff.

### DIFF
--- a/cmd/titus-oci-hook/main.go
+++ b/cmd/titus-oci-hook/main.go
@@ -1,0 +1,89 @@
+package main
+
+import (
+	"encoding/json"
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"os"
+
+	"github.com/sirupsen/logrus"
+
+	"github.com/wercker/journalhook"
+
+	"github.com/opencontainers/runtime-spec/specs-go"
+)
+
+func doPrestart() error {
+	journalhook.Enable()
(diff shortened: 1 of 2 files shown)


solarkennedy:
Once of the first things I think I want here is an oci hook that rejects pods that slipped through the pod mutation webhook.

In theory this should never happen, but it does on my devstack when I goof around, and I'm sure it will happen in prod some day.

It would be nice if these pod never even tried to launch, and I think the oci webhook is our last line of defense to do that.

solarkennedy:
I'm going to close this for now till we have a use case (I thought the use case was re-using some of our system sidecars like ssh, but I ended up going a different direction)

PR #504, opened 2020-12-15
Unlock version of golangci-lint by specifying it at runtime

sargun:


solarkennedy:
This is cool. Does it create lots of bogus go.mod file changes though after lint runs?

sargun:
It doesn't mess with go.mod from what I can tell.

rgulewich (approved):

solarkennedy (approved):

rgulewich:
This is great - thank-you!

PR #505, opened 2020-12-16
Metric for how long docker pull took.

alekar:

### DIFF
--- a/executor/runtime/docker/docker_image_pull.go
+++ b/executor/runtime/docker/docker_image_pull.go
@@ -44,13 +44,20 @@ func isBadImageErr(err error) bool {
 }
 
 func doDockerPull(ctx context.Context, metrics metrics.Reporter, client *docker.Client, ref string) error {
+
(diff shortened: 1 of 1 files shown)


executor/runtime/docker/docker_image_pull.go
```diff
@@ -44,13 +44,20 @@ func isBadImageErr(err error) bool {
...
+	duration := time.Since(start)
+
 	defer func() {
 		if resp != nil {
 			// We really don't care what the error is here, we can't do anything about it
 			shouldClose(resp)
 		}
 	}()
+
+	metrics.Timer("titus.executor.dockerPullDuration", duration, map[string]string{"image": ref})
```
sargun, line 59: Why? This should be extrapolated using evevents. Also, you have the `titus.executor.imagePullTime` metric.

alekar: I suppose I could use `titus.executor.imagePullTime`.

PR #506, opened 2020-12-22
Add titusParameter.agent.log.finalUploadMatchRegexp

alekar:
This parameter allows the user to specify that only the files that
are watched and uploaded during rotation be picked up during final
upload.

### DIFF
--- a/executor/runner/runner.go
+++ b/executor/runner/runner.go
@@ -368,7 +368,7 @@ func (r *Runner) maybeSetupExternalLogger(ctx context.Context, logDir string) er
 	}
 	logger.G(ctx).Info("Starting external logger")
 
-	wConf := filesystems.NewWatchConfig(logDir, r.container.UploadDir("logs"), r.container.LogUploadRegexp(), *r.container.LogUploadCheckInterval(), *r.container.LogUploadThresholdTime(), *r.container.LogStdioCheckInterval(), r.container.LogKeepLocalFileAfterUpload())
+	wConf := filesystems.NewWatchConfig(logDir, r.container.UploadDir("logs"), r.container.LogUploadRegexp(), *r.container.LogUploadCheckInterval(), *r.container.LogUploadThresholdTime(), *r.container.LogStdioCheckInterval(), r.container.LogKeepLocalFileAfterUpload(), r.container.FinalUploadMatchRegexp())
 	uploader, err := uploader.NewUploader(&r.config, r.container.LogUploaderConfig(), *r.container.IamRole(), r.container.TaskID(), r.metrics)
 	if err != nil {
 		return err
--- a/executor/runtime/types/container.go
+++ b/executor/runtime/types/container.go
@@ -50,6 +50,7 @@ const (
 	hostnameStyleParam               = "titusParameter.agent.hostnameStyle"
 	logUploadThresholdTimeParam      = "titusParameter.agent.log.uploadThresholdTime"
 	logUploadCheckIntervalParam      = "titusParameter.agent.log.uploadCheckInterval"
+	finalUploadMatchRegex            = "titusParameter.agent.log.finalUploadMatchRegex"
 	logStdioCheckIntervalParam       = "titusParameter.agent.log.stdioCheckInterval"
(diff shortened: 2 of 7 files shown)


andrew-leung (approved):

executor/runner/runner.go
```diff
@@ -368,7 +368,7 @@ func (r *Runner) maybeSetupExternalLogger(ctx context.Context, logDir string) er
 	}
 	logger.G(ctx).Info("Starting external logger")
 
-	wConf := filesystems.NewWatchConfig(logDir, r.container.UploadDir("logs"), r.container.LogUploadRegexp(), *r.container.LogUploadCheckInterval(), *r.container.LogUploadThresholdTime(), *r.container.LogStdioCheckInterval(), r.container.LogKeepLocalFileAfterUpload())
+	wConf := filesystems.NewWatchConfig(logDir, r.container.UploadDir("logs"), r.container.LogUploadRegexp(), *r.container.LogUploadCheckInterval(), *r.container.LogUploadThresholdTime(), *r.container.LogStdioCheckInterval(), r.container.LogKeepLocalFileAfterUpload(), r.container.FinalUploadMatchRegexp())
```
andrew-leung, line 371: What is the difference between `r.container.LogUploadRegexp()` and `r.container.FinalUploadMatchRegexp()`?

andrew-leung (approved):

executor/runner/runner.go
alekar: It's a regex, which answers the question 'should I only upload files that match regex in final upload'? Defaults to false.

filesystems/watcher.go
```diff
@@ -68,24 +68,26 @@ var defaultUploadRegexpList = []*regexp.Regexp{
...
+	uploadRegexp           *regexp.Regexp
+	uploadCheckInterval    time.Duration // The interval at which we check for files that need uploading.
+	uploadThresholdTime    time.Duration // Duration for which a file must be untouched to be uploader.
+	stdioLogCheckInterval  time.Duration
+	keepFileAfterUpload    bool
+	finalUploadMatchRegexp bool
 }
 
-func NewWatchConfig(localDir, uploadDir string, uploadRegexp *regexp.Regexp, uploadCheckInterval, uploadThresholdTime, stdioLogCheckInterval time.Duration, keepFileAfterUpload bool) WatchConfig {
+func NewWatchConfig(localDir, uploadDir string, uploadRegexp *regexp.Regexp, uploadCheckInterval, uploadThresholdTime, stdioLogCheckInterval time.Duration, keepFileAfterUpload, finalUploadMatchRegexp bool) WatchConfig {
```
solarkennedy, line 81: I don't think we need this layer of indirection. It adds no value as a constructor hides the keyword/order of arguments in the calling function.

executor/runner/runner.go
solarkennedy: I was also confused by this variable. What if was renamed to `UploadLogsBasedOnRotateRegexp`
(Although maybe it hints that there should just be 2 regexes, one for rotate (`defaultUploadRegexpList`)and one for upload (/logs/.*), and in this case users would want to update their upload one?)

solarkennedy (approved):

executor/runner/runner.go
rgulewich: Not to pile on too much, but I also got confused by this. :)

I _think_ it might read more clearly if this was another Regexp - `FinalUploadRegexp()`.  If the agent param is set, then this is set to `LogUploadRegexp()`.  Otherwise, it defaults to `.*`.  I'm a bit on the fence about how necessary this is, though.

executor/runtime/types/container.go
```diff
@@ -434,6 +437,15 @@ func (c *TitusInfoContainer) updateLogAttributes(titusInfo *titus.ContainerInfo)
 		c.logUploadRegexp = uploadRegexp
 	}
 
+	finalUploadMatchRegexStr, ok := titusInfo.GetPassthroughAttributes()[finalUploadMatchRegex]
```
rgulewich, line 440: Might want to use `getPassthroughBool()` here, to be consistent with other passthrough bools.

rgulewich (approved):

PR #507, opened 2020-12-22
Change titus-repear into a oneshot unit

sargun:


alekar (approved):

alekar:
Can we also get a metric published when we reap?

sargun:
I ripped out metrics. Any reason you can't generate metrics from raven?

alekar:
It seems pretty obvious what metrics we want and that we want them.
Besides, Raven can have a bad day?

sargun:
@alekar the problem with counter metrics is that they happen so infrequently that often times it'll mis-report as 0, and you can't drill-down to nf.node

PR #508, opened 2021-01-07
Use DAD for IPv6 address assignment

sargun:
Use DAD, but now we're on a newer kernel we can use optimistic addresses while the kernel is settling.

This should make IPv6 networking available / come up faster.

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -5,11 +5,11 @@ import (
 	"bufio"
 	"bytes"
 	"context"
+	"crypto/rand"
 	"encoding/json"
 	"fmt"
 	"io"
 	"io/ioutil"
-	"math/rand"
 	"net"
 	"os"
 	"os/exec"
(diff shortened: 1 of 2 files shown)


sargun:
I need to fix the buildkite build by having it pull in a newer kernel. :|.

sargun:
Alternatively, I could add logic to check if /proc/sys/net/ipv/default/optimistic_dad exists. Does anyone have a preference?

rgulewich:
> Alternatively, I could add logic to check if /proc/sys/net/ipv/default/optimistic_dad exists. Does anyone have a preference?

Let's add this check - I'm betting that I'm not the only one that's running tests on an older kernel.

rgulewich (approved):

executor/runtime/docker/docker.go
```diff
@@ -322,6 +322,18 @@ func stableSecret() string {
 	}
 	return net.IP(ipBuf).String()
 }
+func maybeAddOptimisticDad(sysctl map[string]string) {
+	if unix.Access("/proc/sys/net/ipv6/conf/default/use_optimistic", 0) == nil {
+		sysctl["net.ipv6.conf.default.use_optimistic"] = "1"
+	} else {
+		log.Warning("Not enabling use_optimistic")
+	}
+	if unix.Access("/proc/sys/net/ipv6/conf/default/optimistic_dad", 0) == nil {
```
solarkennedy, line 331: It looks like we were doing `all` and `default` before, but this happens before the interface is up (right?) if so then we just need default, so that is fine.

solarkennedy (approved):

PR #510, opened 2021-01-09
If an EIP is allocated to a task, return that IP from the IMDS proxy's public-ipv4 endpoint

rgulewich:
Now hitting http://169.254.169.254/latest/meta-data/public-ipv4 on a task with an EIP associated will return the EIP's address, not the task's IPv4 address.

According to [the AWS docs](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-categories.html):
> The public IPv4 address. If an Elastic IP address is associated with the instance, the value returned is the Elastic IP address.

Note that this also changes the behaviour to match the actual AWS IMDS - if there's no EIP attached, then the endpoint returns 404.

I didn't implement this in the CNI, since elastic IP support isn't present there, yet.

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -986,6 +986,9 @@ func (r *DockerRuntime) Prepare(parentCtx context.Context) error { // nolint: go
 			Success:     true,
 			Error:       "",
 			BranchENIID: "eni-cat-dog",
+			ElasticAddress: &vpcapi.ElasticAddress{
+				Ip: "203.0.113.11",
+			},
 		}
 		r.c.SetVPCAllocation(allocation)
 		logger.G(ctx).Info("Mocking networking configuration in dev mode to IP: ", allocation)
--- a/metadataserver/server.go
+++ b/metadataserver/server.go
@@ -29,6 +29,17 @@ import (
 
 // Derived from big data portal reports
 const defaultMacAddress = "00:00:00:00:00:00"
+const notFoundBody = `<?xml version="1.0" encoding="iso-8859-1"?>
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN"
+                 "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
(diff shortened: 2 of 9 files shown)


metadataserver/server.go
```diff
@@ -220,7 +233,13 @@ func (ms *MetadataServer) localIPV4(w http.ResponseWriter, r *http.Request) {
 
 func (ms *MetadataServer) publicIPV4(w http.ResponseWriter, r *http.Request) {
 	metrics.PublishIncrementCounter("handler.publicIPV4.count")
-	if _, err := fmt.Fprint(w, ms.ipv4Address); err != nil {
+	if ms.publicIpv4Address == nil {
```
sargun, line 236: I wonder if this will break anyone.

vpc/types/conversions.go
```diff
@@ -27,5 +27,9 @@ func AssignmentToAllocation(assignment *vpcapi.AssignIPResponseV3) Allocation {
 		alloc.IPV4Address = assignment.Ipv4Address
 	}
 
+	if assignment.ElasticAddress != nil {
+		alloc.ElasticAddress = assignment.ElasticAddress
```
sargun, line 31: Does VPC Service actually set this?

sargun (approved):

vpc/types/conversions.go
rgulewich: Yeah, at least in my testing.

metadataserver/server.go
rgulewich: This endpoint gets called a *lot* by eureka-java-client, but not much otherwise - I'm going to check with them before rolling this out.

executor/runtime/docker/docker.go
```diff
@@ -986,6 +986,9 @@ func (r *DockerRuntime) Prepare(parentCtx context.Context) error { // nolint: go
 			Success:     true,
 			Error:       "",
 			BranchENIID: "eni-cat-dog",
+			ElasticAddress: &vpcapi.ElasticAddress{
+				Ip: "8.8.8.8",
```
solarkennedy, line 990: How about one from the documentation subnet? (192.0.2.0/24, etc)

solarkennedy (approved):

PR #511, opened 2021-01-09
metadata proxy injector: retry rtnl_link_add()

rgulewich:
On 5.9 kernels, `rtnl_link_add()` is failing occasionally with EPERM. While I'm figuring out what's causing that, add retries so that the tasks don't fail.

### DIFF
--- a/inject/titus-inject-metadataproxy.c
+++ b/inject/titus-inject-metadataproxy.c
@@ -126,7 +126,14 @@ static void setup_veth(int private_ns_fd, int container_ns_fd) {
 	rtnl_link_set_name(peer, IN_CONTAINER_INTERFACE_NAME);
 	rtnl_link_set_ns_fd(peer, container_ns_fd);
 
-	err = rtnl_link_add(nls, veth, NLM_F_CREATE | NLM_F_EXCL);
+	for (int i = 0; i < 5; i++) {
+		err = rtnl_link_add(nls, veth, NLM_F_CREATE | NLM_F_EXCL);
+		if (err == 0) {
+			break;
+		}
+		WARN_ON(err < 0, "%s: unable to add link", nl_geterror(err));
+		sleep(1);
+	}
 	BUG_ON(err < 0, "%s: unable to add link", nl_geterror(err));
 
 	rtnl_link_put(peer);
--- a/inject/titus-inject-metadataproxy.c
+++ b/inject/titus-inject-metadataproxy.c
@@ -17,6 +17,7 @@
 #include <sched.h>
 #include <unistd.h>
 /* Netlink */
+#include <netlink/handlers.h>
 #include <netlink/socket.h>
 #include <netlink/route/link.h>
 #include <netlink/route/link/veth.h>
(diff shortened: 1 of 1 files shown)


sargun:
How did you figure out the issue was around rtnl link add.

inject/titus-inject-metadataproxy.c
```diff
@@ -126,7 +126,14 @@ static void setup_veth(int private_ns_fd, int container_ns_fd) {
 	rtnl_link_set_name(peer, IN_CONTAINER_INTERFACE_NAME);
 	rtnl_link_set_ns_fd(peer, container_ns_fd);
 
-	err = rtnl_link_add(nls, veth, NLM_F_CREATE | NLM_F_EXCL);
+	for (int i = 0; i < 5; i++) {
+		err = rtnl_link_add(nls, veth, NLM_F_CREATE | NLM_F_EXCL);
+		if (err == 0) {
+			break;
+		}
+		WARN_ON(err < 0, "%s: unable to add link", nl_geterror(err));
```
sargun, line 134: Please add retrying + retry count.

inject/titus-inject-metadataproxy.c
```diff
@@ -126,7 +126,14 @@ static void setup_veth(int private_ns_fd, int container_ns_fd) {
 	rtnl_link_set_name(peer, IN_CONTAINER_INTERFACE_NAME);
 	rtnl_link_set_ns_fd(peer, container_ns_fd);
 
-	err = rtnl_link_add(nls, veth, NLM_F_CREATE | NLM_F_EXCL);
+	for (int i = 0; i < 5; i++) {
+		err = rtnl_link_add(nls, veth, NLM_F_CREATE | NLM_F_EXCL);
+		if (err == 0) {
+			break;
```
sargun, line 132:
```c
if (!err)
  break
```

sargun (commented):
See comments

inject/titus-inject-metadataproxy.c
```diff
@@ -126,7 +126,14 @@ static void setup_veth(int private_ns_fd, int container_ns_fd) {
 	rtnl_link_set_name(peer, IN_CONTAINER_INTERFACE_NAME);
 	rtnl_link_set_ns_fd(peer, container_ns_fd);
 
-	err = rtnl_link_add(nls, veth, NLM_F_CREATE | NLM_F_EXCL);
+	for (int i = 0; i < 5; i++) {
+		err = rtnl_link_add(nls, veth, NLM_F_CREATE | NLM_F_EXCL);
```
sargun, line 130: Can you dump the values of nls / veth? Is there an rtnl pretty print function?

rgulewich:
> How did you figure out the issue was around rtnl link add.

The error message being returned is `BUG:setup_veth:130: Operation not permitted: unable to add link`

inject/titus-inject-metadataproxy.c
solarkennedy: Just kinda doing my own research, never used libnl before.
Looking at the call graph of this function:
https://www.infradead.org/~tgr/libnl/doc/api/group__link.html#ga010cb3c6c598f7a60f97b67b8df9b227
I think the only place we could get his error is here when we send our request over the netlink socket:
https://www.infradead.org/~tgr/libnl/doc/api/nl_8c_source.html#l00251

Not unexpectedly, I think this just means it comes from the kernel (of course, since the thing that changed was the new kernel version).

From here I don't know how to trace into the kernel (`RTM_NEWLINK`?) what is going on.

If we could figure out what can cause perm denied in the kernel when we make new links like this, it may be able to answer the question of "will retries even help?"

Still though, I'm in favor of trying anything. Let's see if they help :)

inject/titus-inject-metadataproxy.c
solarkennedy: Yes please. Otherwise we'll have to like, count the number of times we see this for a task to know if it had to retry.
Or it could just say "Failed on the 5th attempt: permission denied"

inject/titus-inject-metadataproxy.c
rgulewich: You're correct that it's coming from the kernel, though my attempts to catch it via bpftrace haven't turned anything up (it's not `rtnl_newlink()` returing EPERM, which is unexpected).  My theory is that it's a race, so if retries make the problem go away, then we know for sure.

There's a pretty print function - it didn't seem to include much in the way of useful information, but I can add it just in case.

PR #512, opened 2021-01-14
Add hook to CNI to create symlink to netns mount

cconger:
We'll use this for making calls into the network namespace from the host.  The filename is to be a well-known format that can be determined from a pod-name.


alekar (approved):

rgulewich (approved):

solarkennedy:
Is the existing format not already constructable from the pod name? (task id)

cconger:
Closing as its part of #513

PR #513, opened 2021-01-16
Use CNI Based netns dialer for proxy mode on kubelet

cconger:
Detects if we're on a kubelet host and configures titus-logviewer to run in proxy mode.
Uses the CNI created netns aliases to dial into the pod for log viewing.

Depends on #512
Replaces #478

### DIFF
--- a/logviewer/api/proxy.go
+++ b/logviewer/api/proxy.go
@@ -24,10 +24,6 @@ func getNetNsPath(containerID string) string {
 	return fmt.Sprintf("%s/%s/ns/net", utils.TitusInits, containerID)
 }
 
-func getKubeletNetNsPath(podID string) string {
-	return fmt.Sprintf("/var/run/pod/netns-%s", podID)
-}
-
 func doProxy(w http.ResponseWriter, r *http.Request, containerID string, proxy *httputil.ReverseProxy) {
 	proxyURL, _ := url.Parse(inContainerURL)
 
--- a/logviewer/api/proxy.go
+++ b/logviewer/api/proxy.go
@@ -7,16 +7,25 @@ import (
 	"net/url"
 	"time"
 
-	"github.com/Netflix/titus-executor/utils/netns"
-
+	"github.com/Netflix/titus-executor/logviewer/conf"
 	"github.com/Netflix/titus-executor/utils"
+	"github.com/Netflix/titus-executor/utils/netns"
+	"github.com/Netflix/titus-executor/vpc/tool/cni"
 )
 
 // The URL to connect to once we've done a Setns() into the container
 const inContainerURL = "http://127.0.0.1:8004"
 
 func getNetNsPath(containerID string) string {
+	if conf.KubeletMode {
+		return cni.NSAliasPath(containerID)
+	}
+
 	return fmt.Sprintf("%s/%s/ns/net", utils.TitusInits, containerID)
+}
+
+func getKubeletNetNsPath(podID string) string {
+	return fmt.Sprintf("/var/run/pod/netns-%s", podID)
 }
 
 func doProxy(w http.ResponseWriter, r *http.Request, containerID string, proxy *httputil.ReverseProxy) {
--- a/logviewer/conf/conf.go
+++ b/logviewer/conf/conf.go
@@ -15,14 +15,6 @@ var (
 	KubeletMode        = false
 )
 
-const confTemplate = `Starting with config:
-ContainersHome     = %s
-ContainerID        = %s
-RunningInContainer = %t
-ProxyMode          = %t
-KubeletMode        = %t
-`
-
 func init() {
 	loc := os.Getenv("CONTAINER_HOME")
 	if loc != "" {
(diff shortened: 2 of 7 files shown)


vpc/tool/cni/cni.go
```diff
@@ -278,6 +278,12 @@ func (c *Command) Add(args *skel.CmdArgs) error {
 		return err
 	}
 
+	err = createNetNSAlias(pod.Name, args.Netns)
+	if err != nil {
+		logger.G(ctx).WithError(err).Error("Could not create symlink for network nameespace")
```
solarkennedy, line 283: namespace

logviewer/conf/conf.go
```diff
@@ -29,6 +30,12 @@ func init() {
 		ContainersHome = "/"
 	}
 
+	kubeMode := os.Getenv("KUBELET_MODE")
+	if kubeMode != "" {
+		log.Println("KUBELET_MODE set")
```
solarkennedy, line 35: prob don't need to print this

cconger: Was emulating the file, but happy to remove.

logviewer/api/proxy.go
```diff
@@ -7,18 +7,27 @@ import (
...
 
 func getNetNsPath(containerID string) string {
+	if conf.KubeletMode {
+		return cni.NSAliasPath(containerID)
+	}
+
 	return fmt.Sprintf("%s/%s/ns/net", utils.TitusInits, containerID)
 }
 
+func getKubeletNetNsPath(podID string) string {
```
solarkennedy, line 27: I don't see this function used?

logviewer/api/proxy.go
```diff
@@ -7,18 +7,27 @@ import (
...
+	"github.com/Netflix/titus-executor/utils/netns"
+	"github.com/Netflix/titus-executor/vpc/tool/cni"
 )
 
 // The URL to connect to once we've done a Setns() into the container
 const inContainerURL = "http://127.0.0.1:8004"
 
 func getNetNsPath(containerID string) string {
+	if conf.KubeletMode {
+		return cni.NSAliasPath(containerID)
```
solarkennedy, line 21: If we make a symlink here, just so we can use that alternate path in KubeleMode, why don't we just use the actual file in the first place instead of having an alternate mode and the symlink?

logviewer/conf/conf.go
rgulewich: Yeah, this was the precedent here, but it's a little weird.  It might be more valuable to consolidate all of the `log.Println()`s in this file into one line that prints out their current values.

rgulewich (approved):

logviewer/api/proxy.go
cconger: The symlink affords a "wellknown" name that we can devine from just a pod.  The actual filename contains an unrelated guid that I could not lookup without querying containerd which would mean I would have to query kubernetes first to get the container id to ask about.  The symlink seemed to be a fast way to add a weak reference by pod id, but open to other ideas for mapping between the two.

solarkennedy: Gotcha. I assumed that guid was actually the task name, but I guess that doesn't make much sense now that I know it is containerd doing it.

logviewer/conf/conf.go
cconger: Made a change to use a single logrus field powered log line.  On startup you'll see a line like this:
```
Jan 16 01:32:36 hostname titus-logviewer[3754328]: time="2021-01-16T01:32:36Z" level=info msg="Config loaded" ContainerID= ContainersHome=/var/lib/titus-container-logs KubeletMode=true ProxyMode=true RunningInContainer=false
```

logviewer/conf/conf.go
```diff
@@ -12,26 +12,37 @@ var (
...
+		KubeletMode = true
+	}
+
 	proxyMode := os.Getenv("DISABLE_PROXY_MODE")
 	if proxyMode != "" {
-		log.Println("DISABLE_PROXY_MODE set")
 		ProxyMode = false
 	}
+
+	log.WithFields(log.Fields{
```
solarkennedy, line 41: Cool

vpc/tool/cni/namespace_linux.go
```diff
@@ -0,0 +1,40 @@
...
+	if errors.Is(err, os.ErrNotExist) {
+		// Create base dir if missing
+		err := os.MkdirAll(filepath.Dir(aliasPath), 0755)
+		if err != nil {
+			return err
+		}
+		return os.Symlink(netnsPath, aliasPath)
+	}
+
+	return nil
```
solarkennedy, line 30: do you want to return `err` here? Otherwise the original error is swallowed. 
If it is nil, that is fine.

solarkennedy (approved):

logviewer/conf/conf.go
rgulewich: Thank-you!

alekar (approved):

PR #515, opened 2021-01-18
Added the Titus Seccomp Agent (tsa)

solarkennedy:
This new system sidecar (currenly vk-only) intercepts
syscalls on behalf of the container.

The first style of interception is to enable perf-related
syscalls (perf_event_open, bpf) where they were otherwise
not allowed (and wouldn't function even if they were allowed).

This feature is behind a container param:

    titusParameter.agent.seccompAgentEnabledForPerfSyscalls

That instructs the agent to setup the:

* Seccomp policy with perf/bpf allowed
* Env variables for tini to consume to relay the seccomp notify socket
* To spin up the tsa agent itself under systemd

The reason this is behind a perf-specific bool is that we intend
to expand the scope of this agent to someday include network
syscalls too, but not all containers are going to want both features.
Having different bools will allow us to enable new interceptions without
affecting containers that have opted in for only one kind of behavior.


executor/runtime/docker/seccomp/allow-perf-syscalls.json
```diff
@@ -0,0 +1,668 @@
...
+      "subArchitectures": ["SCMP_ARCH_MIPSEL", "SCMP_ARCH_MIPSEL64"]
+    },
+    {
+      "architecture": "SCMP_ARCH_S390X",
+      "subArchitectures": ["SCMP_ARCH_S390"]
+    }
+  ],
+  "syscalls": [
+    {
+      "names": [
```
rgulewich, line 35: Is this policy the same as default, but with `bpf` and `perf_event_open` added?

rgulewich (approved):

executor/runtime/docker/seccomp/allow-perf-syscalls.json
solarkennedy: Yep. This being the 3rd json file, maybe we need to come up with a "base" json file and then some describe the diff in some way? idk.

rgulewich: Good idea.  I think for now, it's fine - I was mostly checking to make sure I wasn't missing anything. :)

root/lib/systemd/system/titus-seccomp-agent@.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Titus seccomp agent for container %i
+# If the service restarts more than 10 times in 30 seconds, let it die
+StartLimitIntervalSec=30
+StartLimitBurst=10
```
sargun, line 5: Don't you want a conditionexists? as part of this?

sargun (approved):
Everything looks okay. I think you need to add a "conditionexists" to your systemd file. Maybe don't call it "perfsyscalls" and something like "tsaprivilegedsyscallgate"?

executor/runtime/docker/capabilities.go
```diff
@@ -51,6 +51,13 @@ func setupAdditionalCapabilities(c runtimeTypes.Container, hostCfg *container.Ho
 		seccompProfile = "fuse-container.json"
 	}
 
+	if c.SeccompAgentEnabledForPerfSyscalls() {
+		if c.FuseEnabled() {
+			return fmt.Errorf("Enabling the seccomp agent is currently not compatible with enabling Fuse too")
```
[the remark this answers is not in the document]
solarkennedy: Because we only get 1 json file to work with for the seccomp policy

root/lib/systemd/system/titus-seccomp-agent@.service
solarkennedy: what would the condition be, that a path exists or something?

PR #516, opened 2021-01-19
Log fewer lines when making requests to titus-isolate

rgulewich:
The `Titus Isolate did not return code 200` message gets logged a *lot*, and can be potentially misleading.  Log it less.

### DIFF
--- a/executor/runtime/docker/docker_isolate.go
+++ b/executor/runtime/docker/docker_isolate.go
@@ -24,21 +24,26 @@ func waitForTitusIsolateWithHost(parentCtx context.Context, taskID, host string,
 	defer cancel()
 
 	ticker := time.NewTicker(retryTime)
+	tries := 0
 
 	firstTry := make(chan struct{}, 1)
 	firstTry <- struct{}{}
 	for {
 		select {
 		case <-ctx.Done():
-			logrus.WithError(ctx.Err()).Warn("Context completed prior to getting a successful result from titus isolate")
+			logrus.WithError(ctx.Err()).WithField("tries", tries).Warn("Context completed prior to getting a successful result from titus isolate")
 			return false
 
 		case <-ticker.C:
+			tries++
 			if workloadIsolated(ctx, taskID, host) {
+				logrus.WithField("tries", tries).Info("Titus Isolate returned success")
 				return true
 			}
 		case <-firstTry:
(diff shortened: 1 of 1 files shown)


executor/runtime/docker/docker_isolate.go
```diff
@@ -82,6 +87,11 @@ func workloadIsolated(ctx context.Context, taskID, host string) bool {
...
 	}
 
+	if resp.StatusCode == 404 {
+		// Only log unexpected status codes below
+		return false
+	}
+
 	logrus.WithFields(map[string]interface{}{
 		"statusCode": resp.StatusCode,
 		"status":     resp.Status,
```
solarkennedy, line 97: Can you make it say `Titus isloate didn't return 200 after N tries?` ?

executor/runtime/docker/docker_isolate.go
```diff
@@ -82,6 +87,11 @@ func workloadIsolated(ctx context.Context, taskID, host string) bool {
 		return true
 	}
 
+	if resp.StatusCode == 404 {
+		// Only log unexpected status codes below
```
solarkennedy, line 91: Is this expected? why would we get a 404 after all these attempts?

solarkennedy (approved):

gabrielhartmann (approved):

sargun:
Why is it happening so much? Why is it misleading?

gabrielhartmann:
titus-executor polls titus-isolate very quickly. The current logging looks
like there's something wrong when it's really by design behavior.

executor/runtime/docker/docker_isolate.go
rgulewich: 404 is isolate's response when it hasn't completed the isolation yet.

executor/runtime/docker/docker_isolate.go
rgulewich: Yeah, I added the line about tries above.  I don't think it's important to log what try an unexpected request occurred on - only what was returned.

sargun:
Can we add a comment indicating that this poll loop is non-blocking on the titus isolate side?

rgulewich:
@sargun - https://github.com/Netflix/titus-executor/pull/521

PR #517, opened 2021-01-19
Make the ssh system sidecar optional

solarkennedy:
The ssh sidecar should be considered optional,
the show must go on if it can't be started for
whatever reason. (weird images, timeouts, who knows)

Additionally, this change makes it so if there is a timeout
starting a sidecar service, and it is not required, then
we continue anyway.

### DIFF
--- a/executor/runtime/docker/docker_linux.go
+++ b/executor/runtime/docker/docker_linux.go
@@ -64,15 +64,18 @@ var systemServices = []serviceOpts{
 		humanName:    "spectatord",
 		unitName:     "titus-spectatord",
 		enabledCheck: shouldStartSpectatord,
+		required:     false,
 	},
 	{
 		humanName: "atlas",
 		unitName:  "atlas-titus-agent",
+		required:  false,
 	},
 	{
 		humanName:    "ssh",
 		unitName:     "titus-sshd",
 		enabledCheck: shouldStartSSHD,
+		required:     false,
 	},
 	{
 		humanName: "metadata proxy",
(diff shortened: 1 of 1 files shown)


executor/runtime/docker/docker_linux.go
```diff
@@ -254,9 +255,12 @@ func startSystemdUnit(ctx context.Context, conn *dbus.Conn, taskID string, cID s
 	select {
 	case <-ctx.Done():
 		doneErr := ctx.Err()
-
 		if doneErr == context.DeadlineExceeded {
-			return errors.Wrapf(doneErr, "timeout starting %s service", opts.humanName)
+			if opts.required {
+				return errors.Wrapf(doneErr, "timeout starting %s service", opts.humanName)
+			}
+			l.Errorf("timeout starting %s service (not required to launch this task)", opts.humanName)
```
sargun, line 262: please use `withField`

sargun (approved):

sargun:
Won't this implicitly allow all sidecars without required set at all to timeout since it defaults to false?

rgulewich (approved):

solarkennedy:
> Won't this implicitly allow all sidecars without required set at all to timeout since it defaults to false?

Yep (intended)! Would you like me to explicitly set `required` on all the sidecar configs?
Right now the ones I think are required are set.

executor/runtime/docker/docker_linux.go
solarkennedy: Will adding another field actually provide much value?
`l` already has the unitName (as well as taskid and containerid)

If we add the humanName too as a field, we'll have something like `ssh` (human name) *and* `titus-sshd` (unit name), I think I would rather have unitName consistently.

PR #522, opened 2021-01-21
Make log profile IAM credentials available within the pod via the imds proxy.

cconger:
Add second IAM proxy for log profile.

We need to provide credentials to the log upload profile in the network namespace of the pod. We add an env variable to the imds proxy to add a second assume role proxy if started by the imds-cni hooks.

These credentials are available at /latest/meta-data/iam/logging/...

Additionally added a credential provider to the uploader package for loading them.  This is just a helpful convenience for customers who might want to load the creds.

For more context review this [problem doc](https://docs.google.com/document/d/1voOrKrXEhnetyRWNLRfADTyVo9_8XUXgfciZ9t17f5I/)


uploader/creds.go
```diff
@@ -0,0 +1,140 @@
...
+	SecretAccessKey string
+	Token           string
+
+	// Error state
+	Code    string
+	Message string
+}
+
+func requestCredList(ctx aws.Context, client *ec2metadata.EC2Metadata) ([]string, error) {
+	resp, err := client.GetMetadataWithContext(ctx, credentialPath)
```
sargun, line 90: Any reason you have to implement your own AWS endpoint resolver, and you can't just supply an alternate endpoint to the one constructed by ec2metadatutils when you configure it? You should be able to have a custom endpoint resolver, and it will do all the renew / retry magic for you:
https://docs.aws.amazon.com/sdk-for-go/api/aws/#Config.WithEndpoint

cmd/titus-imds-cni/env.go
```diff
@@ -106,6 +107,7 @@ func extractEnv(prev *current.Result, pod *v1.Pod) (map[string]string, error) {
 
 	env["TITUS_IAM_ROLE"] = pod.Annotations[mt.IamRoleArnAnnotation]
 	env["X_FORWARDED_FOR_BLOCKING_MODE"] = pod.Annotations[mt.XForwardedForBlockingModeAnnotation]
+	env["TITUS_LOG_IAM_ROLE"] = pod.Annotations[podCommon.AnnotationKeyLogS3WriterIAMRole]
```
sargun, line 110: Is this always set, what if it's blank?

sargun (commented):
I think that there is a simpler way to do this -- you should be able to override the creds the ec2metadatautils credential provider. If this isn't built-in / the case, I think we should take the time to add it. 

The other question I have:
1. Why don't we always wire up the /logging route on the IMDS, and if the logging role is equal to the normal role, then just point them at the same handler?
2. An alternate idea might be to use a special header, or user agent to indicate that this is coming from the logging daemon and the IMDS does the right thing and hands back the right creds.

My worry is this will double our requests to STS service as well. Do you have any idea how pervasive this capability will be?

cmd/titus-imds-cni/env.go
cconger: If its blank the second iam proxy will not bind, I liked this behavior because if we deploy today nothing changes until this annotation starts getting set by the mutating web hook.

https://github.com/Netflix/titus-executor/pull/522/files#diff-7b7c269c93709bb9c141084f83098abbd1aa53dc7c294a961d5d84a263bfdfccR140

sargun: Any idea what % of containers will have this set?

cconger: I believe 100% of containers deployed on the real kubelet.

uploader/creds.go
cconger: I'd love to not have to re-implement much of this.  I'm not sure I see how the endpoint would help here since the path that I'm overriding here was similarly a [const in the aws-sdk](https://github.com/aws/aws-sdk-go/blob/master/aws/credentials/ec2rolecreds/ec2_role_provider.go#L137).  It could very easily afford configuration but it does not currently.

Are you suggesting that we host a entire duplicate endpoint for the logging case?  I'm open to the idea just seems like there's more work to be done to do that instead of just providing the provider here.

cconger:
> * Why don't we always wire up the /logging route on the IMDS, and if the logging role is equal to the normal role, then just point them at the same handler?

This seems a reasonable change, might as well always provide it and have it fall through to the task profile.

> * An alternate idea might be to use a special header, or user agent to indicate that this is coming from the logging daemon and the IMDS does the right thing and hands back the right creds.

Yes this likely will increase sts load.  I only say likely because as far as I could tell the agents were doing this previously every time a uploader was created in the executor https://github.com/Netflix/titus-executor/blob/master/executor/runner/runner.go#L372

We talked to Mike Grima about this on Tuesday and he suggested that if we're only doing it once per agent we're probably ok.  I admittedly didn't worry too much since our workloads on real kubelet are 0 right now.  I believe we could add logic to do agent level caching of these creds since almost all containers will be using the same profile on the same agent.

@alekar curious for you to weigh in on this.

uploader/creds.go
sargun: Hrrrm. That seems to have different behaviour than every other AWS resolver where it appends to the endpoint that the thing responds with.

Something like this should work though:
```
package main

import (
	"fmt"
	"github.com/aws/aws-sdk-go/aws"
	"github.com/aws/aws-sdk-go/aws/credentials/ec2rolecreds"
	"github.com/aws/aws-sdk-go/aws/request"
	"github.com/aws/aws-sdk-go/aws/session"
	"log"
	"os"
)
import "github.com/aws/aws-sdk-go/aws/ec2metadata"

func main() {
	log.SetOutput(os.Stderr)

	cfg := (&aws.Config{
	}).WithLogLevel(aws.LogDebugWithHTTPBody | aws.LogDebugWithRequestRetries | aws.LogDebugWithRequestErrors | aws.LogDebugWithEventStreamBody | aws.LogDebugWithSigning)
	s := session.Must(session.NewSessionWithOptions(session.Options{
		Config: *cfg,
		EC2IMDSEndpoint: "http://localhost:1234",
	}))
	s.Handlers.Build.PushBack(func(request *request.Request) {
		log.Println("original url: %s", request.HTTPRequest.URL.String())
		request.HTTPRequest.URL.Path = "/logging" + request.HTTPRequest.URL.Path
		log.Println("new url: %s", request.HTTPRequest.URL.String())
	})
	md := ec2metadata.New(s)
	creds := ec2rolecreds.NewCredentialsWithClient(md)
	val, err :=creds.Get()
	fmt.Println(val)
	fmt.Println(err)
}
```

See the output:
```
sargun:imds sargun$ go run .
2021/01/21 00:22:31 original url: %s http://localhost:1234/latest/meta-data/iam/security-credentials/
2021/01/21 00:22:31 new url: %s http://localhost:1234/logging/latest/meta-data/iam/security-credentials/
2021/01/21 00:22:31 original url: %s http://localhost:1234/latest/api/token
2021/01/21 00:22:31 new url: %s http://localhost:1234/logging/latest/api/token
2021/01/21 00:22:31 DEBUG: Request ec2metadata/GetToken Details:
---[ REQUEST POST-SIGN ]-----------------------------
PUT /logging/latest/api/token HTTP/1.1
Host: localhost:1234
User-Agent: aws-sdk-go/1.36.29 (go1.13.14; darwin; amd64)
Content-Length: 0
X-Aws-Ec2-Metadata-Token-Ttl-Seconds: 21600
Accept-Encoding: gzip


-----------------------------------------------------
2021/01/21 00:22:31 DEBUG: Send Request ec2metadata/GetToken failed, attempt 0/2, error RequestError: send request failed
caused by: Put http://localhost:1234/logging/latest/api/token: dial tcp [::1]:1234: connect: connection refused
2021/01/21 00:22:31 DEBUG: Retrying Request ec2metadata/GetToken, attempt 1
2021/01/21 00:22:31 DEBUG: Request ec2metadata/GetToken Details:
---[ REQUEST POST-SIGN ]-----------------------------
PUT /logging/latest/api/token HTTP/1.1
Host: localhost:1234
User-Agent: aws-sdk-go/1.36.29 (go1.13.14; darwin; amd64)
Content-Length: 0
X-Aws-Ec2-Metadata-Token-Ttl-Seconds: 21600
Accept-Encoding: gzip
```

I 
```

Comment (sargun) [reply uploader/creds.go:90]: You shouldn't rewrite the `/logging/latest/api/token` endpoint, so maybe a *tiny* bit more work required. But, yeah, I think the above approach should work of adding a hook in the session.

Alternatively, the user agent or headers trick means no rewriting required, just injection of the info.

**Comment by sargun** [comment]
I think we can handle the scaling concern if it's gradual, so I'm not worried about scale today. More so fewer code paths of how the server gets setup, and how logging creds are retrieved. I like the idea that the logger client supplies a custom header on fetching creds, or a custom user agent, and the IMDS knows to respond with different creds. That way you could test the log uploader on a vanilla EC2 instance because the vanilla IMDS would just ignore it.

**Comment by alekar** [comment]
I think we can start w/o caching. Then move to per instance caching if required as we discussed w/ Mike. Hopefully off-box STS will get prioritized and we won't have to worry about STS rate limiting.

**Review by alekar — CHANGES_REQUESTED** [review]
Can you create a _different_ pull request for the vendoring changes?

**Review by alekar — APPROVED** [review]

**Review by solarkennedy — APPROVED** [review]
It kinda sucks that we have to make the imds proxy do *two* things like this, but I don't know of a better way to support what we want to do.
I hope it doesn't lead down a slippery slope of having N iams.

On the plus side though, *we* control the producer (imds), and the consumer (log viewer), so we can change this in the future.
This can't turn into a public interface as a weird way to get genpop creds from every container at netflix :(

Too bad we don't have a socat running on our agents that does:

    socat tcp:169.xxx:80 -> unix:/var/lib/titus-sockets/agent-imds.sock

And then sidcars that need it could get it? Probably worse and terrible.

File: metadataserver/iam.go [hunk 3]
```diff
@@ -98,9 +95,44 @@ func newIamProxy(ctx context.Context, config types.MetadataServerConfiguration)
...
+	}
+
+	return proxy
+}
+
+func newRoleProxy(ctx context.Context, iamARN, taskID string, stsClient *sts.STS) *roleProxy {
+	/* This will automatically use *our* metadata proxy to setup the IAM role. */
+	parsedArn, err := arn.Parse(iamARN)
+	if err != nil {
+		log.Fatal("Unable to parse ARN: ", err)
```
sargun, line 126: log.witherror(err).fatal("unable to parse arn")

metadataserver/server.go
```diff
@@ -195,6 +196,9 @@ func (ms *MetadataServer) installIMDSCommonHandlers(ctx context.Context, router
 
 	/* IAM Stuffs */
 	ms.iamProxy.AttachRoutes(metaData.PathPrefix("/iam").Subrouter())
+	if ms.logIAMProxy != nil {
+		ms.logIAMProxy.AttachRoutes(metaData.PathPrefix("/iam/logging").Subrouter())
```
sargun, line 200: Do we still need this code path if we have the header code path?

sargun (changes requested):

sargun (approved):

metadataserver/server.go
cconger: No, removing.

PR #523, opened 2021-01-21
Populate status.PodIPs, and add EIP to it

rgulewich:
This makes VK consistent with how RK bubbles up IP addresses in the pod's status object.

### DIFF
--- a/executor/runtime/types/types.go
+++ b/executor/runtime/types/types.go
@@ -301,12 +301,13 @@ type Resources struct {
 
 // NetworkConfigurationDetails used to pass results back to master
 type NetworkConfigurationDetails struct {
-	IsRoutableIP   bool
-	IPAddress      string
-	EniIPAddress   string
-	EniIPv6Address string
-	EniID          string
-	ResourceID     string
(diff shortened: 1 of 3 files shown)


alekar (approved):

executor/runtime/types/types.go
```diff
@@ -319,6 +320,9 @@ func (n *NetworkConfigurationDetails) ToMap() map[string]string {
 	if n.EniIPv6Address != "" {
 		m["EniIPv6Address"] = n.EniIPv6Address
 	}
+	if n.ElasticIPAddress != "" {
+		m["ElasticIPAddress"] = n.ElasticIPAddress
```
solarkennedy, line 324: Huh, do you know if we are ever going to have ElasticIPv6Addresses?

solarkennedy (approved):

executor/runtime/types/types.go
rgulewich: Not sure - they don't exist today: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-instance-addressing.html

PR #527, opened 2021-01-29
Fix accountId in kubelet IMDS instance-identity document

solarkennedy:
The NETFLIX_ACCOUNT_ID is set by TJC, we can't use whatever it is
set to on the agent (could be any crazy account ID).

The IMDS uses this when it returns data for

    curl http://169.254.169.254/latest/dynamic/instance-identity/document

Previously this was unset, causing `accountId` to be the empty string.
With this change it will show the account ID set to whatever is set
in the main container's environment.

### DIFF
--- a/cmd/titus-imds-cni/env.go
+++ b/cmd/titus-imds-cni/env.go
@@ -80,12 +80,10 @@ func extractEnv(prev *current.Result, pod *v1.Pod) (map[string]string, error) {
 	env := map[string]string{}
 
 	env["LISTEN_PORT"] = "80"
-
 	env["PEER_NAMESPACE"] = getPeerNsPath(pod.Name)
 	env["EC2_REGION"] = os.Getenv("EC2_REGION")
 
 	var ok bool
-
 	for _, ip := range prev.IPs {
 		switch ip.Version {
 		case "4":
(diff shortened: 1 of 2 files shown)


cmd/titus-imds-cni/env.go
```diff
@@ -96,21 +94,27 @@ func extractEnv(prev *current.Result, pod *v1.Pod) (map[string]string, error) {
...
 	if mEnvVal, ok := getEnvVal(c, mt.TitusMetatronVariableName); ok {
 		env[mt.TitusMetatronVariableName] = mEnvVal
 	}
+	// We can't use the NETFLIX_ACCOUNT_ID variable present in the global env,
+	// as that represents at titus_agent account ID. For the IMDS, we want to
+	// reflect the account id of the *pod*, so for that we reach in and copy it.
+	if mEnvVal, ok := getEnvVal(c, mt.NetflixAccountIDVarName); ok {
+		env[mt.NetflixAccountIDVarName] = mEnvVal
+	} else {
+		env[mt.NetflixAccountIDVarName] = fmt.Sprintf("Unknown, %s not set for the main container", mt.NetflixAccountIDVarName)
```
alekar, line 116: I prefer something that looks like an error code more than an error message here (think `UNKNOWN_ACCOUNT`) since it is an environment variable. Or leaving it empty.

alekar (approved):

cmd/titus-imds-cni/env.go
solarkennedy: Yea... I also had that original inclination, but then I couldn't find any good reason (other than style) to not give more clue about *why* the variable is unknown. It is an environment variable here, but it in the end it ends up in the JSON of apps that are reading imds stuff. 

Unset also sounds like a very sane thing to do, but it is that kind of silent error behavior that made it so hard to find this error in the first place!

alekar: I think it's a little more than style. Instead of seeing an unusable constant string, downstream code could see weird escaping/ parsing errors with an error message like that. That said, I don't think it matters so much here.

PR #532, opened 2021-02-08
Update generated ssh authorized_principals to be consistent with base os.

JackOfMostTrades:
This updates the generated authorized_principals files to expect the v3 SSH principals. This more generally makes the generated files to consistent with base os (which just have the one expected principal and the emergency access principal).

Note that config.Config.SSHAccountID could be dropped since we only care about the environment in the new principal, but for some reason the SSH account ID seems to be informing the VPC account ID and I didn't want to go poking around in those bits of code that I don't know about.

### DIFF
--- a/executor/runtime/docker/docker_ssh.go
+++ b/executor/runtime/docker/docker_ssh.go
@@ -8,8 +8,7 @@ import (
 
 	"github.com/Netflix/titus-executor/config"
 	runtimeTypes "github.com/Netflix/titus-executor/executor/runtime/types"
-	"github.com/aws/aws-sdk-go/aws/arn"
-	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
 )
 
 const sshdConfig = `
@@ -90,24 +89,14 @@ PidFile /run/sshd.pid
 `
 
 func addContainerSSHDConfig(c runtimeTypes.Container, tw *tar.Writer, cfg config.Config) error {
-	iamProfileARN := c.IamRole()
-	if iamProfileARN == nil {
-		return errors.New("Could not get IAM role from container")
-	}
-
-	iamProfile, err := arn.Parse(*iamProfileARN)
-	if err != nil {
-		return err
-	}
-
 	caData, err := ioutil.ReadFile(cfg.ContainerSSHDCAFile)
 	if err != nil {
 		return err
 	}
-	return addContainerSSHDConfigWithData(c, tw, cfg, caData, iamProfile.AccountID, cfg.SSHAccountID)
+	return addContainerSSHDConfigWithData(c, tw, cfg, caData)
 }
 
-func addContainerSSHDConfigWithData(c runtimeTypes.Container, tw *tar.Writer, cfg config.Config, caData []byte, accountIDs ...string) error {
+func addContainerSSHDConfigWithData(c runtimeTypes.Container, tw *tar.Writer, cfg config.Config, caData []byte) error {
 	sshConfigBytes := []byte(sshdConfig)
 	err := tw.WriteHeader(&tar.Header{
 		Name: "/titus/etc/ssh/sshd_config",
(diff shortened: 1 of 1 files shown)


executor/runtime/docker/docker_ssh.go
```diff
@@ -135,21 +123,12 @@ func addContainerSSHDConfigWithData(c runtimeTypes.Container, tw *tar.Writer, cf
...
-		for _, accountID := range accountIDs {
-			lines = append(
-				lines,
-				fmt.Sprintf("%s:%s:%s:%s", username, c.AppName(), accountID, c.TaskID()), // key scoped to username, appname, account ID, and task ID
-				fmt.Sprintf("%s:%s:%s", c.AppName(), accountID, c.TaskID()),              // key has access to any username for this given app in this given account, with this task ID
-				fmt.Sprintf("%s:%s", c.AppName(), accountID),                             // key has access to any username for this given app in this given account
-				c.TaskID(), // key has access to any username on this task ID
-				fmt.Sprintf("%s:%s", username, c.TaskID()), // key has access to this given username on this task ID
-			)
+		lines := []string{
```
solarkennedy, line 129: We are already on v3 everywhere server-side?

executor/runtime/docker/docker_ssh.go
```diff
@@ -135,21 +123,12 @@ func addContainerSSHDConfigWithData(c runtimeTypes.Container, tw *tar.Writer, cf
 		return err
 	}
 
-	// The format that is used for SSH Users is:
-	// $(unix username):$(app name):$(aws account id):$(task id)
-
+	containerEnv := c.Env()["NETFLIX_ENVIRONMENT"]
```
solarkennedy, line 126: Are you ok with this defaulting to "" if we don't have this variable?

If not, can you add a log line if !ok to say "SSH will not work!"?

executor/runtime/docker/docker_ssh.go
```diff
@@ -135,21 +123,12 @@ func addContainerSSHDConfigWithData(c runtimeTypes.Container, tw *tar.Writer, cf
...
-			lines = append(
-				lines,
-				fmt.Sprintf("%s:%s:%s:%s", username, c.AppName(), accountID, c.TaskID()), // key scoped to username, appname, account ID, and task ID
-				fmt.Sprintf("%s:%s:%s", c.AppName(), accountID, c.TaskID()),              // key has access to any username for this given app in this given account, with this task ID
-				fmt.Sprintf("%s:%s", c.AppName(), accountID),                             // key has access to any username for this given app in this given account
-				c.TaskID(), // key has access to any username on this task ID
-				fmt.Sprintf("%s:%s", username, c.TaskID()), // key has access to this given username on this task ID
-			)
+		lines := []string{
+			fmt.Sprintf("~v3:titus:%s:%s:%s:%s:%s:%s", username, c.AppName(), containerEnv, c.TaskID(), c.JobGroupStack(), c.JobGroupDetail()),
```
solarkennedy, line 130: Kinda same deal here, except this ends up in the file itself.

Can you add a Comment string (not a golang comment, a `fmt.Sprintf("# This line allows Bla, and if any fields are empty, then ssh won't work. The "bless" infrastructure at Netflix is what generates certificates with these principals...`), just a nice to have for the next person looking at the file wondering where these magic strings came from.

solarkennedy (approved):
This will break ssh access to non-spinnaker tasks, right?

I think they were working before because of the more lax principal lines.

executor/runtime/docker/docker_ssh.go
JackOfMostTrades: The 2.0 flow (go/ssh) is producing certs including the v3 principals now. Probably the bastion 1.0 / titus-ssh infrastructure is not, although that should be getting torn down entirely in the next few weeks.

executor/runtime/docker/docker_ssh.go
JackOfMostTrades: Yea, let me add an update here. I'll add something similar to the EC2 script which logs and just leaves the emergency principal.

executor/runtime/docker/docker_ssh.go
JackOfMostTrades: :+1: Will add a link to the bless docs.

JackOfMostTrades:
> This will break ssh access to non-spinnaker tasks, right?
> 
> I think they were working before because of the more lax principal lines.

I believe this will still work for non-spinnaker tasks. As long as we can resolve the asg (or equivalently job stack/detail) when we do a sonar lookup the server-side should be inserting matching principals.

executor/runtime/docker/docker_ssh.go
russell-lewis: I believe some of the above principals were used by the titus team themselves for power user access. @solarkennedy do you know if y'all still use the app:account, taskId, or user:taskId formats for your custom CA tooling?

Removing support for the v1 line, app:account:taskID will prevent SSH Certs for tasks with non-spinnaker compatible stack/detail names from accepting certs issued to them.

JackOfMostTrades: > Yeah, merging this would break the bastion 1.0 users, but if we need to support that before the tear down, let me know and we can patch it.

Yea, bastion 1.0 is to be torn down very imminently, so we can probably save the effort involved.

> Removing support for the v1 line, app:account:taskID will prevent SSH Certs for tasks with non-spinnaker compatible stack/detail names from accepting certs issued to them.

At least in the 2.0 flow we lookup the stack/detail via the titus API and put them in the cert, so even if it's a non-spinnaker compatible name as long as it doesn't have a colon it should still work.

> I believe some of the above principals were used by the titus team themselves for power user access.

Yea, that's the main think that's a bit unclear to me. If there's some other tooling making use of these principals I don't want to break that. But I pulled the thread on the trusted CA keys and it's just the bless keys, so I believe whatever tooling it is would have to be calling the 1.0 BLESS service.

solarkennedy (approved):

solarkennedy:
Well, you know how this goes :)
It will almost certainly break some things.

* Can we know beforehand if this is going to break things that use titus_agent accounts?
* Can we know beforehand if this will break things that use an IAM ARN from a different account?
* Can we know if this will break non-spinnaker tasks (kinda covered, and I guess the answer is .. maybe? blanks with colons?)

Also I can deploy this to my dev stack if you want, give it a real test on something.

executor/runtime/docker/docker_ssh.go
russell-lewis: Ah, thanks for pulling the CA keys trust set. There was a time when "alternative" CAs were being added to the mix. Nothing on the bless CAs has used those other formats.

JackOfMostTrades:
> * Can we know beforehand if this is going to break things that use titus_agent accounts?
> * Can we know beforehand if this will break things that use an IAM ARN from a different account?
> * Can we know if this will break non-spinnaker tasks (kinda covered, and I guess the answer is .. maybe? blanks with colons?)

To the extent that we're talking about bastion 2.0 flow none of these things should break. The ssh certificate service was only ever injecting the test/prod account IDs, regardless off the titus_agent accounts or IAM ARN account ID. Likewise the certificate service was already rejecting requests to mint certs for jobs with app/stack/detail values that had colons, so those already never worked.

More subtly, there may have been cases with some different behavior (such as colon values working) in the bastion 1.0 titus-ssh workflow. Since this effectively forcefully cuts us over to the 2.0 flow this would mean there's no (easy) going back to titus-ssh for corner cases where the new flow intentionally doesn't work. But I would have expected those sort of mentions to come up regardless of these changes. So I would estimate disruption to be unlikely.

PR #533, opened 2021-02-12
Titus standalone

sargun:


cmd/titus-standalone/pod.jsonnet
```diff
@@ -0,0 +1,140 @@
+{
+  "metadata": {
```
solarkennedy, line 2: What is your thinking behind leaving this here.
Don't you think it will just be stale and not work?

cmd/titus-standalone/main.go
```diff
@@ -0,0 +1,110 @@
...
+
+	dockerCfg, dockerCfgFlags := docker.NewConfig()
+	app.Flags = append(app.Flags, dockerCfgFlags...)
+
+	app.Action = func(c *cli.Context) error {
+		return cli.NewExitError(mainWithError(c, dockerCfg, cfg, &options), 1)
+	}
+
+	if err := app.Run(os.Args); err != nil {
+		panic(err)
```
solarkennedy, line 58: Do we have to panic?
What kind of errors do we get from this, can we just print to stderr and exit non-zero?

solarkennedy (approved):
Neat!

cmd/titus-standalone/pod.jsonnet
sargun: The idea was to give someone a thing to work from. What do you think?

solarkennedy: I think we should remove all the things we can, strip it to bare bones

solarkennedy:
Really looking forward to using this!

PR #535, opened 2021-02-14
Remove vendor and alter Makefile to vendor on demand

sargun:
vendor/ is a pure pain. At this point, I don't see the point in checking it in since we so frequently let it desync.


sargun:
[Docker hub is currently rate limiting the buildkite tests...]

solarkennedy:
Gosh.
I guess this is fine, because in practice we don't even build against this folder. (right? no mod=vendor)

Given that, what is the point of this folder actually?

sargun:
I think there are three thoughts:
1. Before go 1.13, you still needed this folder
2. It's nice to be able to run grep
3. It's a fallback in case upstream accidentally disappears.

rgulewich:
> 2. It's nice to be able to run grep
> 3. It's a fallback in case upstream accidentally disappears.

Do we think these give us enough benefit to continue vendoring (@sargun, @solarkennedy)?  If so, maybe we could add some commit hooks, or a check in the build to make sure things are vendored correctly?

If not, I'm game for removing it.

rgulewich (approved):

sargun:
@rgulewich You can always vendor locally to be able to run grep.

rgulewich:
> @rgulewich You can always vendor locally to be able to run grep.

I more meant #3, but sure. :)

PR #536, opened 2021-02-15
Limit parallel to 1

sargun:


sargun:
I'm going to merge this to fix the build. We can back it out later or adjust it if need be.

PR #537, opened 2021-02-16
WIP: titus-storage system service with ebs support

solarkennedy:
This adds a new system sidecar: `titus-storage`.
The goal of this sidecar is to mount all external storage,
ebs to start, and nfs later (or if you want I can do nfs first?).

I think there could be few nice things about doing it this way:

* Long term could be externalized into a "true" external sidecar
* Way easier to troubleshoot / debug (it is a systemd unit, not an
  internal titus-executor thing)

Some downsides:

* Being external to the titus-executor codebase, most of our message
  passing comes via environment variables (Hard to do N ebs mounts?)
* I don't know what to think of having multiple mount binaries, but
  same deal, could be externalized to other teams some day
* This system service *must* wait, otherwise there isn't a good
  way for it to "stop" and detach.
* All attach/detach is happening client side, not centrally orchestrated
  like the aws-ebs-csi driver.

This POC is in bash because a lot of bash code existing already.
For production I would rewrite it in golang, with real error codes
and not having to "tail" to wait for a pid.

Also in this POC I'm hard-coding the volume ID. I'll need to get that
from the task context somehow. Also somehow I need to extract
mount, rw, etc preferences.

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -980,6 +980,13 @@ func (r *DockerRuntime) Prepare(parentCtx context.Context) error { // nolint: go
 		}
 	}
 
+	if shouldStartTitusStorage(&r.cfg, r.c) {
+		// KYLE figure out how to... get this from the taskContext?
+		r.c.SetEnvs(map[string]string{
+			"TITUS_EBS_VOLUME_ID": "vol-0cc20fbab97755b62",
+		})
+	}
+
 	if r.cfg.UseNewNetworkDriver {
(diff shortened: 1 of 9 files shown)


solarkennedy:
This was a good POC.
I'm going to decline this one and port it to go where I can have real logging and real functions.

But I'll continue with the method of having a standalone c binary to do the crazy mount syscalls + namespace, and a `titus-storage` system sidecar with start/stop hooks to do the orchestration

rgulewich:
Big 👍 from me on moving in this direction - thanks for prototyping it!

PR #541, opened 2021-02-16
Update abmetrix service definitions for new version

siow4096:
### Description of the Change

The newest version of `abmetrix` moves to a periodic trigger rather than a long-lived daemon. This PR updates the service file(s) to match those included in the BaseOS version.

### DIFF
--- a/root/lib/systemd/system/titus-abmetrix@.timer
+++ b/root/lib/systemd/system/titus-abmetrix@.timer
@@ -1,0 +1,10 @@
+[Unit]
+Description=Timer for refreshing abmetrix state
+
+[Timer]
+Unit=abmetrix.service
+OnUnitActiveSec=21600
+Persistent=true
+
+[Install]
+WantedBy=multi-user.target
--- a/root/lib/systemd/system/titus-abmetrix@.service
+++ b/root/lib/systemd/system/titus-abmetrix@.service
@@ -1,5 +1,6 @@
 [Unit]
 Description=Titus abmetrix for container %i
+Wants=abmetrix.timer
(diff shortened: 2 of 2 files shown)


solarkennedy:
Huh, do we have any prior art for this?
Would you like to (or have you already) tested this on a real agent?

siow4096:
The prior art for this change is matching the definitions of other timer-based services in `titus-executor` as well as basing the service files off the same service for the Netflix base OS.

I haven't tested this on a real agent deployment yet, but would be interested in doing so. Happy to follow any instructions around how to perform such a test.

solarkennedy:
@sargun this (testing this locally) is another great use case for the titus-standalone binary.

@siow4096 please ping me on slack and let's coordinate an adhoc test, as we currently don't have an easy way to test it without putting this change on a dev stack.

root/lib/systemd/system/titus-abmetrix@.timer
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=Timer for refreshing abmetrix state
+
+[Timer]
+Unit=abmetrix.service
+OnUnitActiveSec=21600
+Persistent=true
+
+[Install]
```
solarkennedy, line 9: I don't think we want this part, the main target doesn't want this installed

solarkennedy:
I don't think this is going to work without more changes to docker_linux.go:106.

Look for `qualifiedUnitName`, and you can see we hard code `.service` because we assume that is all we will ever be starting.

For this change, we would *not* start the service, and instead start the timer, and let it start the service.

solarkennedy:
As discussed, this PR is now obsolete with the new changes to make abmetrix behave as it did before (as a daemon)

PR #544, opened 2021-02-18
Put a 5 second (30 if a required service) timeout on system services

solarkennedy:
We've seen this in the wild before, where one system service
that is slow to start (even if not required), ends up blocking
everything, which is "fine".

But then a later systemd unit will started, but "fail" instantly,
even though it actually did work.

The reason this is happening is because they are hitting the overall
task deadline.

This PR adds a per-system-service start timeout, which will prevent
us from using up our entire task start time budget on a single slow
(in my case abmetrix, which is not required) startup.

### DIFF
--- a/executor/runtime/docker/docker_linux.go
+++ b/executor/runtime/docker/docker_linux.go
@@ -255,7 +255,12 @@ func startSystemdUnit(ctx context.Context, conn *dbus.Conn, taskID string, cID s
 		return err
 	}
 
-	l.Infof("Starting systemd unit %s", qualifiedUnitName)
+	timeout := 5 * time.Second
+	if opts.required {
+		timeout = 30 * time.Second
+	}
+
+	l.Infof("starting systemd unit %s", qualifiedUnitName)
 	ch := make(chan string, 1)
 	if _, err := conn.StartUnit(qualifiedUnitName, "fail", ch); err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


executor/runtime/docker/docker_linux.go
```diff
@@ -265,9 +270,9 @@ func startSystemdUnit(ctx context.Context, conn *dbus.Conn, taskID string, cID s
 		doneErr := ctx.Err()
 		if doneErr == context.DeadlineExceeded {
 			if opts.required {
-				return errors.Wrapf(doneErr, "timeout starting %s service", opts.humanName)
+				return errors.Wrapf(doneErr, "timeout (overall task start dealine exceeded) starting %s service", opts.humanName)
 			}
-			l.Errorf("timeout starting %s service (not required to launch this task)", opts.humanName)
+			l.Errorf("timeout (overall task start dealine exceeded) starting %s service (not required to launch this task)", opts.humanName)
```
rgulewich, line 275: Thinking about this a bit: mind changing this to do a `.WithField()` for the service name, so we can more easily search for these log lines in ES and generate a list of top offenders (and the same for the log line below)?  I know this is the way it was before, but this is probably a good time to do it, since you're already in here.

rgulewich (approved):

sargun:
Do we know if Metatron starts up quickly enough for this?

executor/runtime/docker/docker_linux.go
solarkennedy: `l` is already:

```
	l := logrus.WithFields(logrus.Fields{
		"containerId": cID,			"containerId": cID,
		"taskID":      taskID,			"taskID":      taskID,
		"unit":        opts.unitName,			"unit":        opts.unitName,
	})		})
```

solarkennedy:
> Do we know if Metatron starts up quickly enough for this?

30 seconds?

Not sure. Keep in mind this time is essentially "the time it takes to run systemctl start bla"

It doesn't look like metatrond is a one-shot, nor does it notify, so it (the systemctl start) really should return in 30s!

sargun (approved):

PR #545, opened 2021-02-18
Make the waitForTitusIsolate function properly timeout

solarkennedy:
The select call is pseudo-random, you can't control
which messages it selects.

In prod, we had instances where this function was never
timing out, expending the entire task budget time
waiting for titus isolate.

I believe this was because the ticker spammed the channel,
and the done message never got in on time.

This PR makes it so we give Done a fighting chance, based off
of https://stackoverflow.com/questions/46200343/force-priority-of-go-select-statement/46202533#46202533

Playground repro:
https://play.golang.org/p/soqUFWhZ7sj

And a solution:
https://play.golang.org/p/Dw96oMVZJst

Internal ticket TITUS-5023

### DIFF
--- a/executor/runtime/docker/docker_isolate.go
+++ b/executor/runtime/docker/docker_isolate.go
@@ -30,11 +30,15 @@ func waitForTitusIsolateWithHost(parentCtx context.Context, taskID, host string,
 	firstTry <- struct{}{}
 	for {
 		select {
-		case <-ctx.Done():
-			logrus.WithError(ctx.Err()).WithField("tries", tries).Warn("Context completed prior to getting a successful result from titus isolate")
-			return false
-
 		case <-ticker.C:
+			// select is pseudo-random, so simply selecting for ctx.Done() is not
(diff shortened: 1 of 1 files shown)


executor/runtime/docker/docker_isolate.go
```diff
@@ -35,6 +35,15 @@ func waitForTitusIsolateWithHost(parentCtx context.Context, taskID, host string,
 			return false
 
 		case <-ticker.C:
+			//give priority to a possible concurrent Done() event non-blocking way
+			// This is the only way to provide "priority" to our done messages
+			// otherwise we cannot control if the outer select picks them over the ticker
+			select {
```
sargun, line 41: You should be able to just check ctx.Err() here instead of doing a full select.

rgulewich (approved):

executor/runtime/docker/docker_isolate.go
solarkennedy: Ah, that is slightly cleaner

PR #547, opened 2021-02-19
Allow Tini to intercept networking system calls

MyUmmaGumma:
Add networking system calls to seccomp filter conditional upon an environment variable.

Intercept/Listen to the first sendmsg syscall via seccomp so that we avoid a catch-22 with the Titus Seccomp Agent while sending across the notifyfd.

Tini logs should read like this
```
Networking system calls will be intercepted
Listening to networking syscall notifications on 6..
Can send the notify fd now!
  fd=6; events: EPOLLIN
Intercepted sendmsg - passthrough!
Stopped polling notifyfd on Tini, over to TSA!
Sent the notify fd to the seccomp agent socket at /titus-executor-sockets/titus-seccomp-agent.sock

### DIFF
--- a/tini/CMakeLists.txt
+++ b/tini/CMakeLists.txt
@@ -64,8 +64,8 @@ target_link_libraries(seccomp-fd-notify cap)
 
 add_executable (tini-static src/tini.c src/seccomp_fd_notify.c)
 target_link_libraries(tini-static seccomp-fd-notify cap -static)
-TARGET_LINK_LIBRARIES (tini  ${CMAKE_THREAD_LIBS_INIT} )
-TARGET_LINK_LIBRARIES (tini-static  ${CMAKE_THREAD_LIBS_INIT} )
+target_link_libraries (tini  ${CMAKE_THREAD_LIBS_INIT} )
+target_link_libraries (tini-static  ${CMAKE_THREAD_LIBS_INIT} )
 set_target_properties (tini-static PROPERTIES LINK_FLAGS "-Wl,--no-export-dynamic -static")
 
 # Installation
--- a/tini/CMakeLists.txt
+++ b/tini/CMakeLists.txt
@@ -55,7 +55,7 @@ configure_file (
 
 
 include_directories ("${PROJECT_BINARY_DIR}")
-
+find_package (Threads REQUIRED)
 add_executable (tini src/tini.c src/seccomp_fd_notify.c)
 target_link_libraries(tini seccomp-fd-notify cap)
 
@@ -64,6 +64,9 @@ target_link_libraries(seccomp-fd-notify cap)
 
 add_executable (tini-static src/tini.c src/seccomp_fd_notify.c)
 target_link_libraries(tini-static seccomp-fd-notify cap -static)
+TARGET_LINK_LIBRARIES (tini  ${CMAKE_THREAD_LIBS_INIT} )
+TARGET_LINK_LIBRARIES (tini-static  ${CMAKE_THREAD_LIBS_INIT} )
+set_target_properties (tini-static PROPERTIES LINK_FLAGS "-Wl,--no-export-dynamic -static")
 
 # Installation
 install (TARGETS tini DESTINATION bin)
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by solarkennedy** [comment]
@MyUmmaGumma (now that we have this as a normal folder in this titus-executor repo) can you look at the failing build?
https://app.circleci.com/pipelines/github/Netflix/titus-executor/632/workflows/2ba37fa0-08da-4cd2-94d5-1480569152d1/jobs/11461

File: tini/src/seccomp_fd_notify.c [hunk 1]
```diff
@@ -84,11 +96,42 @@ static int install_notify_filter(void) {
...
+		BPF_STMT(BPF_RET | BPF_K, SECCOMP_RET_ALLOW),
 	};
 
+	char *is_handle_net = getenv("TITUS_SECCOMP_AGENT_HANDLE_NET_SYSCALLS");
+	if (is_handle_net != NULL) {
+		prog.filter = net_perf_filter;
+		prog.len = 
+			(unsigned short)(sizeof(net_perf_filter) / sizeof(net_perf_filter[0]));
+		PRINT_INFO("Networking system calls will be intercepted");
+	} else {
```
solarkennedy, line 128: can you gate this on `TITUS_SECCOMP_AGENT_HANDLE_PERF_SYSCALLS` ?

solarkennedy:
I understand why we need this change conceptually, but the implementation is a tad over my head.

My #1 concern is that tini remains to be safe, because 100% of titus workloads use it as the pid1.
This should continue to be the case, because we never try to run any of this crazy code if `TITUS_SECCOMP_NOTIFY_SOCK_PATH` is unset, so that is good.

Otherwise.... shipit? :)

solarkennedy (approved):

tini/src/seccomp_fd_notify.c
```diff
@@ -109,30 +116,198 @@ static int install_notify_filter(void)
...
+		prog.len = 
+			(unsigned short)(sizeof(perf_filter) / sizeof(perf_filter[0]));
+		PRINT_INFO("BPF/Perf system calls will be intercepted");
+	} else {
+		PRINT_INFO("No env variables set, no interception of system calls");
+		return -1;
+	}
+
+	/* Setting up permissions to issue seccomp system call */
+	if (prctl(PR_SET_NO_NEW_PRIVS, 1, 0, 0, 0)) {
```
solarkennedy, line 160: lol

Somehow this breaks ssh using this code, but I'll help figure that out.

solarkennedy:
Thank you for rebasing. Let me make sure tini is working it its current state on master (@alekar is helping confirm), once it is green then I'll merge this in, but I don't want to add more to the existing breakage.

tini/src/seccomp_fd_notify.c
solarkennedy: This happens on line :162 already

tini/CMakeLists.txt
```diff
@@ -64,6 +64,9 @@ target_link_libraries(seccomp-fd-notify cap)
 
 add_executable (tini-static src/tini.c src/seccomp_fd_notify.c)
 target_link_libraries(tini-static seccomp-fd-notify cap -static)
+TARGET_LINK_LIBRARIES (tini  ${CMAKE_THREAD_LIBS_INIT} )
```
solarkennedy, line 67: Is it really in caps like this?

solarkennedy:
Just fix the NO_NEW_PRIVS thing and I think this is mergable

MyUmmaGumma:
Fixed the NO_NEW_PRIVS check and the capitals on the Makefile

tini/src/seccomp_fd_notify.c
```diff
@@ -123,9 +130,27 @@ static int install_notify_filter(void)
 		BPF_STMT(BPF_RET | BPF_K, SECCOMP_RET_ALLOW),
 	};
 
-	struct sock_fprog prog = {
-		.len = (unsigned short)(sizeof(filter) / sizeof(filter[0])),
-		.filter = filter,
+struct sock_filter net_perf_filter[] = {
```
solarkennedy, line 133: This variable is called `net_perf`, but it only has `net`.

I think we'll need ti to *also* have perf, unless you want to concat them somehow?

solarkennedy: Also can you add sendmmesg too?

PR #548, opened 2021-02-19
Make titus-mount-block-device use fd instead of pid

solarkennedy:
In general fds are more robust than PIDs, but with
titus, it isn't super easy to know what this PID
of your container actually is.

But we do know a well-known path. From that path,
with some C trickery, we can get all the namespaces
we would have needed otherwise and gotten from a pidfd.

----

This is a tradeoff between where I put the complexity.
The pid1 approach was nice, but it required complexity
in `titus-storage` to even get the pid1. With this change,
`titus-storage` can be simpler (can just pass the env variable
on through) and we make `titus-mount-block-device` pick
apart the namespaces given a directory.

### DIFF
--- a/mount/Makefile
+++ b/mount/Makefile
@@ -1,3 +1,5 @@
+all: titus-mount titus-mount-block-device
+
 titus-mount: mount.c scm_rights.c
 	# musl needs this extra path here
 	# so it can pick up our linux headers for syscalls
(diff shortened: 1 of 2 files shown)


alekar:
My preference is to write something like this:
```c
int get_pid(char *path) {
	char fname[256];
	sprintf(fname, "%s/stat", path);
	printf("%s\n", fname);
	FILE *in = fopen(fname, "r");
	int pid;
	fscanf(in, "%d", &pid);
	printf("%d\n", pid);
}
```

That said, the current approach works well, so :thu:

alekar (approved):

PR #550, opened 2021-02-20
Added the titus-storage service, with EBS support

solarkennedy:
This is the new titus-storage system service.
Unlike previous storage options (nfs), this can
run as a system service, which should make it easier to debug
and iterate on. Eventually I would like for it to handle
other storage needs, perhaps nfs too.

It is pretty conservative, and will refuse to do anything scary.

Probably my biggest fear is around leaving orphaned attached volumes
that need to be manually cleaned up. I don't know how often this will
happen, but I think we'll need a reaper.

### DIFF
--- a/cmd/titus-storage/ebs.go
+++ b/cmd/titus-storage/ebs.go
@@ -1,0 +1,313 @@
+package main
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"regexp"
+	"time"
+
+	"github.com/Netflix/titus-executor/logger"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/credentials/ec2rolecreds"
+	"github.com/aws/aws-sdk-go/aws/ec2metadata"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/mvisonneau/go-ebsnvme/pkg/ebsnvme"
+)
+
+func ebsRunner(ctx context.Context, command string, config MountConfig) error {
+	if config.ebsVolumeID == "" {
+		return errors.New("TITUS_EBS_VOLUME_ID unset, unable to perform any EBS-related operations")
+	}
+	ec2Client := getEc2Client()
+	var err error
+
+	switch command {
+	case "start":
+		err = ebsStart(ctx, ec2Client, config)
+	case "stop":
+		err = ebsStop(ctx, ec2Client, config)
+	default:
+		return fmt.Errorf("Command %q unsupported. Must be either start or stop", command)
+	}
+	if err != nil {
+		return fmt.Errorf("Unable to run command %q: %w", command, err)
+	}
+	return nil
+}
+
+func ebsStart(ctx context.Context, ec2Client *ec2.EC2, c MountConfig) error {
+	var err error
+	l := logger.GetLogger(ctx)
+
+	ec2DeviceName, ok := getDeviceName(c.ebsVolumeID)
+	if ok {
+		l.Printf("EBS volume %s is already attached locally, moving on", c.ebsVolumeID)
+	} else {
+		err = waitForVolumeToBeUnused(ctx, c.ebsVolumeID, ec2Client, 30)
+		if err != nil {
+			return fmt.Errorf("%s is not unused. Not safe to detach. %s", c.ebsVolumeID, err)
+		}
+		ec2DeviceName, err = attachEBS(ctx, c.ebsVolumeID, ec2Client, c.instanceID)
+		if err != nil {
+			return err
+		}
+	}
+
+	device, err := GetActualBlockDeviceName(ec2DeviceName)
+	if err != nil {
+		return err
+	}
+
+	err = mkfsIfNeeded(device, c.ebsFStype)
+	if err != nil {
+		return err
+	}
+
+	err = mountIt(ctx, device, c.ebsFStype, c.ebsMountPoint, c.ebsMountPerm, c.pid1Dir)
+	if err != nil {
+		return err
+	}
+	l.Printf("finished setting up EBS %s at %s inside the container", c.ebsVolumeID, c.ebsMountPoint)
+	return nil
+}
+
+func ebsStop(ctx context.Context, ec2Client *ec2.EC2, c MountConfig) error {
+	l := logger.GetLogger(ctx)
+	ec2Device, ok := getDeviceName(c.ebsVolumeID)
+	if !ok {
+		l.Printf("EBS volume %s doesn't look attached. Nothing to do", c.ebsVolumeID)
+		return nil
+	}
+
+	actualDeviceName, err := GetActualBlockDeviceName(ec2Device)
+	if err != nil {
+		return err
+	}
+
+	// Sometimes ebsStop gets called *before* a container dies, so we need to wait.
+	// Sometimes ebsStop gets called *after* a container dies, that is fine, will move on quickly
+	// *Sometimes* ebsStop gets called *defore* a container dies, but *after* a *different* container
+	// comes up, see it is attached, and mounts. In this case we need to still panic here, but it is "ok"
+	// In that situation we should not continue to detach a volume out from underneath it
+	l.Printf("Waiting up to %d seconds for %s to be not in use...", 30, actualDeviceName)
+	err = waitForDeviceToBeNotInUse(actualDeviceName, 30)
+	if err != nil {
+		return fmt.Errorf("Not detaching volume: %s", err)
+	}
+
+	// By the time we get here, we are confident the volume is attached, not in use, and safe to detatch
+	err = detatchEBS(ctx, c.ebsVolumeID, ec2Client, c.instanceID)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func getEc2Client() *ec2.EC2 {
+	sess := session.Must(session.NewSessionWithOptions(session.Options{
+		SharedConfigState: session.SharedConfigEnable,
+	}))
+	creds := credentials.NewCredentials(
+		&ec2rolecreds.EC2RoleProvider{
+			Client:       ec2metadata.New(sess),
+			ExpiryWindow: 5 * time.Minute,
+		},
+	)
+	sess.Config.Credentials = creds
+	sess.Config.Region = aws.String(os.Getenv("EC2_REGION"))
+	return ec2.New(sess)
+}
+
+func waitForVolumeToBeUnused(ctx context.Context, ebsVolumeID string, ec2Client *ec2.EC2, timeout int) error {
+	l := logger.GetLogger(ctx)
+	for attempt := 0; attempt < timeout; attempt++ {
+		req := ec2.DescribeVolumeStatusInput{
+			VolumeIds: []*string{aws.String(ebsVolumeID)},
+		}
+		response, err := ec2Client.DescribeVolumeStatus(&req)
+		if err != nil {
+			return err
+		}
+		if len(response.VolumeStatuses) > 0 {
+			s := response.VolumeStatuses[0].VolumeStatus.Status
+			if aws.StringValue(s) == "ok" {
+				return nil
+			}
+			l.Printf("Volume Status %+v", *s)
+		}
+		// This is a simple way to sleep longer and longer each time,
+		// to not hammer the aws API
+		time.Sleep(time.Duration(attempt) * time.Second)
+	}
+	return fmt.Errorf("Waited too long for volume to be unused")
+}
+
+func randDriveNamePicker() (string, error) {
+	deviceName := "/dev/xvd"
+	runes := []rune("fghijklmnopqrstuvwxyz")
+	for i := 0; i < len(runes); i++ {
+		drive := deviceName + string(runes[i])
+		if !doesDriveExist(drive) {
+			return drive, nil
+		}
+	}
+	return "", errors.New("Ran out of drive names")
+}
+
+func doesDriveExist(driveName string) bool {
+	if _, err := os.Stat(driveName); os.IsNotExist(err) {
+		for _, file := range listNvmeBlockDevices() {
+			if d, _ := ebsnvme.ScanDevice(file); d.Name == driveName {
+				return true
+			}
+		}
+		return false
+	}
+	return true
+}
+
+// waitForDeviceToDisappear is an inexpensive way to poll for an EBS
+// detachment without calling the EC2 API. It does this by waiting
+// for the block device it was attached with to dissapear from /dev
+func waitForDeviceToDisappear(ctx context.Context, driveName string, maxAttempts int) error {
+	start := time.Now()
+	l := logger.GetLogger(ctx)
+	var attempts int
+	for doesDriveExist(driveName) {
+		time.Sleep(2 * time.Second)
+		attempts++
+		if attempts >= maxAttempts {
+			elapsed := int(time.Since(start).Seconds())
+			return fmt.Errorf("waited the max %d seconds for %s to go, but it is still here", elapsed, driveName)
+		}
+	}
+	elapsed := int(time.Since(start).Seconds())
+	l.Printf("waited for %s to disappear, and it is now gone after %d seconds", driveName, elapsed)
+	return nil
+}
+
+// waitForDriveToExistWithTimeout is an inexpensive way to poll for an EBS
+// attachment without calling the EC2 API. It does this by scaning the
+// list of nvme volumes on disk and looking at their 'serial number'
+func waitForDriveToExistWithTimeout(driveName string, maxAttempts int) error {
+	var attempts int
+	for !doesDriveExist(driveName) {
+		time.Sleep(2 * time.Second)
+		attempts++
+		if attempts >= maxAttempts {
+			return fmt.Errorf("drive %s still does't exist after waiting %d seconds", driveName, attempts*2)
+		}
+	}
+	return nil
+}
+
+// GetActualBlockDeviceName returns the actual name of the block device seen
+// within the instance, like /dev/nvme1
+// This is often different from the ec2 device name, which is usually
+// something like /dev/sdx
+func GetActualBlockDeviceName(name string) (string, error) {
+	for _, device := range listNvmeBlockDevices() {
+		if d, _ := ebsnvme.ScanDevice(device); d.Name == name {
+			return device, nil
+		}
+	}
+	if _, err := os.Stat(name); os.IsNotExist(err) {
+		return "", err
+	}
+	return name, nil
+}
+
+func listNvmeBlockDevices() (devices []string) {
+	re := regexp.MustCompile(`(^\/dev\/nvme[0-9]+n1$)`)
+	f, _ := filepath.Glob("/dev/nvme*")
+	for _, d := range f {
+		if re.Match([]byte(d)) {
+			devices = append(devices, d)
+		}
+	}
+	return
+}
+
+// getDeviceName scans the Nvme on disk and looks at the
+// serial numbers of the devices and returns the device name
+// (/dev/nmve1). Behaves like a hash get, returning key, ok.
+// returning "", false if unfound.
+func getDeviceName(ebsVolumeID string) (string, bool) {
+	for _, device := range listNvmeBlockDevices() {
+		if d, _ := ebsnvme.ScanDevice(device); d.VolumeID == ebsVolumeID {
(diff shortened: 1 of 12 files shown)


cmd/titus-storage/ebs.go
```diff
@@ -0,0 +1,272 @@
...
+			if d, _ := ebsnvme.ScanDevice(file); d.Name == driveName {
+				return true
+			}
+		}
+		return false
+	}
+	return true
+}
+
+func waitForDeviceToDisappear(driveName string, maxAttempts int) bool {
```
alekar, line 155: Consider returning error?

cmd/titus-storage/ebs.go
```diff
@@ -0,0 +1,272 @@
...
+		if attempts >= maxAttempts {
+			log.Printf("Waited the max %d seconds for %s to go, but it is still here!", attempts*2, driveName)
+			return false
+		}
+	}
+	log.Printf("Waited for %s to disappear, and it is now gone after %d seconds", driveName, attempts*2)
+	return true
+}
+
+func doesDriveExistWithTimeout(driveName string, maxAttempts int) bool {
```
alekar, line 169: Consider returning error?

cmd/titus-storage/main.go
```diff
@@ -0,0 +1,22 @@
+package main
+
+import (
+	"log"
+	"os"
+)
+
+func main() {
+	command := os.Args[1]
```
alekar, line 9: You should use `flags.Parse` or cobra.

cmd/titus-storage/main.go
```diff
@@ -0,0 +1,22 @@
+package main
+
+import (
+	"log"
```
alekar, line 4: Please use journalhook, consider using structured logging, aka zerolog?

cmd/titus-storage/ebs.go
```diff
@@ -0,0 +1,272 @@
...
+		log.Printf("volume %s doesn't seem to connected. Not bothering to detach using the AWS API", ebsVolumeID)
+	}
+	detachVolIn := &ec2.DetachVolumeInput{
+		Device:     &deviceName,
+		InstanceId: &instanceID,
+		VolumeId:   &ebsVolumeID,
+	}
+	volAttachments, err := ec2Client.DetachVolume(detachVolIn)
+	if err != nil {
+		log.Fatal(err)
```
alekar, line 261: Maybe return error here, there might be cleanup needed in caller?

cmd/titus-storage/ebs.go
```diff
@@ -0,0 +1,272 @@
...
+		log.Fatal("Couldn't find an unused drive name")
+	}
+	attachVolIn := &ec2.AttachVolumeInput{
+		Device:     &deviceName,
+		InstanceId: &instanceID,
+		VolumeId:   &ebsVolumeID,
+	}
+	volAttachments, err := ec2Client.AttachVolume(attachVolIn)
+	if err != nil {
+		log.Fatalf("Couldn't attach: %v", err)
```
alekar, line 239: Maybe return error here, there could be cleanup required in the caller?

cmd/titus-storage/ebs.go
```diff
@@ -0,0 +1,272 @@
...
+		}
+	}
+	return
+}
+
+// getDeviceName scans the Nvme on disk and looks at the
+// serial numbers of the devices and returns the device name
+// (/dev/nmve1). Behaves like a hash get, returning key, ok.
+// returning "", false if unfound.
+func getDeviceName(ebsVolumeID string) (string, bool) {
```
alekar, line 212: Consider using `(string, error)` as the return type.

cmd/titus-storage/ebs.go
```diff
@@ -0,0 +1,272 @@
...
+			ExpiryWindow: 5 * time.Minute,
+		},
+	)
+	sess.Config.Credentials = creds
+	*sess.Config.Region = os.Getenv("EC2_REGION")
+	return ec2.New(sess)
+}
+
+func waitForVolumeToBeUnused(ebsVolumeID string, ec2Client *ec2.EC2, timeout int) bool {
+	vols := []*string{&ebsVolumeID}
```
alekar, line 111: AWS SDK provides a helper for this `aws.String`, I think?

cmd/titus-storage/main.go
solarkennedy: Ok. Just seemed like a lot of complexity for 1 cli argument.
I don't want to use flags because it doesn't support bare args.

cmd/titus-storage/ebs.go
```diff
@@ -0,0 +1,272 @@
...
+	return deviceName, deviceName != ""
+}
+
+func attachEBS(ebsVolumeID string, ec2Client *ec2.EC2, instanceID string) (string, error) {
+	deviceName, ok := getDeviceName(ebsVolumeID)
+	if ok {
+		log.Printf("EBS Volume %s already attached as %s", ebsVolumeID, deviceName)
+		return deviceName, nil
+	}
+	deviceName, err := randDriveNamePicker()
```
rgulewich, line 228: Is this racey?  If you have 2 different attaches happening at the same time, and they're going through the list in the same order (which it looks like they are), then it's possible that they both pick the same drive name, and the slower one fails, right?  This gets much more likely if the attach takes many seconds.  I'm wondering if this function should retry on an error from trying to attach a drive name that already exists.

(I know this is a Draft PR, so it may well be that you're planning on fixing this later, but figured I'd ask.)

cmd/titus-storage/ebs.go
rgulewich: I'm wondering (and this is an open question to you all) how we feel about using `log.Fatal()` all over the place?  It seems like if we ever want to write unit tests for this code (and maybe we don't), that returning errors and then having `main()` do the `log.Fatal()` might make that easier.

cmd/titus-storage/main.go
rgulewich: I'm wondering if you might want to, since there are a bunch of `os.GetEnv()`s in various places in this code that you could consolidate into the cobra parsing.  You could then log the total config for the CLI tool (based on flags and Env vars) in one place, which would be nice for debugging.

cmd/titus-storage/main.go
solarkennedy: will do. This will be especially handy for tracing the lifetime of a particular volume id

cmd/titus-storage/ebs.go
solarkennedy: I agree. I'm going to consolidate all the fatals into returns and let the main runner handle the actual exits

cmd/titus-storage/ebs.go
solarkennedy: Hmm. In my bash version I `flockd` everything so that there was never a chance for races on the single host.

But I think it would be simpler to just fail and let titus relaunch it with its own backoff.

alekar: We do have `FSLocker` in titus-executor, if you want to use it.

cmd/titus-storage/mount.go
```diff
@@ -0,0 +1,79 @@
...
+
+func mountIt(dev string, fstype string, mountPoint string, mountPerm string) error {
+	flags := calculateFlags(mountPerm)
+	_ = os.Chdir("/tmp")
+	pid1 := os.Getenv("TITUS_PID_1_DIR")
+	if pid1 == "" {
+		log.Fatalf("TITUS_PID_1_DIR is not set!")
+	}
+	log.Printf("Running %s to mount %s onto %s in the container", mountCommand, dev, mountPoint)
+	log.Printf("TITUS_PID_1_DIR=%s MOUNT_TARGET=%s MOUNT_OPTIONS=source=%s MOUNT_FLAGS=%s MOUNT_FSTYPE=%s %s", pid1, mountPoint, dev, flags, fstype, mountCommand)
```
alekar, line 33: This log line might get out of sync, consider logging `cmd.Env`.

cmd/titus-storage/main.go
solarkennedy: That is true, cobra + viper does have some magic to turn variables into cli constructs.

I'll look to see how hard it is to put these together to remove the os.getEnvs()

cmd/titus-storage/mount.go
solarkennedy: great idea

cmd/titus-storage/ebs.go
solarkennedy: No cleanup required.
Kinda the beauty of this design is that the `stop` function can clean up anything, if required.

alekar (approved):

cmd/titus-storage/ebs.go
solarkennedy: Hmm, this is all pretty silly, because this device name (/dev/xvda) isn't actually used in practice with nitro/nvme.

solarkennedy:
Added structured logging, more robust cli parsing, and real error handling. Ready for second round of review if applicable.

cmd/titus-storage/ebs.go
```diff
@@ -0,0 +1,312 @@
...
+	"github.com/aws/aws-sdk-go/aws/credentials/ec2rolecreds"
+	"github.com/aws/aws-sdk-go/aws/ec2metadata"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/mvisonneau/go-ebsnvme/pkg/ebsnvme"
+)
+
+func ebsRunner(ctx context.Context, command string, config MountConfig) error {
+	if config.ebsVolumeID == "" {
+		return fmt.Errorf("TITUS_EBS_VOLUME_ID unset, unable to perform any EBS-related operations")
```
sargun, line 23: This could be an errors.New

cmd/titus-storage/ebs.go
```diff
@@ -0,0 +1,312 @@
...
+}
+
+func getEc2Client() *ec2.EC2 {
+	sess := session.Must(session.NewSessionWithOptions(session.Options{
+		SharedConfigState: session.SharedConfigEnable,
+	}))
+	creds := credentials.NewCredentials(
+		&ec2rolecreds.EC2RoleProvider{
+			Client:       ec2metadata.New(sess),
+			ExpiryWindow: 5 * time.Minute,
```
sargun, line 117: Why is this setup like so?

cmd/titus-storage/ebs.go
```diff
@@ -0,0 +1,312 @@
...
+		},
+	)
+	sess.Config.Credentials = creds
+	*sess.Config.Region = os.Getenv("EC2_REGION")
```
sargun, line 121: I think you can do:
sess.Config.Region = aws.string(os.Getenv("EC2_REGION"))

cmd/titus-storage/ebs.go
```diff
@@ -0,0 +1,312 @@
...
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/mvisonneau/go-ebsnvme/pkg/ebsnvme"
+)
+
+func ebsRunner(ctx context.Context, command string, config MountConfig) error {
+	if config.ebsVolumeID == "" {
+		return fmt.Errorf("TITUS_EBS_VOLUME_ID unset, unable to perform any EBS-related operations")
+	}
+	ec2Client := getEc2Client()
+	if command == "start" {
```
sargun, line 26: A switch statement like so seems like it's work:

```go
var err error

switch (command) {
case "start":
err = ebsStart(ctx, ec2Client, config)
case "stop":
err = ebsStop(ctx, ec2Client, config)
default:
return fmt.Errorf("Command %q unsupported. Must be either start or stop", command)
}

if err != nil {
   return fmt.Errorf("Unable to run command %q: %w", comand, err)
}
return nil
}
```

cmd/titus-storage/ebs.go
```diff
@@ -0,0 +1,312 @@
...
+
+// getDeviceName scans the Nvme on disk and looks at the
+// serial numbers of the devices and returns the device name
+// (/dev/nmve1). Behaves like a hash get, returning key, ok.
+// returning "", false if unfound.
+func getDeviceName(ebsVolumeID string) (string, bool) {
+	deviceName := ""
+	for _, device := range listNvmeBlockDevices() {
+		if d, _ := ebsnvme.ScanDevice(device); d.VolumeID == ebsVolumeID {
+			deviceName = d.Name
```
sargun, line 242: Cant you just return from here? Or can multiple devices have the right volume ID? If so, which one should it be?

cmd/titus-storage/ebs.go
```diff
@@ -0,0 +1,312 @@
...
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/credentials/ec2rolecreds"
+	"github.com/aws/aws-sdk-go/aws/ec2metadata"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/mvisonneau/go-ebsnvme/pkg/ebsnvme"
+)
+
+func ebsRunner(ctx context.Context, command string, config MountConfig) error {
+	if config.ebsVolumeID == "" {
```
sargun, line 22: Seems like we could do getDeviceName up here.

cmd/titus-storage/ebs.go
```diff
@@ -0,0 +1,312 @@
...
+			ExpiryWindow: 5 * time.Minute,
+		},
+	)
+	sess.Config.Credentials = creds
+	*sess.Config.Region = os.Getenv("EC2_REGION")
+	return ec2.New(sess)
+}
+
+func waitForVolumeToBeUnused(ctx context.Context, ebsVolumeID string, ec2Client *ec2.EC2, timeout int) error {
+	l := logger.GetLogger(ctx)
```
sargun, line 126: In general, the "easier" pattern here is:

```go
logger.G(ctx).Info(...)
```

As opposed to getting the logger and using it by hand

cmd/titus-storage/ebs.go
```diff
@@ -0,0 +1,312 @@
...
+	var err error
+	l := logger.GetLogger(ctx)
+
+	ec2DeviceName, ok := getDeviceName(c.ebsVolumeID)
+	if ok {
+		l.Printf("EBS volume %s is already attached locally, moving on", c.ebsVolumeID)
+	} else {
+		err = waitForVolumeToBeUnused(ctx, c.ebsVolumeID, ec2Client, 30)
+		if err != nil {
+			return fmt.Errorf("%s is not unused. Not safe to detach. %s", c.ebsVolumeID, err)
```
sargun, line 52: Errors should be "formatted" with %w.

cmd/titus-storage/ebs.go
```diff
@@ -0,0 +1,312 @@
...
+		}
+	}
+	return nil
+}
+
+// GetActualBlockDeviceName returns the actual name of the block device seen
+// within the instance, like /dev/nvme1
+// This is often different from the ec2 device name, which is usually
+// something like /dev/sdx
+func GetActualBlockDeviceName(name string) (string, error) {
```
sargun, line 211: Does this need to be "public"?

cmd/titus-storage/ebs.go
```diff
@@ -0,0 +1,312 @@
...
+		Device:     &ec2DeviceName,
+		InstanceId: &instanceID,
+		VolumeId:   &ebsVolumeID,
+	}
+	volAttachments, err := ec2Client.AttachVolume(attachVolIn)
+	if err != nil {
+		return "", err
+	}
+	l.Printf("volAttachments status after AttachVolume: %+v", volAttachments)
+	l.Printf("Now waiting for device to exist as ec2Device name %s", ec2DeviceName)
```
sargun, line 269: WithField is a good trick here maybe

cmd/titus-storage/ebs.go
```diff
@@ -0,0 +1,312 @@
...
+// detachment without calling the EC2 API. It does this by waiting
+// for the block device it was attached with to dissapear from /dev
+func waitForDeviceToDisappear(ctx context.Context, driveName string, maxAttempts int) error {
+	l := logger.GetLogger(ctx)
+	var attempts int
+	for doesDriveExist(driveName) {
+		time.Sleep(2 * time.Second)
+		attempts++
+		if attempts >= maxAttempts {
+			return fmt.Errorf("waited the max %d seconds for %s to go, but it is still here", attempts*2, driveName)
```
sargun, line 185: 1. Use WithField
2. Record the start as time.Now(), and use time.since

cmd/titus-storage/ebs.go
```diff
@@ -0,0 +1,312 @@
...
+		if len(response.VolumeStatuses) > 0 {
+			s := response.VolumeStatuses[0].VolumeStatus.Status
+			if *s == "ok" {
+				return nil
+			}
+			l.Printf("Volume Status %+v", *s)
+		}
+		time.Sleep(time.Duration(attempt) * time.Second)
+	}
+	return fmt.Errorf("Waited too long for volume to be unused")
```
sargun, line 144: errors.New? Or add the volume ID to the error?

cmd/titus-storage/ebs.go
```diff
@@ -0,0 +1,312 @@
...
+			return err
+		}
+		if len(response.VolumeStatuses) > 0 {
+			s := response.VolumeStatuses[0].VolumeStatus.Status
+			if *s == "ok" {
+				return nil
+			}
+			l.Printf("Volume Status %+v", *s)
+		}
+		time.Sleep(time.Duration(attempt) * time.Second)
```
sargun, line 142: This is a little confusing in that it will sleep:
1 + 2 + 3 + 4 + 5 + ... seconds.

Maybe a comment?

cmd/titus-storage/ebs.go
```diff
@@ -0,0 +1,312 @@
...
+		req := ec2.DescribeVolumeStatusInput{
+			VolumeIds: []*string{aws.String(ebsVolumeID)},
+		}
+		response, err := ec2Client.DescribeVolumeStatus(&req)
+		if err != nil {
+			return err
+		}
+		if len(response.VolumeStatuses) > 0 {
+			s := response.VolumeStatuses[0].VolumeStatus.Status
+			if *s == "ok" {
```
sargun, line 137: aws.stringvalue(s)

cmd/titus-storage/ebs.go
```diff
@@ -0,0 +1,312 @@
...
+		time.Sleep(time.Duration(attempt) * time.Second)
+	}
+	return fmt.Errorf("Waited too long for volume to be unused")
+}
+
+func randDriveNamePicker() (string, error) {
+	ctr := 0
+	deviceName := "/dev/xvd"
+	runes := []rune("fghijklmnopqrstuvwxyz")
+	for {
```
sargun, line 151:
```go
for i := 0; i < len(runes); i ++ {
  drive := deviceName + string(runes[ctr])
  if !doesDriveExist(drive) {
    return drive, nil
  }
}

return "", errrors.New("Ran out of drive names")
}

cmd/titus-storage/ebs.go
solarkennedy: yea it was cheap "exponential" backoff

cmd/titus-storage/ebs.go
solarkennedy: I have to re-learn what is best practice here every time for some reason.

I thought fmt.errorf (+ and optional %w to wrap another error) was considered best practice?
https://blog.golang.org/go1.13-errors

Would you like me to replace all the fmt.errorfs with errors.New? Or just this one instance?

cmd/titus-storage/ebs.go
solarkennedy: no particular reason, is there are style of ec2 cred session making I could emulate, maybe the one from titus-vpc-service?

cmd/titus-storage/ebs.go
solarkennedy: oh yea of course. No I don't think we could ever have 2 devices with the same volume id

cmd/titus-storage/ebs.go
solarkennedy: 1. which field would you like added, the device name?
`l` should already have the task id and volume id

2. will do

cmd/titus-storage/ebs.go
solarkennedy: I agree, more normal looking

solarkennedy:
Sorry for the git noise. I've rebased on master to make it mergeable again and tidied up the commits.

@sargun can you look at this one more time and clarify the errors.new and session issue?

solarkennedy:
@sargun I'm going to go ahead and merge to get a build I can actually play with, but please feel free to review any parts you think need improvement and I'll incorporate them.

PR #551, opened 2021-02-24
Prevent invalid environment variables from slipping into profile.sh

solarkennedy:
We don't strictly enforce the env variable strings at our
outer API layers. I think that is "kinda" ok. Maybe
some day we'll be stricter.

But deep down where we are actually deploying a shell-compatible
file, then we can not put in shell-incompatible variables.

But, for discovery, I put the invalid ones at the bottom and comment
them out.


executor/runtime/docker/docker_env.go
```diff
@@ -23,6 +24,13 @@ export {{ $key }}={{ $val | escape_sq }}
 export {{ $key }}={{ escapeSQWithFallback $key $val }}
 {{ end -}}
 {{ end -}}
+
+{{ if .InvalidEnv }}
+# Any extra invalid env variables will be commented out here:
+{{ range $key, $val := .InvalidEnv -}}
+# export {{ $key }}={{ $val | escape_sq }}
```
alekar, line 31: Left over?

alekar (approved):
Sweet!

executor/runtime/docker/docker_env.go
solarkennedy: What do you mean?
It is left over in the sense that I didn't want to silently discard them?

alekar:
I meant the line `# export {{ $key }}={{ $val | escape_sq }}`.

executor/runtime/docker/docker_env.go
alekar: I meant this line `# export {{ $key }}={{ $val | escape_sq }}`.

solarkennedy: Yea, line 31? That is commented out in the template, so that when someone goes to look at the profile.sh file they will see "why" those env variables are not set (they are commented out)

alekar: Ah, I was misreading. Cool.

tbak (approved):

solarkennedy:
Confirmed from our partner teams that they don't use env variables that have `.`s from the env directly, instead spaas base64 json encodes the shell-invalid variables into a *different* env variable (that bash can understand), and their python loader script parses does a `os.env().append(json.decodes(base64.decode(os.env["MY_MEGA_VAR"])))` before exec'ing!!!

PR #552, opened 2021-02-24
Add metric reporter to test watcher.

alekar:
Add discard reporter to test watcher.


solarkennedy:
```filesystems/watcher_test.go:5: File is not `goimports`-ed (goimports)
	"github.com/Netflix/metrics-client-go/metrics"```

PR #553, opened 2021-02-24
Add ability to inject atlas agent as a system service.

alekar:
Add ability to inject atlas agent as system service. When the agent is injected as a system service, don't inject the agent from the host.


root/lib/systemd/system/titus-atlasd@.service
```diff
@@ -0,0 +1,16 @@
+[Unit]
+Description=Spectatord for container %i
```
solarkennedy, line 2: copy/paste

executor/runtime/docker/docker_linux.go
```diff
@@ -67,9 +67,10 @@ var systemServices = []serviceOpts{
 		required:     false,
 	},
 	{
-		humanName: "atlas",
-		unitName:  "atlas-titus-agent",
-		required:  false,
+		humanName:    "atlas",
+		unitName:     "atlas-titus-agent",
```
solarkennedy, line 71: you called it `titus-altasd` down there

executor/runtime/docker/docker_config.go
```diff
@@ -166,6 +166,22 @@ func shouldStartSpectatord(cfg *config.Config, c runtimeTypes.Container) bool {
...
+	}
+
+	if cfg.AtlasdServiceImage == "" {
+		return false
+	}
+	return true
+}
+
+func shouldStartAtlasAgent(cfg *config.Config, c runtimeTypes.Container) bool {
+	return !shouldStartAtlasd(cfg, c)
```
solarkennedy, line 182: Is this because Atlast-agent and atlasd are incompatible with each other? (one in the container, one on the host?)
if so can you add a comment here to say that?

executor/runtime/docker/docker_config.go
```diff
@@ -166,6 +166,22 @@ func shouldStartSpectatord(cfg *config.Config, c runtimeTypes.Container) bool {
 	return true
 }
 
+func shouldStartAtlasd(cfg *config.Config, c runtimeTypes.Container) bool {
+	enabled := cfg.ContainerAtlasd
+	if !enabled {
```
solarkennedy, line 171: What do you think about

```
	if cfg.AtlasdServiceImage == "" {
		return false
	}
        return cfg.ContainerAtlasd
```

config/config.go
```diff
@@ -215,6 +219,16 @@ func NewConfig() (*Config, []cli.Flag) {
 			EnvVar:      "SPECTATORD_SERVICE_IMAGE",
 			Destination: &cfg.SpectatordServiceImage,
 		},
+		cli.BoolFlag{
+			Name:        "container-atlasd",
+			EnvVar:      "CONTAINER_ATLASD",
+			Destination: &cfg.ContainerAtlasd,
+		},
+		cli.StringFlag{
```
solarkennedy, line 227: Do we need both?

Why can't it just be the "image" flag, and if `""` we don't run it?

executor/runtime/types/sidecars.go
```diff
@@ -46,4 +47,10 @@ var sideCars = []SidecarContainerConfig{
 			"/titus/spectatord": {},
 		},
 	},
+	{
+		ServiceName: SidecarServiceSpectatord,
```
solarkennedy, line 51: copy/paste

executor/runtime/docker/docker_linux.go
alekar: I missed adding the new unit here.

executor/runtime/types/sidecars.go
alekar: ty

root/lib/systemd/system/titus-atlasd@.service
alekar: ty

config/config.go
alekar: @solarkennedy and I decided to take this up in a separate pull request.

executor/runtime/docker/docker_config.go
alekar: @solarkennedy and I decided to clean this up in a different pull request.

root/lib/systemd/system/titus-atlasd.service
```diff
@@ -0,0 +1,16 @@
...
+ConditionPathIsDirectory=/var/lib/titus-inits/%i/ns
+
+# If the service restarts more than 10 times in 30 seconds, let it die
+StartLimitIntervalSec=30
+StartLimitBurst=10
+
+[Service]
+EnvironmentFile=/var/lib/titus-environments/%i.env
+# Run as root (UID 0, GID 0) and with CAP_DAC_OVERRIDE so that containers with a `USER` instruction work
+ExecStart=/usr/bin/runc --root /var/run/docker/runtime-${TITUS_OCI_RUNTIME}/moby exec --user 0:0 --cap CAP_DAC_OVERRIDE ${TITUS_CONTAINER_ID} /titus/atlas-titus-agent/start-atlas-titus-agent
```
solarkennedy, line 12: Is the command line the same for atlasd?

executor/runtime/types/sidecars.go
```diff
@@ -46,4 +47,10 @@ var sideCars = []SidecarContainerConfig{
 			"/titus/spectatord": {},
 		},
 	},
+	{
+		ServiceName: SidecarServiceAtlasd,
+		Volumes: map[string]struct{}{
+			"/titus/atlas-titus-agent": {},
```
solarkennedy, line 53: for consistency can this be as simple as `/titus/atlasd` ?

root/lib/systemd/system/titus-atlasd.service
solarkennedy: It would be cool if we could start a convention of just

`/titus/atlastd/start`

`/titus/$sidecar/start`

executor/runtime/docker/docker_linux.go
```diff
@@ -67,9 +67,16 @@ var systemServices = []serviceOpts{
...
-		humanName: "atlas",
-		unitName:  "atlas-titus-agent",
-		required:  false,
+		humanName:    "atlas",
+		unitName:     "titus-atlasd",
+		enabledCheck: shouldStartAtlasd,
+		required:     false,
+	},
+	{
+		humanName:    "atlas",
```
solarkennedy, line 76: I've thought about removing the humanName, I don't think we need it and can just use the unitname. Would you like it to be distinct from the other atlasd so you can tell easily from the logs which one it is talking about?

solarkennedy (approved):

executor/runtime/docker/docker_linux.go
alekar: I intend to wrap this up, then clean up all the sidecars as per the RFC. Can roll removing humanName into that pull request.

executor/runtime/types/sidecars.go
alekar: Depends on the image. Current image is published with that path.

root/lib/systemd/system/titus-atlasd.service
alekar: Can you elaborate, I don't understand.

solarkennedy: If some day we move to dynamic sidecars instead of encoding them here, having a convention for what the command is will help a bit.

For now in this review, I just wanted to confirm that `/titus/atlas-titus-agent/start-atlas-titus-agent` was the actual path (and not something with `atlasd` in it)

alekar: Ah. It's the actual path.

executor/runtime/types/sidecars.go
copperlight: The new image will be published with `/titus/atlas-titus-agent` as the path.

As mentioned elsewhere, we have a separate service that performs a different function, which is already named `atlasd`.

copperlight (commented):
Looks reasonable to me.

We noted that `atlasd` is another service we own which has a function separate and distinct from the `atlas-titus-agent` service. I see the changes to allow for the `atlas-titus-agent` name in a few places and I assume that the rest of the references to `atlasd` are embedded in enough other places in the code, which are not user-exposed, that they are not worth modifying.

PR #554, opened 2021-03-03
Set network namespace

aloktiagi:
This change creates an ipvlan interface in the transition namespace when network system calls need to be intercepted in a container

1. check if network system calls have to be intercepted.
2. create an ipvlan interface in the transition namespace.
3. Assign IPv6 address to the interface in the container's namespace and IPv4 address to the interface in the transition namespace.

### DIFF
--- a/cmd/titus-vpc-tool/setup_container.go
+++ b/cmd/titus-vpc-tool/setup_container.go
@@ -13,20 +13,20 @@ import (
 )
 
 func setupContainercommand(ctx context.Context, v *pkgviper.Viper, iipGetter instanceIdentityProviderGetter) *cobra.Command {
-	withTransition := false
-
-	netNS := []interface{}{v.GetInt("netns")}
-	if v.GetString("trans-netns") != "" {
-		withTransition = true
-		netNS = append(netNS, v.GetString("trans-netns"))
-	}
 	cmd := &cobra.Command{
 		Use:   "setup-container",
 		Short: "Setup networking for a particular container",
 		RunE: func(cmd *cobra.Command, args []string) error {
+			withTransition := false
 			bandwidth := v.GetInt64("bandwidth")
 			burst := v.GetBool("burst")
 			jumbo := v.GetBool("jumbo")
+
+			netNS := []interface{}{v.GetInt("netns")}
+			if v.GetString("transition-netns") != "" {
+				withTransition = true
+				netNS = append(netNS, v.GetString("transition-netns"))
+			}
 			switch strings.ToLower(v.GetString(generationFlagName)) {
 			case "v1":
 				return container.SetupContainer(ctx, iipGetter(), netNS, withTransition, uint64(bandwidth), burst, jumbo)
@@ -39,7 +39,7 @@ func setupContainercommand(ctx context.Context, v *pkgviper.Viper, iipGetter ins
 	}
 
 	cmd.Flags().Int("netns", 3, "The File Descriptor # of the network namespace to setup")
-	cmd.Flags().String("transition-netns", "trans-netns", "The name # of the network namespace to setup for transition")
+	cmd.Flags().String("transition-netns", "", "The name # of the network namespace to setup for transition")
 	cmd.Flags().Int64("bandwidth", 128*1024*1024, "Bandwidth to allocate to the device, in bps")
 	cmd.Flags().Bool("burst", false, "Allow this container to burst its network allocation")
 	cmd.Flags().Bool("jumbo", false, "Allow this container to use jumbo frames")
--- a/cmd/titus-vpc-tool/setup_container.go
+++ b/cmd/titus-vpc-tool/setup_container.go
@@ -12,30 +12,32 @@ import (
 	pkgviper "github.com/spf13/viper"
 )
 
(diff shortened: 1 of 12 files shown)


vpc/tool/container/setup_container_linux.go
```diff
@@ -32,51 +32,81 @@ var (
 	errAddressSetupTimeout = errors.New("IPv6 address setup timed out")
 )
 
-func doSetupContainer(ctx context.Context, netnsfd int, bandwidth, ceil uint64, jumbo bool, allocation types.LegacyAllocation) (netlink.Link, error) {
-	parentLink, err := getLinkByMac(allocation.MAC)
-	if err != nil {
-		return nil, err
+func getNsFd(netNS interface{}) (int, bool, error) {
```
solarkennedy, line 35: This is not very go-like.

Are you sure it isn't simpler to just pass 2 nsfds all the time instead of overloading the `netNS` object?

solarkennedy (changes requested):

cmd/titus-vpc-tool/setup_container.go
```diff
@@ -13,26 +13,33 @@ import (
...
-				return container2.SetupContainer(ctx, iipGetter(), netns, uint64(bandwidth), burst)
+				return container2.SetupContainer(ctx, iipGetter(), netNS, withTransition, uint64(bandwidth), burst)
 			default:
 				return fmt.Errorf("Version %q not recognized", v.GetString(generationFlagName))
 			}
 		},
 	}
 
 	cmd.Flags().Int("netns", 3, "The File Descriptor # of the network namespace to setup")
+	cmd.Flags().String("transition-netns", "trans-netns", "The name # of the network namespace to setup for transition")
```
solarkennedy, line 42: I think the fact that we have an fd above (`3`) and a second thing which is a string which is the name of a namespace, we are setting ourselves up for confusion.

Do we even need it to be a name? Can it be a bool ?

executor/runtime/docker/docker.go
```diff
@@ -978,9 +978,10 @@ func (r *DockerRuntime) Prepare(parentCtx context.Context) error { // nolint: go
 				"TITUS_SECCOMP_AGENT_HANDLE_PERF_SYSCALLS": "true",
 			})
 		}
-		if r.c.SeccompAgentEnabledForPerfSyscalls() {
+		if r.c.SeccompAgentEnabledForNetSyscalls() {
 			r.c.SetEnvs(map[string]string{
 				"TITUS_SECCOMP_AGENT_HANDLE_NET_SYSCALLS": "true",
+				"TITUS_SECCOMP_AGENT_TRANSITION_NETNS":    r.c.TaskID() + "trans-netns",
```
solarkennedy, line 984: Do we need this? 
I guess we need to *somehow* tell TSA what to switch into.
Hmm.

vpc/tool/container/setup_container_linux.go
```diff
@@ -144,9 +174,11 @@ func configureLink(ctx context.Context, nsHandle *netlink.Handle, link netlink.L
 		}
 	}
 
-	err = addIPv4AddressAndRoute(ctx, nsHandle, link, allocation.IPV4Address)
-	if err != nil {
-		return err
+	if !withTrans || (withTrans && isTransLink) {
```
solarkennedy, line 177: This is hard for me to understand.

Could this logic be simplified if these toggles were called `enableIPv4` and `enableIPv6` and let the caller figure out exactly what to pass through, given the transition logic?

vpc/tool/container/setup_container_linux.go
```diff
@@ -292,31 +324,38 @@ func ipaddressToHandle(ip net.IP) uint16 {
...
-	ifbEgress, err := netlink.LinkByName(vpc.EgressIFB)
-	if err == nil {
-		removeClass(ctx, ip, ifbEgress)
-	} else {
-		logger.G(ctx).WithError(err).Warning("Unable to find ifb egress, during deallocation")
-	}
+func teardownNetwork(ctx context.Context, allocation types.LegacyAllocation, links []netlink.Link, netnsfd []interface{}) {
+	for i, link := range links {
+		deleteLink(ctx, link, netnsfd[i])
+		ip := net.ParseIP(allocation.IPV4Address.Address.Address)
```
solarkennedy, line 330: If one of the links doesn't have ipv4, does this still work?

vpc/tool/container/setup_container_linux.go
```diff
@@ -292,31 +324,38 @@ func ipaddressToHandle(ip net.IP) uint16 {
...
 	}
 }
 
-func deleteLink(ctx context.Context, link netlink.Link, netnsfd int) {
+func deleteLink(ctx context.Context, link netlink.Link, netNS interface{}) {
 	if link == nil {
 		logger.G(ctx).Debug("Link not setup, not deleting link")
 		return
 	}
+	netnsfd, _, err := getNsFd(netNS)
```
solarkennedy, line 354: Can we get away with this `deleteLink` function remaining the same as before, just the caller gives it a different fd depending on which link it is deleting?

solarkennedy (changes requested):

solarkennedy:
@aloktiagi can you close this if it is no longer relevant ?

aloktiagi:
Closing.

PR #555, opened 2021-03-05
Enable titus-seccomp-agent for sshd

solarkennedy:
titus-seccomp-agent is used to intercept syscalls
for the main process of a container.

It would be confusing if, when you ssh'd to that container
those same syscalls didn't work!

This change makes it so sshd-spawned processes behave the same.

Maybe in the future we'll have other nsentered processes
also utilize TSA.

This makes the seccomp notification slighty more robust, instead
of needing full sidecar ordering.

### DIFF
--- a/hack/builder/titus-executor-builder.sh
+++ b/hack/builder/titus-executor-builder.sh
@@ -17,7 +17,7 @@ mv mount/titus-mount-block-device build/bin/linux-amd64/
 
 # tini
 (
-    mkdir -p build/tini
+    rm -rf build/tini && mkdir -p build/tini
     cd build/tini
     # TODO(sargun): RELWITHDEBINFO
     cmake -DCMAKE_BUILD_TYPE=Release ../../tini
@@ -27,10 +27,10 @@ mv build/tini/tini-static build/bin/linux-amd64
 
 # metadata service injector
 (
-	mkdir -p build/inject
-	cd build/inject
-	cmake ../../inject
-	make V=1
+    rm -rf build/inject && mkdir -p build/inject
+    cd build/inject
+    TINI_INCLUDE_DIR=../tini/src TINI_LIBRARY_DIR=../build/tini cmake ../../inject
+    make V=1
 )
 mv build/inject/titus-inject-metadataproxy build/bin/linux-amd64
 mv build/inject/titus-nsenter build/bin/linux-amd64
--- a/inject/titus-nsenter.c
+++ b/inject/titus-nsenter.c
@@ -1,3 +1,4 @@
+#define _GNU_SOURCE 1
 #include <stdio.h>
 #include <stdlib.h>
 #include <linux/limits.h>
@@ -20,6 +21,7 @@
 #include <unistd.h>
 
 #include "shared.h"
+#include "../tini/src/seccomp_fd_notify.h"
 
 struct namespace {
 	int nstype;
@@ -104,7 +106,7 @@ static int set_up_apparmor(char apparmor_profile[1024], int apparmor_fd) {
 
 	// we can use dprintf, but this just makes error handling cleaner
 	memset(writebuf, 0, sizeof(writebuf));
-	n = sprintf(writebuf, "changeprofile %s", apparmor_profile);
+	n = snprintf(writebuf, sizeof(writebuf) - 1, "changeprofile %s", apparmor_profile);
 	BUG_ON(n < 0 || n >= sizeof(writebuf), "Could not generate exec changehat command");
 
 	if (write(apparmor_fd, writebuf, n) != n) {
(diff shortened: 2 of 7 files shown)


alekar:
I have a different opinion about how we should build this, I opened a pull request [here](https://github.com/Netflix/titus-executor/pull/556) to show you.

tini/src/seccomp_fd_notify.c
```diff
@@ -109,22 +109,34 @@ void maybe_setup_seccomp_notifer() {
...
+		int result = -1;
+		for (int i=1; i<=attempts; i++) {
+			result = connect(sock_fd, (struct sockaddr*)&addr, sizeof(addr));
+			if (result != -1) {
+				break;
+			}
+			sleep(1);
 		}
-
-		if (prctl(PR_SET_NO_NEW_PRIVS, 1, 0, 0, 0)) {
```
alekar, line 126: Any reason to drop this? Comment in the commit explaining would be nice to have.

alekar (changes requested):

tini/src/seccomp_fd_notify.c
solarkennedy: Yea. Realistically I don't think I could have gotten away with this in the long term.
Specifically, this prevents newly forked processes forking differently than their parent.
With this in place, we couldn't exit sshd.

solarkennedy:
> I have a different opinion about how we should build this, I opened a pull request [here](https://github.com/Netflix/titus-executor/pull/556) to show you.

Thank you. I see you understand cmake a lot better than I do.
I'll incorporate your changes.

alekar (approved):

hack/builder/titus-executor-builder.sh
```diff
@@ -17,7 +17,7 @@ mv mount/titus-mount-block-device build/bin/linux-amd64/
 
 # tini
 (
-    mkdir -p build/tini
+    rm -rf build/tini && mkdir -p build/tini
```
sargun, line 20: wont this fail if the dir `build` doesn't exist?

inject/titus-nsenter.c
```diff
@@ -104,7 +106,7 @@ static int set_up_apparmor(char apparmor_profile[1024], int apparmor_fd) {
 
 	// we can use dprintf, but this just makes error handling cleaner
 	memset(writebuf, 0, sizeof(writebuf));
-	n = sprintf(writebuf, "changeprofile %s", apparmor_profile);
+	n = snprintf(writebuf, 1024, "changeprofile %s", apparmor_profile);
```
sargun, line 109: Please do not use 1024, and instead sizeof(writebuf) - 1.

tini/src/seccomp_fd_notify.c
```diff
@@ -109,22 +110,34 @@ void maybe_setup_seccomp_notifer() {
...
-		sock_fd = socket(AF_UNIX, SOCK_STREAM, 0);
+		// Sometimes things are not perfect, and the socket is not ready at first
+		// Instead of enforcing strict ordering, we can be defensive and retry.
+		int attempts = 10;
+		for (int i=1; i<=attempts; i++) {
+			sock_fd = socket(AF_UNIX, SOCK_STREAM, 0);
+			if (sock_fd != -1) {
+				break;
+			}
+			sleep(1);
```
sargun, line 121: Please write a warning message or something when this happens.

tini/src/seccomp_fd_notify.c
```diff
@@ -109,22 +110,34 @@ void maybe_setup_seccomp_notifer() {
 	if (socket_path) {
 
 		int sock_fd = -1;
-		sock_fd = socket(AF_UNIX, SOCK_STREAM, 0);
+		// Sometimes things are not perfect, and the socket is not ready at first
+		// Instead of enforcing strict ordering, we can be defensive and retry.
+		int attempts = 10;
+		for (int i=1; i<=attempts; i++) {
```
sargun, line 116: `(int i = 1; i <= attempts; i++)`

tini/src/seccomp_fd_notify.c
```diff
@@ -109,22 +110,34 @@ void maybe_setup_seccomp_notifer() {
...
+			sock_fd = socket(AF_UNIX, SOCK_STREAM, 0);
+			if (sock_fd != -1) {
+				break;
```
sargun, line 119: I would do sock_fd >= 0

tini/src/seccomp_fd_notify.c
sargun: Can you even legally do this?

sargun (approved):

hack/builder/titus-executor-builder.sh
solarkennedy: No, `-f` will mean it wont fail

tini/src/seccomp_fd_notify.c
solarkennedy: I think ideally we would PR_SET_NO_NEW_PRIVS, but I think in practice we can't

inject/titus-nsenter.c
solarkennedy: @alekar this came from your cmake change commit.

alekar: True. Also, writebuf needs to be larger.

```
prod titusagent-main01cell003-m5.metal.001-v002490 us-west-2 i-0516f98bdf9357b3f
(nfsuper) ~ $ sudo cat /sys/module/apparmor/parameters/path_max
8192
```

solarkennedy: Ill make a second review for this

PR #556, opened 2021-03-05
Tsa suggestions

alekar:
This is a suggestion about how we build the changes for tsa.


alekar:
I only opened this pull request to demonstrate some ideas around building tini and inject.

solarkennedy:
Hehe, this doesn't actually build. But let me see if I can take it from here with the library approach you are suggesting.

alekar:
I intended it as a sketch. There was a typo in the `titus-executor-builder.sh`. I have pushed a fix.

PR #557, opened 2021-03-05
Tighten the tini ioctl seccomp filter

solarkennedy:
Previously we were sending every single
ioctl call to TSA, which was noisy.
    
Now we only send the ioctls that require
privilege (perf-related) to TSA to handle.

----

Some clang-format noise too
Also tini makefile helpers


sargun (commented):
lgtm

PR #559, opened 2021-03-09
Autoformat all C code

solarkennedy:
My editor always tries to format on save, making a much larger diff than I want.

Instead of fighting it, I would like to bring autoformatting to our C code just like our go code.

The .clang-format file came from the TSA (which I copied from the linux kernel I think).


alekar (approved):
Cool.

sargun:
lgtm

PR #560, opened 2021-03-09
Upped the max apparmor path to 8192

solarkennedy:


inject/titus-nsenter.c
```diff
@@ -99,9 +99,8 @@ int populate_namespaces(int titus_pid_1_fd, int namespace_fds[]) {
 	return ret;
 }
 
-static int set_up_apparmor(char apparmor_profile[1024], int apparmor_fd) {
-	// This will break on older kernels
-	char writebuf[1024];
+static int set_up_apparmor(char apparmor_profile[8192], int apparmor_fd) {
+	char writebuf[8192];
```
alekar, line 103: If we're pedantic this should be,  `8192 + strlen("changeprofile ")`.

alekar (approved):

PR #561, opened 2021-03-10
Removed init command functionality from system sidecars

solarkennedy:
Trying to simplify system sidecars a bit.

I think we can get away with this instead of having an "init command" or an "init container".

On staging, this works. However, the second time it runs I get this:

```
time="2021-03-10T03:46:12Z" level=error msg="exec failed: container_linux.go:349: starting container process caused \"read init-p: connection rese...
```
Which me think it should *not* work? But I guess the init command is saving the day.

### DIFF
--- a/executor/runtime/docker/docker_linux.go
+++ b/executor/runtime/docker/docker_linux.go
@@ -3,13 +3,11 @@
 package docker
 
 import (
-	"bytes"
 	"context"
 	"fmt"
 	"io/ioutil"
 	"net"
 	"os"
-	"os/exec"
 	"path/filepath"
 	"strconv"
 	"strings"
@@ -44,7 +42,6 @@ type serviceEnabledFunc func(cfg *config.Config, c runtimeTypes.Container) bool
 
 type serviceOpts struct {
 	humanName    string
-	initCommand  string
 	required     bool
 	unitName     string
 	enabledCheck serviceEnabledFunc
@@ -55,7 +52,6 @@ const (
 	systemServiceStartTimeout = 90 * time.Second
 	umountNoFollow            = 0x8
 	sysFsCgroup               = "/sys/fs/cgroup"
-	runcArgFormat             = "--root /var/run/docker/runtime-%s/moby exec --user 0:0 --cap CAP_DAC_OVERRIDE %s %s"
 	defaultOomScore           = 1000
 )
 
@@ -93,7 +89,6 @@ var systemServices = []serviceOpts{
 		humanName:    "metatron",
 		unitName:     "titus-metatron-sync",
 		required:     true,
-		initCommand:  "/titus/metatron/bin/titus-metatrond --init",
 		enabledCheck: shouldStartMetatronSync,
 	},
 	{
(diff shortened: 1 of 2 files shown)


JackOfMostTrades:
Will this still have the effect of not starting the task if the "metatrond --init" fails? I _think_ that's a desirable feature of the current behavior.

solarkennedy:
> Will this still have the effect of not starting the task if the "metatrond --init" fails? I _think_ that's a desirable feature of the current behavior.

Yes. Here is what the systemd logs look like (it is currently configured to try to start 10 times:
```
 titus-metatron-sync@b4376dbf-fcab-49da-9a1c-5fb5f403510e.service: Control process exited, code=exited status=1
 titus-metatron-sync@b4376dbf-fcab-49da-9a1c-5fb5f403510e.service: Failed with result 'exit-code'.
 Failed to start Metatron sync for container b4376dbf-fcab-49da-9a1c-5fb5f403510e.
 titus-metatron-sync@b4376dbf-fcab-49da-9a1c-5fb5f403510e.service: Service hold-off time over, scheduling restart.
 titus-metatron-sync@b4376dbf-fcab-49da-9a1c-5fb5f403510e.service: Scheduled restart job, restart counter is at 1.
 Stopped Metatron sync for container b4376dbf-fcab-49da-9a1c-5fb5f403510e.
 Starting Metatron sync for container b4376dbf-fcab-49da-9a1c-5fb5f403510e...
 Invalid argument: --foo
```

It comes up as task failed with:

```
Finished - (failed) 03/10/2021 09:04:56
could not start metatron service (titus-metatron-sync@b4376dbf-fcab-49da-9a1c-5fb5f403510e.service): failed
Please refer to our guide on Debugging Containers.
```

JackOfMostTrades (approved):

alekar (approved):

rgulewich:
Hmmm, we moved away from `ExecStartPre` in 6a290be408f1fa55f22e5f02f7d89102d5f88ab9, and I'm trying to remember exactly why (should've added better notes, I guess).  I _think_ it was for two reasons:

1) It was race-y somehow - I think there were instances where the ExecStartPre would take a while, and systemd would say that the unit was started when there weren't actually creds on disk.
2) The error message returned wasn't super helpful to users, resulting in them pinging someone in the Titus or Security channels - by putting the message in the hands of the service owners (Metatron in this case), it helped users help themselves (eg: by bubbling up that it was a secret issue vs an internal error).

I'm definitely in favour of simplifying this code, but I want to make sure that we're not regressing on either of those.

solarkennedy:
> Hmmm, we moved away from `ExecStartPre` in [6a290be](https://github.com/Netflix/titus-executor/commit/6a290be408f1fa55f22e5f02f7d89102d5f88ab9), and I'm trying to remember exactly why (should've added better notes, I guess). I _think_ it was for two reasons:
> 
> 1. It was race-y somehow - I think there were instances where the ExecStartPre would take a while, and systemd would say that the unit was started when there weren't actually creds on disk.

Hmm. That really shouldn't happen. @JackOfMostTrades is there any situation where --init would return 0 but have no creds on disk? (I'm assuming system is actually doing what it should)

Another thing this could have been was the timeout bug I fixed, where if 1 unit takes a long time, other units "fail" with deadline exceeded?

> 2. The error message returned wasn't super helpful to users, resulting in them pinging someone in the Titus or Security channels - by putting the message in the hands of the service owners (Metatron in this case), it helped users help themselves (eg: by bubbling up that it was a secret issue vs an internal error).

You can see my example of what the error message looks like (assuming that Pre returns non-0) above. It pinpoints metatron exactly as the culprit.

> 
> I'm definitely in favour of simplifying this code, but I want to make sure that we're not regressing on either of those.

JackOfMostTrades:
> Hmm. That really shouldn't happen. @JackOfMostTrades is there any situation where --init would return 0 but have no creds on disk? (I'm assuming system is actually doing what it should)

No, if the --init run exits with status code 0 then creds should be available. A quick read of the man pages suggests that for Type=simple (which should be the implicit service type here) the unit won't show as started until after ExecStartPre is done and ExecStart has fork-exec'd. So I think this should work.

solarkennedy:
This PR is on hold because I'm not sure if we want to use *more* systemd or less.

On the plus side, systemd is cool. On the downside, we are overloading it.

Like my overseer branch, this one service (metatron) is the only thing that uses this one special feature of systemd, but it is currently implemented in golang directly.

Maybe we should take it as an example! :)

solarkennedy:
Closing this for now, seems like if it is not broke, not fix.
As we move towards nested containers, we might revive this PR to double-down on systemd constructs.

PR #563, opened 2021-03-10
Allow arbitrary EFS hostnames

solarkennedy:
This is change to allow one to bypass the standard
EFS hostname interpolation, and instead pass in
your own NFS hostname as the "EFS ID".

A hack, but for the usecase where you just have an
NFS hostname you want, this will work.

----

This is a rebase of @rgulewich 's changes.

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -1315,15 +1315,34 @@ func (r *DockerRuntime) processEFSMounts(c runtimeTypes.Container) ([]efsMountIn
 			return nil, fmt.Errorf("Invalid EFS mount (read/write flag): %+v", configInfo)
 		}
 
-		if r.awsRegion == "" {
-			// We don't validate at client creation time, because we don't get this during testing.
-			return nil, errors.New("Could not retrieve EC2 region")
+		isRealEfsID, err := isEFSID(emi.efsFsID)
+		if err != nil {
+			return nil, err
 		}
-		emi.hostname = fmt.Sprintf("%s.efs.%s.amazonaws.com", emi.efsFsID, r.awsRegion)
+		if isRealEfsID {
+			if r.awsRegion == "" {
+				// We don't validate at client creation time, because we don't get this during testing.
+				return nil, errors.New("Could not retrieve EC2 region")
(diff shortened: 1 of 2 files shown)


executor/runtime/docker/docker.go
```diff
@@ -1315,11 +1315,18 @@ func (r *DockerRuntime) processEFSMounts(c runtimeTypes.Container) ([]efsMountIn
...
-		if r.awsRegion == "" {
-			// We don't validate at client creation time, because we don't get this during testing.
-			return nil, errors.New("Could not retrieve EC2 region")
+		isRealEfsID := strings.HasPrefix(emi.efsFsID, "fs-")
+		if isRealEfsID {
+			if r.awsRegion == "" {
+				// We don't validate at client creation time, because we don't get this during testing.
+				return nil, errors.New("Could not retrieve EC2 region")
+			} // Get the remote IP. -- Is this really the best way how? Go doesn't have a simpler API for this?
+			emi.hostname = fmt.Sprintf("%s.efs.%s.amazonaws.com", emi.efsFsID, r.awsRegion)
```
sargun, line 1324: The hostname can begin with a fs-* I would say if the regexp matches:

`^fs-[0-9a-f]+$`

then it is an EFS.

solarkennedy: got it. ty

executor/runtime/docker/docker.go
```diff
@@ -1315,17 +1315,33 @@ func (r *DockerRuntime) processEFSMounts(c runtimeTypes.Container) ([]efsMountIn
...
 	}
 
 	return efsMountInfos, nil
 }
 
+func isEFSID(FsID string) bool {
+	matched, err := regexp.MatchString(`^fs-[0-9a-f]+$`, FsID)
+	if err != nil {
+		// The only type of errors that might hit this are regex compile errors
+		log.Fatalf("Something went really wrong determining if '%s' is an EFS ID: %s", FsID, err)
```
sargun, line 1340: log.witherror...

sargun (approved):

executor/runtime/docker/docker.go
solarkennedy: I'll go ahead and bubble this back up to the caller, which returns an error

PR #564, opened 2021-03-11
Update ssh apparmor profiles a bit

solarkennedy:
1. Remove pid. apparmor was denying it, but we dont need it anyway
2. Allow unconfined signals. This is to allow it to be
actually killed by something, anything. Mostly systemd.
This is to enable restarting the sshd daemon, if not just
for debug purposes.

----
We've probably all seen
```
Mar 11 00:45:50 titusagent-devkyleacell001-m524xlarge000-i-09ddb9b9159899a2e audit[2890106]: AVC apparmor="ALLOWED" operation="open" profile="docker_titus//sshd" name="/run/sshd.pid" pid=2890106 comm="sshd" requested_mask="wc" denied_mask="wc" fsuid=10000 ouid=10000
```

### DIFF
--- a/root/etc/apparmor.d/docker_titus
+++ b/root/etc/apparmor.d/docker_titus
@@ -34,6 +34,8 @@ profile docker_titus flags=(attach_disconnected,mediate_deleted) {
   /titus/sshd/usr/sbin/sshd Cx,
 
   profile sshd /titus/sshd/usr/sbin/sshd {
+    # Allow signals from unconfined, usually systemd
+    signal (receive) peer="unconfined",
     signal (send,receive) peer=@{profile_name},
 
     network,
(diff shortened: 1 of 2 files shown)


root/etc/apparmor.d/docker_titus
```diff
@@ -34,6 +34,8 @@ profile docker_titus flags=(attach_disconnected,mediate_deleted) {
   /titus/sshd/usr/sbin/sshd Cx,
 
   profile sshd /titus/sshd/usr/sbin/sshd {
+    # Allow signals from unconfined, usually systemd
+    signal (send,receive) peer="unconfined",
```
sargun, line 38: Can we restrict this to cap sysadmin on the host or root on the host?

solarkennedy: I don't see how to do that given the grammar:
http://manpages.ubuntu.com/manpages/xenial/man5/apparmor.d.5.html

The only thing it looks like I have access to is peer for signal stuff.

sargun (approved):

root/etc/apparmor.d/docker_titus
sargun: should this be signal receive and not send to unconfined?

solarkennedy: ah right yes

PR #565, opened 2021-03-11
Allow for the excess ENIs to be overridden

sargun:
This changes the behaviour that rather than have 50 warm ENIs per subnet, allow this number to be overridden on the fly.
This is valueable for:
* Incidents where ENI creation is b0rked
* Smaller installations where we don't want to burn 50 ENIs / IPs due to waste.

### Description of the Change


solarkennedy:
Can you say more about your 1, 2 scenarios?

1. In this case, are you thinking we would adjust this on the fly?
But if we went from 50=>10, would that mean that up to 40 enis per agent would get gc'd right away?

2. We have small installations? :)
Seriously though, what is the use-case?

sargun:
But if we went from 50=>10, would that mean that up to 40 enis per agent would get gc'd right away?

These are ENIs per subnet that are "floating".

solarkennedy (approved):

PR #567, opened 2021-03-13
Use the ec2 instance metadata service to get the instance id for titus-storage

solarkennedy:
The EC2_INSTANCE_ID var is not something this sidecar can
rely on, because it gets overwritten as the titus task ID.

Instead, I can call the IMDS to get the "real" instance id.

Also correcting a RW/RO flags issue. I'll clean this up later
when I unify EFS mounting into titus-storage.

### DIFF
--- a/cmd/titus-storage/ebs.go
+++ b/cmd/titus-storage/ebs.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"log"
 	"os"
 	"path/filepath"
 	"regexp"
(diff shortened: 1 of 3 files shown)


alekar:
This environment is available in /etc if you need it.

```
nfsuper) ~ $ cat /etc/nflx/static-environment  | grep INSTANCE
EC2_INSTANCE_ACTION="none"
EC2_INSTANCE_ID="i-074a40642f89c7154"
EC2_INSTANCE_LIFE_CYCLE="on-demand"
EC2_INSTANCE_TYPE="m5.16xlarge"
NETFLIX_INSTANCE_ID="i-074a40642f89c7154"
```

But, I like the idea of using the IMDS better, since it does not create a dependency on things on host's filesystem.

alekar (approved):

sargun (approved):

PR #568, opened 2021-03-16
New VK Backend

sargun:
This VK Backend has a few changes:

* It no longer uses pipes. Instead, it just drops the status
  as a flat file on disk
* It moves the systemd unit for initialization of the titus executor
  into titus-executor repo (this sheds some stuff)


sargun:
@solarkennedy Can you make sure this doesn't interfere with your storage work? Can you add detach / attach bits as drop-in extensions?

solarkennedy:
So the process tree will be:
```
vk ->
  vk-backend ? ->
    titus-executor ->
      docker stuff or whatever
```
Why though? How does this save systemd any burden?

sargun:
vk-backend is titus-executor -- just a different API to titus-executor. It uses flat files to pass state back and forth instead of unix pipes.

sargun:
This does introduce one kind of big change, which is if you're killing titus-executors on the host, you have to run:

```
systemctl kill --kill-who=main titus-executor@...
```

We can work around this by having the VK originate its own status if the unit is killed in an unexpected fashion. Or add an ExecStopPost which holds out for 10 seconds or something to try to have the VK pickup the state.

rgulewich (approved):

root/lib/systemd/system/titus-executor-backend@.service
```diff
@@ -0,0 +1,19 @@
+[Unit]
+Description=Titus Virtual Kubelet Executor Service
+
+[Service]
+Slice=containers.slice
+ExecStartPre=/bin/mkdir -p /run/titus-executor/pod-status
+ExecStart=/apps/titus-executor/bin/run-titus-virtual-kubelet-backend.sh --runtime-dir=/run/titus-executor/%i
+Restart=no
```
sargun, line 8: curl http://localhost:vk/foo -d @/run/titus-executor/%i/statejson

alekar (approved):

solarkennedy (approved):
I would like to know what your plan is with regards to rolling this out immutably.

PR #569, opened 2021-03-16
No systemd masters, no dbus gods, only man pages

solarkennedy:
RFC: Use native golang process supervision instead of systemd

System in theory is great, but in practice we are hitting it too hard.

This POC is a RFC for what it would look like if we only did the "supervision"
of system, but in native golang here, using this overseer package.

```
| Thing        | Old World       | New World                                    |
|-------------------------------------------------------------------------------|
| config       | systemd units   | labels, config maps, json, structs, anything |
| supervisor   | systemd         | titus-executor / overseer                    |
| logs         | journald        | journald                                     |
| control      | systemctl       | N/A                                          |
| status       | systemctl       | pod status.conditions?                       |
```

Once nice thing about this approach is how incremental it is.
We can start with just doing like, abmetrix or something
low-stakes, while we get overseer *right*, then we can convert
units one-by-one, till we don't need systemd for Titus System Services
at all.

-----

Yes, this approach is kinda the "opposite" of the direction the
nested containers + systemd POC is going. This is intentional.
I think we can achieve our goals of supporting multi-container
workloads *and* not require the complexity of nested containers
and nested systemds. I think the solution proposed here is simpler
than other proposals, and can incrementally be achieved.

### Description of the Change


sargun:
So, I think I like the idea, but can we move this logic into an independent go binary? I think something like titus-sshd and titus-metadata service might be okay to have invoked by systemd, but it'd be nice if:
1. We have _one last_ system sidecar called the supervisor, whichruns our supervision work. 
2. We have tini able to fork exec the workload

titus-metadata-proxy uses techniques like systemd watchdog and readiness checks that are just too valuable to give up in my opinion.

solarkennedy:
> So, I think I like the idea, but can we move this logic into an independent go binary? I think something like titus-sshd and titus-metadata service might be okay to have invoked by systemd, but it'd be nice if:
> 
> 1. We have _one last_ system sidecar called the supervisor, whichruns our supervision work.

Interesting idea. Maybe it could be named `titus-system-services-supervisor@id`?
I see where you are going with this, its just a lot of extra plumbing to get all the data + metadata we need out of the golang world and over to a dedicated binary. There is just a lot of interaction and it is easy when, you know, you can pass args in functions. Let me think...

> 2. We have tini able to fork exec the workload
> 

This I don't understand. "workload" here is the system service? Or "workload" == user workload?

> titus-metadata-proxy uses techniques like systemd watchdog and readiness checks that are just too valuable to give up in my opinion.

Hmmmm. My TSA also uses notify (not watchdog though). I agree it sucks to give up that feature, we *could* reimplement the sd_notify + watch server protocol in go too, but yuck. 

Do you happen to know if `systemd-run` is also vulnerable to overloading the systemd pid1? Probably right?

solarkennedy:
Closing as this was just an RFC.

PR #570, opened 2021-03-16
Always mkdir when asked to mount in Titus

solarkennedy:
In kubernetes, if you ask to mount storage, it will
create the folder for you if it doesn't exist.

I think we should match this behavior, on nfs *and* ebs.

### DIFF
--- a/mount/common.h
+++ b/mount/common.h
@@ -1,0 +1,30 @@
+#include <limits.h>
+#include <stdio.h>
+#include <string.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+
+void mkdir_p(const char *dir)
+{
+	char tmp[PATH_MAX];
+	char *p = NULL;
+	size_t len;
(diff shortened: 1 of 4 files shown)


andrew-leung (approved):

mount/common.h
```diff
@@ -0,0 +1,30 @@
...
+
+	snprintf(tmp, sizeof(tmp), "%s", dir);
+	len = strlen(tmp);
+	if (tmp[len - 1] == '/')
+		tmp[len - 1] = 0;
+	for (p = tmp + 1; *p; p++) {
+		if (*p == '/') {
+			*p = 0;
+			if (stat(tmp, &st) == -1) {
+				mkdir(tmp, 0777);
```
sargun, line 22: Any reason not to do the mkdir always?

solarkennedy: idk, just when I was stracing it looked "bad" to see the failed syscalls.

PR #571, opened 2021-03-17
Renamed titus-mount to titus-mount-nfs

solarkennedy:
A (theoretically) noop change to make it more clear that the `titus-mount` command is only for NFS.
I think each new storage mounter we add will have its own special sauce.

Ooo or would you like me to do the existing Unix convention of titus-mount.nfs4, titus-mount.block ?

### DIFF
--- a/mount/titus-mount-nfs.c
+++ b/mount/titus-mount-nfs.c
@@ -34,22 +34,22 @@ static int dns_lookup(const char *hostname, struct sockaddr_in *addr)
 
 	if (inet_aton(hostname, &addr->sin_addr)) {
 		fprintf(stderr,
-			"titus-mount: %s is already an IP. Not doing a DNS lookup\n",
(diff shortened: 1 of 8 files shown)


mount/titus-mount-nfs.c
```diff
@@ -305,15 +305,15 @@ int main(int argc, char *argv[])
 
 	/* Now we can switch net/mount namespaces so we can lookup the ip and eventually mount */
 	switch_namespaces(pidfd);
-	fprintf(stderr, "titus-mount: user-inputed options: %s\n", options);
+	fprintf(stderr, "titus-mount-nfs: user-inputed options: %s\n", options);
```
sargun, line 308: Maybe it makes sense to take some time and write a macro here?

sargun (approved):

mount/titus-mount-nfs.c
solarkennedy: Yes. I can do that next time.

PR #572, opened 2021-03-17
Make titus-seccomp-agent die when a task finishes

solarkennedy:
Unlike many other system sidecars we have, TSA doesn't
inject itself into the pid namespace of a container (yet)?
and doesn't die with the rest of the container "magically".

To make sure it terminates when titus-executor does, preventing
tsa's from hanging around, I make it PartOf titus-executor.

Also, I added a --titus-task-id= command line flag to TSA.
This is kinda a hack, because it is ignored by tsa itself,
but makes it super easy to know which TSA goes to which process
by simply running ps.

### DIFF
--- a/root/lib/systemd/system/titus-seccomp-agent@.service
+++ b/root/lib/systemd/system/titus-seccomp-agent@.service
@@ -1,5 +1,7 @@
 [Unit]
 Description=Titus seccomp agent for container %i
+# PartOf ensures that it will shutdown as part of the main titus-executor when it stops
+PartOf=titus-executor@default__%i.service
 
 StartLimitIntervalSec=30
 StartLimitBurst=5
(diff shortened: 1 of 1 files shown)


MyUmmaGumma (approved):

root/lib/systemd/system/titus-seccomp-agent@.service
```diff
@@ -1,13 +1,15 @@
 [Unit]
 Description=Titus seccomp agent for container %i
+# PartOf ensures that it will shutdown as part of the main titus-executor when it stops
+PartOf=titus-executor@default__%i.service
```
sargun, line 4: Do you want to change this to be partof the new shared target?

sargun (approved):

solarkennedy:
This is on hold as @sargun 's PR will probably obsolete it.

PR #574, opened 2021-03-19
Updated the to the latest default.json seccomp policy from docker

solarkennedy:
This new policy has updated syscalls for modern kernels.
I went ahead and removed the CVE-2020-14386 socket thing
because we are running linux 5.10 now.

Only doing default.json at first because it is the simplest.
I'll do other json files in different commits if I've got this one
right.


sargun (approved):

solarkennedy:
Hmm.. the standalone test is failing... I wonder if it is because of kernel stuff...

solarkennedy:
Oh no, just that Testcve202014386 that (I don't think) we need anymore

PR #575, opened 2021-03-19
Added titus-container-tools injection

solarkennedy:
I want to make it easier to provide tools to our users.
Specifically, debug tools and perf tools.

For now, I think I want to provide that via
injecting scripts. At first these scripts will
be very simple (scripts that apt-get stuff), but
later the tools may be more sofisticated (statically linked things).

For this PR, I just set up the path. An internal
netflix package will provide the rest.


sargun:
Did you mean to add  BIN +19.1 MB filesystems/src-logs-010778665/stderr 
?

solarkennedy:
> Did you mean to add BIN +19.1 MB filesystems/src-logs-010778665/stderr
> ?

Lol no

solarkennedy:
I'm going to try different approach with a Volume container instead of bind-mount from the host.

executor/runtime/docker/docker_env.go
```diff
@@ -31,6 +33,11 @@ export {{ $key }}={{ escapeSQWithFallback $key $val }}
 # export {{ $key }}={{ $val | escape_sq }}
 {{ end -}}
 {{ end -}}
+
+# This path setting makes it convenient for users to run
+# tools that have been injected into the container from the
+# Titus agent.
+export PATH=$PATH:/titus/container-tools/bin
```
rgulewich, line 40: Do we think our users would be ok with this?  I'm wondering if it's possible to only add this to the path if we're in an ssh session.  Maybe I'm being overly paranoid here. :)

rgulewich (approved):

executor/runtime/docker/docker_env.go
solarkennedy: It is last in the path, should be ok.

PR #576, opened 2021-03-22
Rebase seccomp allow-perf-syscalls on top of latest docker default.json

solarkennedy:
I tried to do this as clearly as possible by having 2 commits:
1. Resetting to allow-perf-syscalls.json to default.json from docker
2. Applying the changes again to actually allow perf syscalls

It is pretty risky because of the json diffing.


sargun:
I was going to suggest that it might have made more sense to have a jsonnet file to "Adapt" the policy.

solarkennedy:
> I was going to suggest that it might have made more sense to have a jsonnet file to "Adapt" the policy.

Yea I think now would be the right time to evolve to that.
If you agree, then yea let me have some sort of base default.json that comes from upstream, then some other jssonnet files that represent our "diffs", then have `make` do its thing to output json files then gobin ?

PR #578, opened 2021-03-23
Upgrade go-systemd and add common target unit

sargun:
### Description of the Change

### DIFF
--- a/executor/runtime/docker/docker_linux.go
+++ b/executor/runtime/docker/docker_linux.go
@@ -19,7 +19,7 @@ import (
 
 	"github.com/Netflix/titus-executor/config"
 	runtimeTypes "github.com/Netflix/titus-executor/executor/runtime/types"
-	"github.com/coreos/go-systemd/dbus"
+	"github.com/coreos/go-systemd/v22/dbus"
 	"github.com/opencontainers/runc/libcontainer/cgroups"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
@@ -47,6 +47,7 @@ type serviceOpts struct {
 	required     bool
 	unitName     string
 	enabledCheck serviceEnabledFunc
+	target       bool
 }
 
 const (
@@ -59,6 +60,11 @@ const (
 )
 
 var systemServices = []serviceOpts{
+	{
+		unitName: "titus-container",
+		required: true,
+		target:   true,
+	},
 	{
 		unitName:     "titus-sidecar-spectatord",
 		enabledCheck: shouldStartSpectatord,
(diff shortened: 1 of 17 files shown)


sargun:
We should be able to add a drop-in unit in the titus-virtual-kubelet saying that the target is PartOf titus-executor@default__%i.service, and then when that's stopped by VK, everything should "just work"

executor/runtime/docker/docker_linux.go
```diff
@@ -59,6 +60,11 @@ const (
 )
 
 var systemServices = []serviceOpts{
+	{
+		unitName: "titus-container",
```
solarkennedy, line 64: Would this be better named, `titus-sidecar-container` ?

Or, is titus-executor PartOf titus-container?

Why do we need another unit though, why not just have things part of titus-executor?

executor/runtime/docker/docker_linux.go
```diff
@@ -245,7 +270,11 @@ func runServiceInitCommand(ctx context.Context, log *logrus.Entry, cID string, r
 }
 
 func startSystemdUnit(ctx context.Context, conn *dbus.Conn, taskID string, cID string, runtime string, opts serviceOpts) error {
-	qualifiedUnitName := fmt.Sprintf("%s@%s.service", opts.unitName, taskID)
+	postFix := "service"
+	if opts.target {
```
solarkennedy, line 274: Why can't we just put `.service` or `.target` in the unit name itself?

solarkennedy (approved):

solarkennedy:
Can you confirm: if something fails to stop, is it even bad? (does it bubble up as an error to the user)
Until recently 'titus-sshd@container.service stop' didn't work, there might be others.
Plus, if it is already dead, does it still complain with an error?

Basically I want to make sure this change doesn't the existing behavior of "not caring too much" if sidecars can't be stopped. The show must go on here.

executor/runtime/docker/docker_linux.go
sargun: and parse / unparse it here? That seems uglier than just having a flag for it.

executor/runtime/docker/docker_linux.go
sargun: the target unit isn't a sidecar though. titus-container-sidecars@.target?

executor/runtime/docker/docker_linux.go
solarkennedy: No, just not even parse it at all. Why do we need to parse it?
Just print the whole unit name

sargun: because you need to have something that inserts the container ID between the "@" and the "."

sargun:
> Can you confirm: if something fails to stop, is it even bad? (does it bubble up as an error to the user)
> Until recently '[titus-sshd@container.service](mailto:titus-sshd@container.service) stop' didn't work, there might be others.
> Plus, if it is already dead, does it still complain with an error?
> 
> Basically I want to make sure this change doesn't the existing behavior of "not caring too much" if sidecars can't be stopped. The show must go on here.

Correct

PR #579, opened 2021-03-24
Disable uploads using noop backend.

alekar:
Allows security sensitive workloads to disable log uploads.


rgulewich (approved):

sargun:
Errm. Hrm. Why? Should we make it so you can set the backends at runtime / pod startup time?

alekar:
No. Ultimately, we should make it so you can use whatever you want to
upload logs (insight-logs, filebeat, fluentd ...). This is just filling a
feature gap for now.

PR #580, opened 2021-03-24
Make titus-nsenter (sshd) play well with seccomp notify

solarkennedy:
If you use no_new_privs, like you get when you get a
seccomp-notify filter for TSA, you are not allowed
to switch AppArmor profiles on exec.

You can have an AppArmor profile installed, it just has
to be the exact same profile that you get after
you fork.

This is because the kernel can't be 100% sure that you
wont get new privileges after exec, AppArmor just denies you.

Also expanded the profile to include signals to systemd-shutdown.

### DIFF
--- a/root/etc/apparmor.d/docker_fuse
+++ b/root/etc/apparmor.d/docker_fuse
@@ -1,5 +1,8 @@
 #include <tunables/global>
 
+# DEBUGGING THIS FILE!? :)
+# See the top note in the docker_titus file.
+
 profile docker_fuse flags=(attach_disconnected,mediate_deleted) {
   #include <abstractions/base>
 
@@ -9,6 +12,8 @@ profile docker_fuse flags=(attach_disconnected,mediate_deleted) {
   umount,
   mount,
 
+  # Allow signals send to titus-sshd, usually systemd-shutdown ends up sending these out on shutdown
+  signal (send) peer="docker_titus//sshd",
   signal (send,receive) peer=@{profile_name},
 
   deny mount fstype=tmpfs,
@@ -43,9 +48,11 @@ profile docker_fuse flags=(attach_disconnected,mediate_deleted) {
 
   deny /titus/** wl,
 
-/titus/sshd/usr/sbin/sshd Cx,
+  /titus/sshd/usr/sbin/sshd Cx,
 
   profile sshd /titus/sshd/usr/sbin/sshd {
+    # Allow signals from unconfined, usually systemd on the host
+    signal (receive) peer="unconfined",
     signal (send,receive) peer=@{profile_name},
 
     network,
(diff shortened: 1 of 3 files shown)


root/etc/apparmor.d/docker_fuse
```diff
@@ -9,6 +12,8 @@ profile docker_fuse flags=(attach_disconnected,mediate_deleted) {
   umount,
   mount,
 
+  # Allow signals send to unconfined, usually systemd-shutdown ends up sending these out
+  signal (send) peer="unconfined",
```
sargun, line 16: isn't it received signals from systemd-unconfined?

solarkennedy: No, it sure looked like it was send from the audit log. I'll double check.

solarkennedy: ```Mar 24 22:48:19 titusagent-devint01cell001-r5metal000-i-0b13e629e19cb6726 kernel: [106147.137358] audit: type=1400 audit(1616626099.646:222): apparmor="DENIED" operation="signal" profile="docker_titus" pid=2006590 comm="systemd-shutdow" requested_mask="send" denied_mask="send" signal=cont peer="docker_titus//sshd"```

solarkennedy: Oh I'm wrong about the peer, I could make it be sshd.

So this is 

    systemd-shutdown in the container -> signaling titus-sshd to die

solarkennedy: fixed

solarkennedy:
Crap @sargun , that isn't a 100% solution. This happens when I try to *actually* ssh:
```Mar 25 19:23:46 titusagent-devint01cell001-r5metal000-i-0b13e629e19cb6726 kernel: [180274.905294] audit: type=1400 audit(1616700226.894:259): apparmor="ALLOWED" operation="exec" info="no new privs" error=-1 profile="docker_titus//sshd" name="/bin/bash" pid=3629979 comm="sshd" requested_mask="x" denied_mask="x" fsuid=70004 ouid=10000 target="docker_titus"```

I'm guessing because it is trying to *switch* again back to `docker_titus` for the normal confinement, which is not allowed with no_new_privs on.

sargun:
The check looks a little more complicated than that: https://elixir.bootlin.com/linux/latest/source/security/apparmor/label.c#L562

Any idea what an aa label is?

solarkennedy:
> The check looks a little more complicated than that: https://elixir.bootlin.com/linux/latest/source/security/apparmor/label.c#L562
> 
> Any idea what an aa label is?

I think this is internal kernel jargon borrowed from "selinux labels". It really just looks at the profile name I think:
https://elixir.bootlin.com/linux/latest/source/security/apparmor/include/label.h#L125

solarkennedy:
Even if we get this right, having NO_NEW_PRIVs on breaks sudo.
Closing in favor of #583

PR #581, opened 2021-03-26
Unify sidecars configs into a single place

solarkennedy:
This change is in preparation to elevate our system services to a new level.
Step 1: put all our sidecar config in one place, and just have a map.

Yes, we want to be able to extract them into a new multi-container paradigm, but honestly I'm looking for more short term goals:

1. Make non-required sidecars "ok" to fail if their image doesn't exist. If someone typos a FP on abmetrix, I don't want the launch to fail (currently does, but can trivially be made not to after this refactor)
2. Get us one step closer to extracting the logic of what sidecars to run and what version to run them at *out* of titus-excutor, and into the webhook annotating pods. This is the first part of my system services 2.0 doc, but we don't need to take that all the way to the end. At first it may be just adding annotations, but eventually it may be adding entire containers into the spec.

This is "just" a refactor here. No behavior changes are expected with this PR. It works on my dev stack.
Probably the trickiest thing here is the pointer nonsense with containerName.


executor/runtime/docker/docker.go
```diff
@@ -947,35 +943,14 @@ func (r *DockerRuntime) Prepare(parentCtx context.Context) error { // nolint: go
...
-	if shouldStartServiceMesh(&r.cfg, r.c) {
-		group.Go(r.createVolumeContainerFunc(sidecarConfigs[runtimeTypes.SidecarServiceServiceMesh], &serviceMeshContainerName))
-	}
-
-	if shouldStartAbmetrix(&r.cfg, r.c) {
-		group.Go(r.createVolumeContainerFunc(sidecarConfigs[runtimeTypes.SidecarServiceAbMetrix], &abmetrixContainerName))
+	for sidecarName, sidecarConfig := range sidecarConfigs {
+		if sidecarConfig.Volumes != nil && sidecarConfig.EnabledCheck(&r.cfg, r.c) {
+			sidecarConfig.ContainerName = sidecarName
+			group.Go(r.createVolumeContainerFunc(sidecarName, sidecarConfig))
```
sargun, line 949: Is this valid, or does it create a pointer leak?

executor/runtime/docker/docker.go
```diff
@@ -795,11 +795,11 @@ func cleanContainerName(prefix string, imageName string) string {
 }
 
 // createVolumeContainerFunc returns a function (suitable for running in a Goroutine) that will create a volume container. See createVolumeContainer() below.
-func (r *DockerRuntime) createVolumeContainerFunc(sCfg *runtimeTypes.SidecarContainerConfig, containerName *string) func(ctx context.Context) error {
+func (r *DockerRuntime) createVolumeContainerFunc(serviceName string, sCfg *runtimeTypes.ServiceOpts) func(ctx context.Context) error {
```
alekar, line 798: `sOpts *runtimeTypes.ServiceOpts` perhaps?

rgulewich, line 798: Why's `serviceName` a separate param here?  Seems like it could be rolled up into `ServiceOpts`, no?

rgulewich (approved):

executor/runtime/docker/docker.go
```diff
@@ -947,35 +943,14 @@ func (r *DockerRuntime) Prepare(parentCtx context.Context) error { // nolint: go
...
-
-	if shouldStartServiceMesh(&r.cfg, r.c) {
-		group.Go(r.createVolumeContainerFunc(sidecarConfigs[runtimeTypes.SidecarServiceServiceMesh], &serviceMeshContainerName))
-	}
-
-	if shouldStartAbmetrix(&r.cfg, r.c) {
-		group.Go(r.createVolumeContainerFunc(sidecarConfigs[runtimeTypes.SidecarServiceAbMetrix], &abmetrixContainerName))
+	for sidecarName, sidecarConfig := range sidecarConfigs {
+		if sidecarConfig.Volumes != nil && sidecarConfig.EnabledCheck(&r.cfg, r.c) {
+			sidecarConfig.ContainerName = sidecarName
```
alekar, line 948: Do you need `sidecarName` as an argument if `sidecarConfig` contains it?

executor/runtime/docker/docker_linux.go
```diff
@@ -208,8 +134,14 @@ func setupSystemServices(parentCtx context.Context, c runtimeTypes.Container, cf
...
-	for _, svc := range systemServices {
-		if svc.enabledCheck != nil && !svc.enabledCheck(&cfg, c) {
+	// TODO: it would be nice not to fetch this twice
+	sidecars, err := c.SidecarConfigs()
+	if err != nil {
+		return err
+	}
+	// TODO: Can we somehow make sure titus-container always starts first?
+	for _, svc := range sidecars {
+		if svc.EnabledCheck != nil && !svc.EnabledCheck(&cfg, c) {
```
alekar, line 144: Can `EnabledCheck` be nil? If so, you'll also have to test when you create volume containers, for example.

executor/runtime/docker/docker.go
solarkennedy: I was very confused about this pointer stuff.
I don't really know the answer. Somehow we go from a "base" container name to an expanded one that docker will like.
The code did seem work find in this PR, I don't know if there is a leak or not. If there is it will only be a few here for the lifetime of the task?

executor/runtime/docker/docker.go
solarkennedy: Can do.

The important thing is to not mix this with containerName, which is a long mutated string.

executor/runtime/docker/docker.go
solarkennedy: You are right, it is easier if the Config has the name in it.
I'll redo to be a list of sidcars instead of a map, so we can just pass around one sidecar object that "knows its name"

executor/runtime/docker/docker_linux.go
solarkennedy: I think in practice we happen to not have any sidecars that have volumes but *not* a enabledCheck.
But good idea, I'll make sure to test just in case we ever have one.

executor/runtime/docker/docker.go
solarkennedy: fixed

executor/runtime/docker/docker.go
solarkennedy: Can you clarify though (it may be outdated now with the changes in the PR) how a leak would happen?

Each `sidecarConfig` has a `ContainerName`, should be self-contained?

alekar (approved):

PR #582, opened 2021-03-26
Added kernel module, debug, and headers when perf is enabled

solarkennedy:
Many perf tools require linux headers and other
related kernel-specific files to operate.

Instead of requiring every container to have these packages,
bind-mounting them is more convenient, and ensures that they
are always accurate and correct for the running kernel
as it changes.

Additionally it works for any container, regardless of their distro.

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -1100,6 +1100,9 @@ func (r *DockerRuntime) Prepare(ctx context.Context) error { // nolint: gocyclo
 	r.volumeContainers = volumeContainers
 
 	bindMounts = append(bindMounts, getLXCFsBindMounts()...)
+	if r.c.SeccompAgentEnabledForPerfSyscalls() {
+		bindMounts = append(bindMounts, getKernelBindMounts()...)
+	}
 
 	dockerCfg, hostCfg, err = r.mainContainerDockerConfig(r.c, bindMounts, size, volumeContainers)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


executor/runtime/docker/docker.go
```diff
@@ -1158,6 +1161,31 @@ func (r *DockerRuntime) createTitusEnvironmentFile(c runtimeTypes.Container) err
 	return writeTitusEnvironmentFile(c.Env(), f)
 }
 
+func getUnameR() (string, error) {
+	cmd := exec.Command("/bin/uname", "-r")
```
sargun, line 1165: Can't we call uname directly from the executor? 
https://pkg.go.dev/golang.org/x/sys/unix#Utsname

executor/runtime/docker/docker.go
```diff
@@ -1158,6 +1161,31 @@ func (r *DockerRuntime) createTitusEnvironmentFile(c runtimeTypes.Container) err
...
+
+func getKernelBindMounts() []string {
+	mounts := []string{
+		"/boot:/boot:ro",
+		"/lib/modules:/lib/modules:ro",
+	}
+	unameR, err := getUnameR()
+	if err != nil {
+		kernelHeaders := fmt.Sprintf("/usr/src/linux-headers-%s:/usr/src/linux-headers-%s:ro", unameR, unameR)
+		mounts = append(mounts, kernelHeaders)
```
sargun, line 1184: Will this work with bind mounts, or since the uid is unmapped, will it show up as uid 0? (nobody / unbound?)

solarkennedy: Shows up as nobody/nogroup, works fine.

executor/runtime/docker/docker.go
solarkennedy: Can do.

sargun (approved):

PR #584, opened 2021-03-26
First pass MVP at titus-logviewer using HTTPS

solarkennedy:
Since we use browsers to access this endpoint and read these files,
it is becoming increasingly more important to use HTTPS.

It this is a good approach (seems easy), I can add rotation triggers,
add the env variables to our local config in Netflix to point to
web-ready certs, and change the UI to point at https instead of
http.

I have verified that the web.certs we have *do* respond cleanly
to their IP:<ip>.

### DIFF
--- a/cmd/titus-logviewer/main.go
+++ b/cmd/titus-logviewer/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"crypto/tls"
 	"net/http"
 
 	log2 "github.com/Netflix/titus-executor/utils/log"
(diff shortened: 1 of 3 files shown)


cconger:
LGTM.  I think if you go the `GetCertificate` route on your TLS config it'll allow easier rotation with GetCertificate just loading from a mutex guarded variable that can be updated at some frequency. 

Here's an example doing the pattern I'm thinking: https://github.com/diogomonica/certificate-rotation/blob/master/certificate_rotation.go

cconger (approved):

corindwyer (approved):

solarkennedy:
> LGTM. I think if you go the `GetCertificate` route on your TLS config it'll allow easier rotation with GetCertificate just loading from a mutex guarded variable that can be updated at some frequency.
> 
> Here's an example doing the pattern I'm thinking: https://github.com/diogomonica/certificate-rotation/blob/master/certificate_rotation.go

Cool, ty for the example. I'll get this on the next pass (if I need it?).

rgulewich:
> Cool, ty for the example. I'll get this on the next pass (if I need it?).

You're going to need this - the web certs are rotated every 2 weeks (might be less time now), so this will break after the first rotation.  There's code that does something very similar in `metadataserver/`.

LGTM otherwise, though.

PR #585, opened 2021-03-29
Seccomp rebase default3

solarkennedy:
Final pass of the great seccomp json rebase.


solarkennedy:
@sargun I tried to do this via jsonnet and json patch, neither look good.

With jssonnet, I couldn't figure out how to take an object and "find" the right syscalls block that I needed to edit. 
Using std.objectFields I could iterate over the big map, but I didn't know how I could reliably find the one element of the `syscalls` array (`13`) that I wanted. I think if I could reliably find it, then I could pluck out the elements of the allowed syscalls by name (bpf/perf_event_open, etc), but I couldn't get that step 2.

In json patch method, the patch isn't something that makes any sense. Here is me removing bpf from the SYSADMIN allow blocks:
```
[
  {
    "op": "remove",
    "path": "/syscalls/13/names/0"
  },
  {
    "op": "remove",
    "path": "/syscalls/13/names/11"
  },
  {
    "op": "move",
    "from": "/syscalls/13/names/11",
    "path": "/syscalls/13/names/11"
  },
  {
    "op": "move",
    "from": "/syscalls/13/names/12",
    "path": "/syscalls/13/names/12"
  },
  {
    "op": "move",
    "from": "/syscalls/13/names/13",
    "path": "/syscalls/13/names/13"
  },
  {
    "op": "move",
    "from": "/syscalls/13/names/14",
    "path": "/syscalls/13/names/14"
  },
  {
    "op": "move",
    "from": "/syscalls/13/names/15",
    "path": "/syscalls/13/names/15"
  },
  {
    "op": "move",
    "from": "/syscalls/13/names/16",
    "path": "/syscalls/13/names/16"
  },
  {
    "op": "move",
    "from": "/syscalls/13/names/17",
    "path": "/syscalls/13/names/17"
  }
]
```

Any other ideas? I could do plain old patch/diff? That is kinda what git is doing, but we are missing the semantics of "what" was patched exactly. I could make that more explicit. With a `default.patch`.

solarkennedy:
I've tried the diff/patch approach in #589. @sargun if you like that approach better (compared to this PR's approach where the "diff" is just encoded in git itself) then let me know and I'll close this one in favor of that.

solarkennedy:
Discarding this one because we lose data about intent

PR #586, opened 2021-03-31
Added auto mkfs/fsck support to titus-storage

solarkennedy:
We knew we would need this eventually, so here it is!

Much of this code and logic is lifted from kubelet, which
should be a good indication that it is a solid approach.

Tested using the titus functional tests on a fresh (unformatted)
ebs volume.

### DIFF
--- a/cmd/titus-storage/ebs.go
+++ b/cmd/titus-storage/ebs.go
@@ -71,16 +71,47 @@ func ebsStart(ctx context.Context, ec2Client *ec2.EC2, c MountConfig, ec2Instanc
 		return err
 	}
 
-	err = mkfsIfNeeded(device, c.ebsFStype)
-	if err != nil {
-		return err
+	if c.ebsMountPerm == "RW" {
+		// Replicating kubelet behavior, we try to fsck before mount, but if there
+		// are errors we try to continue anyway
+		fsck(ctx, device, c.ebsFStype)
 	}
 
-	err = mountIt(ctx, device, c.ebsFStype, c.ebsMountPoint, c.ebsMountPerm, c.pid1Dir)
-	if err != nil {
-		return err
+	mc := MountCommand{
+		device:     device,
+		fstype:     c.ebsFStype,
+		mountPoint: c.ebsMountPoint,
+		perms:      c.ebsMountPerm,
+		pid1Dir:    c.pid1Dir,
 	}
-	l.Printf("finished setting up EBS %s at %s inside the container", c.ebsVolumeID, c.ebsMountPoint)
+	mountErr := mountBlockDeviceInContainer(ctx, mc)
+	if mountErr != nil {
+		// Replicating kubelet behavior, it is safer to just try to mount something,
+		// and only if it fails do we risk running mkfs and trying again.
+		mkfsIsNeeded, mkfsIsNeedederr := isMkfsNeeded(ctx, device, c.ebsFStype)
+		if mkfsIsNeedederr != nil {
+			return mkfsIsNeedederr
+		}
+		if mkfsIsNeeded {
+			if c.ebsMountPerm != "RW" {
+				return fmt.Errorf("A mkfs is needed on this device, but the mount permissions were not RW, they were %q", c.ebsMountPerm)
+			}
+			l.Infof("mkfs required on %s", device)
+			mkfsErr := mkfs(ctx, device, c.ebsFStype)
+			if mkfsErr == nil {
+				l.Infof("mkfs finished on %s, trying to mount again", device)
+				secondMountErr := mountBlockDeviceInContainer(ctx, mc)
+				if secondMountErr != nil {
+					return fmt.Errorf("Failed to mount a second time, even after a mkfs: %w", secondMountErr)
+				}
+				l.Infof("finished setting up EBS %s at %s inside the container after mkfs", c.ebsVolumeID, c.ebsMountPoint)
+				return nil
+			}
+			return mkfsErr
+		}
+		return mountErr
+	}
+	l.Infof("finished setting up EBS %s at %s inside the container", c.ebsVolumeID, c.ebsMountPoint)
 	return nil
 }
 
--- a/cmd/titus-storage/fs.go
+++ b/cmd/titus-storage/fs.go
@@ -1,0 +1,123 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"os/exec"
+	"strings"
+
+	"github.com/Netflix/titus-executor/logger"
+)
+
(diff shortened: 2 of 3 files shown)


cmd/titus-storage/fs.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+func mkfs(ctx context.Context, device string, fstype string) error {
+	if fstype == "ext4" || fstype == "ext3" || fstype == "" {
+		args := []string{
+			"-F",  // Force flag
+			"-m0", // Zero blocks reserved for super-user
+			device,
+		}
+		return runMkfs(ctx, fstype, device, args)
```
sargun, line 32: What happens when fstype == "", wont it run "mkfs."

cmd/titus-storage/fs.go
```diff
@@ -0,0 +1,121 @@
...
+func runMkfs(ctx context.Context, device string, fstype string, args []string) error {
+	l := logger.GetLogger(ctx)
+	mkfsCmd := "mkfs." + fstype
+	out, err := exec.Command(mkfsCmd, args...).CombinedOutput()
+	if err == nil {
+		// the disk has been formatted successfully try to mount it again.
+		l.Infof("Disk successfully formatted (mkfs): %s - %s", fstype, device)
+		return nil
+	}
+	l.Errorf("format of disk %s failed: %s - %s)", device, err, string(out))
```
sargun, line 53: l.WithError(err)...

cmd/titus-storage/fs.go
```diff
@@ -0,0 +1,121 @@
...
+
+// getDiskFormat is copied from kubelet. Returns empty string if there is no filesystem
+// in place, indicating it is safe to mkfs
+func getDiskFormat(ctx context.Context, disk string) (string, error) {
+	l := logger.GetLogger(ctx)
+	args := []string{"-p", "-s", "TYPE", "-s", "PTTYPE", "-o", "export", disk}
+	l.Infof("Attempting to determine if disk %q is formatted using blkid with args: (%v)", disk, args)
+	dataOut, err := exec.Command("blkid", args...).CombinedOutput()
+	output := string(dataOut)
+	l.Infof("blkid Output: %q, exit code: %v", output, err)
```
sargun, line 65: witherr...

cmd/titus-storage/fs.go
```diff
@@ -0,0 +1,121 @@
...
+				return "", nil
+			}
+		}
+		l.Errorf("Could not determine if disk %q is formatted (%v)", disk, err)
+		return "", err
+	}
+
+	var fstype, pttype string
+
+	lines := strings.Split(output, "\n")
```
sargun, line 83: Can you add a test / break this out and have some canned blkid junk to test against?

cmd/titus-storage/ebs.go
```diff
@@ -66,16 +66,36 @@ func ebsStart(ctx context.Context, ec2Client *ec2.EC2, c MountConfig, ec2Instanc
...
+	if c.ebsMountPerm == "RW" {
+		// Replicating kubelet behavior, we try to fsck before mount, but if there
+		// are errors we try to continue anyway
+		fsck(ctx, device, c.ebsFStype)
 	}
 
-	err = mountIt(ctx, device, c.ebsFStype, c.ebsMountPoint, c.ebsMountPerm, c.pid1Dir)
-	if err != nil {
-		return err
+	mountErr := mountIt(ctx, device, c.ebsFStype, c.ebsMountPoint, c.ebsMountPerm, c.pid1Dir)
```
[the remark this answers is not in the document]
solarkennedy: good idea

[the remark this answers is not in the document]
solarkennedy: Eh, I want to leave this function "ebs unaware" and leave it just dealing with block devices.

I'll still use a struct through to make sure we get all the strings right.

cmd/titus-storage/ebs.go
```diff
@@ -66,16 +66,36 @@ func ebsStart(ctx context.Context, ec2Client *ec2.EC2, c MountConfig, ec2Instanc
...
+				l.Infof("mkfs finished on %s, trying to mount again", device)
+				secondMountErr := mountIt(ctx, device, c.ebsFStype, c.ebsMountPoint, c.ebsMountPerm, c.pid1Dir)
+				return secondMountErr
+			}
+			return mkfsErr
+		}
+		return mountErr
 	}
-	l.Printf("finished setting up EBS %s at %s inside the container", c.ebsVolumeID, c.ebsMountPoint)
+	l.Infof("finished setting up EBS %s at %s inside the container", c.ebsVolumeID, c.ebsMountPoint)
```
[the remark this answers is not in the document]
solarkennedy: This `l` already has taskid and ebs volume id, if you want I can omit them in the actual plain text field too. They are technically redundant but it reads a bit better when you are just reading the message

cmd/titus-storage/fs.go
solarkennedy: err..  good point...

cmd/titus-storage/fs.go
solarkennedy: Ehh, I would rather not.
My reasoning here is that because I copy/pasted it literally from the k8s codebase (only to avoid pulling in other objects, sometimes it is easier to just copy a function than to try to adopt all the runners around it, splitting the build into linux/darwin, etc), it should be pretty battle hardened.

If I do add tests, that means I need to refactor this function (the splitting it out), which could lead other different bugs, because that means I'll have to adjust this function in some way, and drift from kubelet.

I'm trying to have my cake (reusing k8s code) and eat it too (having a simpler build, without having to adopt k8s object factories and junk, only pulling in the functions I need without needing windows support and klog).

cmd/titus-storage/ebs.go
```diff
@@ -71,16 +71,47 @@ func ebsStart(ctx context.Context, ec2Client *ec2.EC2, c MountConfig, ec2Instanc
...
+		if mkfsIsNeedederr != nil {
+			return mkfsIsNeedederr
+		}
+		if mkfsIsNeeded {
+			if c.ebsMountPerm != "RW" {
+				return fmt.Errorf("A mkfs is needed on this device, but the mount permissions were not RW, they were %q", c.ebsMountPerm)
+			}
+			l.Infof("mkfs required on %s", device)
+			mkfsErr := mkfs(ctx, device, c.ebsFStype)
+			if mkfsErr == nil {
```
rgulewich, line 101: This is a minor nit, but I always do a double take when code reverses the idiomatic go convention of `if err != nil {}` and instead branches based on the error being present.

rgulewich (approved):

PR #589, opened 2021-04-02
Create seccomp policy files from JSON patches/diffs

solarkennedy:
This creates new Makefile helpers and jsonnet to make it
easier to encode human intent into our seccomp policies.

The goal is to have the machine-readable json forms, but also
have human-readable patches, which have the diff compared
to the upstream docker version of their seccomp policy.

This should allow us to "rebase" or policies and get updated
syscall references and best practices, while still retaining
our original intent with the local titus policy modifications.


solarkennedy:
Closing for now in favor of a different jsonnet-based approach

PR #591, opened 2021-04-06
Switch to the GCC image for building negative seccomp image

sargun:
The buildroot image if no longer available.


hack/test-images/ubuntu/Dockerfile
```diff
@@ -1,4 +1,5 @@
-FROM registry.svc.ci.openshift.org/coreos/cosa-buildroot as builder-negative-seccomp
+FROM gcc:10 as builder-negative-seccomp
+RUN mkdir /srv
```
solarkennedy, line 2: mkdir: cannot create directory '/srv': File exists

solarkennedy (approved):
fix and ship

PR #592, opened 2021-04-07
Remove custom sslTitusAgentCA logic from titus-vpc-service

solarkennedy:
This sslTitusAgentCA logic was a mistake, I didn't really
understand how mutualtls worked on grpc, and thought
we needed to do additional verification on the certificates
in this auth function.

This auth function used the 'web.crt', which happened to have
the intermediate CA, so it "worked". However, when that
intermediate CA rolled, there was mistmatch between the
client and server, so this verification function failed.

It failed, *even though* the golang tls stuff was doing its
job by verifying that the 'web.crt' was *actually valid* by
traversing all the way up to the normal SSL CA stuff.

This PR removes my not-needed verification in favor of
the normal stuff. A different PR (#427) will fix the todo
and add *authz* to complement the mtls authn.

### DIFF
--- a/cmd/titus-vpc-service/main.go
+++ b/cmd/titus-vpc-service/main.go
@@ -51,10 +51,9 @@ const (
 	maxConcurrentRequestsFlagName = "max-concurrent-requests"
 	workerRoleFlagName            = "worker-role"
 
-	sslCertFlagName         = "ssl-cert"
-	sslPrivateKeyFlagName   = "ssl-private-key"
-	sslCAFlagName           = "ssl-ca"
-	sslTitusAgentCAFlagName = "ssl-titusagent-ca"
+	sslCertFlagName       = "ssl-cert"
+	sslPrivateKeyFlagName = "ssl-private-key"
+	sslCAFlagName         = "ssl-ca"
 
 	enabledLongLivedTasksFlagName = "enabled-long-lived-tasks"
 	enabledTaskLoopsFlagName      = "enabled-task-loops"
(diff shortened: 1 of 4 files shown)


JackOfMostTrades (approved):

vpc/service/auth.go
```diff
@@ -39,30 +29,7 @@ type titusVPCAgentServiceAuthFuncOverride struct {
...
-		return ctx, errors.Wrap(err, "Cannot parse remote certificate")
-	}
-	_, err = cert.Verify(x509.VerifyOptions{
-		Roots:     vpcService.TitusAgentCACertPool,
-		KeyUsages: []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth},
-	})
-	if err != nil {
-		return nil, errors.Wrap(err, "Unable to verify client cert")
-	}
+	// TODO: Authn happens via the normal TLS verification, here is where
```
sargun, line 32: If we ever move "back" to a Titus Agent specific PKI, we will want to be able to add that root for authz.

cmd/titus-vpc-service/main.go
```diff
@@ -370,10 +358,6 @@ func bindVariables(v *pkgviper.Viper) {
 		panic(err)
 	}
 
-	if err := v.BindEnv(sslTitusAgentCAFlagName, "SSL_TITUSAGENT_CA"); err != nil {
```
sargun, line 373: How is the root cert that the client certs are signed with getting loaded into the grpc server?

sargun (approved):

cmd/titus-vpc-service/main.go
solarkennedy: That comes from SSL_CA, as part of the `tls_config`.

vpc/service/auth.go
solarkennedy: I don't think we need to move to Titus Agent specific PKI (not that we had that before anyway).
I think we just need:
* A good root of trust for authn (done)
* A regex to look at the CN to look for titus-agent for authz (in #427)

PR #593, opened 2021-04-07
Use a new cached cert provider so we don't load certs on every request

solarkennedy:
Use a cached TLS cert loader instead of loading cert on every request
    
It is not very performant to load our ssl certificates
on every request using the `GetCertificate` method.
    
This new wrapper is based of our internal Netflix version,
which reloads the cert once per hour.

I've applied it to the titus-vpc-service and logviewer.

### DIFF
--- a/utils/tls/tls.go
+++ b/utils/tls/tls.go
@@ -2,6 +2,7 @@ package tls
 
 import (
 	"crypto/tls"
+	"sync"
 	"time"
 )
 
@@ -10,6 +11,7 @@ type CachedCertificateLoader struct {
 	KeyPath     string
 	certificate *tls.Certificate
 	nextReload  time.Time
+	lock        sync.Mutex
 }
 
 // Get Certificate uses the cachedCertificateLoader struct
@@ -22,6 +24,8 @@ func (c *CachedCertificateLoader) GetCertificate(nowFunc func() time.Time) (*tls
 		now = nowFunc()
 	}
 	if now.After(c.nextReload) {
+		c.lock.Lock()
+		defer c.lock.Unlock()
 		cert, err := tls.LoadX509KeyPair(c.CertPath, c.KeyPath)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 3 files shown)


utils/tls/tls.go
```diff
@@ -0,0 +1,33 @@
...
+		now = time.Now()
+	} else {
+		now = nowFunc()
+	}
+	if now.After(c.nextReload) {
+		cert, err := tls.LoadX509KeyPair(c.CertPath, c.KeyPath)
+		if err != nil {
+			return nil, err
+		}
+		c.certificate = &cert
```
sargun, line 29: Don't you need to protect this with a lock?

sargun (changes requested):

utils/tls/tls.go
solarkennedy: Hmm. I copied it from our golang mtls code, I assume it is production ready.

In golang writing to maps is not lock-safe, but just setting a variable is right?

JackOfMostTrades: > In golang writing to maps is not lock-safe, but just setting a variable is right?

Yea, that's the case as I understand it. So there's a race here where _n_ goroutines getting a cert at the same time will all have their gorountines reading from disk, but AFAIK this just introduces extra overhead but wouldn't be buggy. You could do some locking if you want to avoid that.

JackOfMostTrades (approved):

utils/tls/tls.go
sargun: Unless the TLS library is doing locking on our behalf, I'm not sure it is. There's a very clear race condition though, where if two routines hit this at the same time, they will both do a read-modify-write on the variable, and that will trigger the race detector.

solarkennedy: Ok, I think I've added a mutex to lock the loading/setting of the cert, but only if it is time to do so.

rgulewich (approved):

PR #594, opened 2021-04-09
Added a new titus-mount-bind command

solarkennedy:
In the same family as the other titus-mount commands,
but for bind mounts.

In the future I want to be able to use this to generate
on-the-fly bind mounts on *running* containers.

This will be useful to do for debug mode, but also
for dynamic mounts from the titus-storage command for
ephemeral drives.

I was sad that I couldn't use the newer syscalls (fsmount),
but they don't seem ready to have a "superblock" for
a bind mount to use.

See the comments in the approach for rational.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -85,16 +85,18 @@ checkstyle-result.xml
 /root/apps/titus-executor/bin/titus-reaper
 /root/apps/titus-executor/bin/titus-logviewer
 /root/apps/titus-executor/bin/titus-metadata-service
-/root/apps/titus-executor/bin/titus-mount-nfs
+/root/apps/titus-executor/bin/titus-mount-bind
 /root/apps/titus-executor/bin/titus-mount-block-device
+/root/apps/titus-executor/bin/titus-mount-nfs
 /root/apps/titus-executor/bin/tini-static
 /root/apps/titus-executor/bin/titus-inject-metadataproxy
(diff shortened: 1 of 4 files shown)


sargun:
Can't you do open_tree (maybe you need OPEN_TREE_CLONE?), and move_mount?

solarkennedy:
> Can't you do open_tree (maybe you need OPEN_TREE_CLONE?), and move_mount?

From reading https://lwn.net/Articles/809125/ 
it looked like that method was going to be a non-starter if I couldn't do things as read-only.

solarkennedy:
> > Can't you do open_tree (maybe you need OPEN_TREE_CLONE?), and move_mount?
> 
> From reading https://lwn.net/Articles/809125/
> it looked like that method was going to be a non-starter if I couldn't do things as read-only.

Maybe there is another way:
https://yhbt.net/lore/all/20210306101010.243666-1-christian.brauner@ubuntu.com/T/
Looks like an example of some trickery

Looks like __NR_open_tree is the syscall to use?

sargun (approved):
Should we wait until 5.12 gets released for idmapped mounts?

solarkennedy:
> Should we wait until 5.12 gets released for idmapped mounts?

Heh no :)

I made #595 to not forget to use the new apis when they are available.

PR #596, opened 2021-04-13
Delegate cgroups to titus-executor

sargun:


solarkennedy:
Reading
https://www.freedesktop.org/software/systemd/man/systemd.resource-control.html#Delegate=

I guess we didn't need this before, because we were always using docker to make the containers / cgroups for us.

Why do we need this now?

sargun:
Technically we've needed it for a while. It's to tell systemd to also expect cgroups under th3 unit and handle them correctly

solarkennedy (approved):

PR #597, opened 2021-04-13
User systemd

sargun:
Move to systemd with user units
    
    This has the following problems:
     * Logging is "Broken" (in the sense that _SYSTEMD_USER_UNIT is not set)
     * All of the user units need to be turned into non-parameterized units
     * The debugging / introspection story sucks


solarkennedy:
> ```
> Move to systemd with user units
> 
> This has the following problems:
>  * Logging is "Broken" (in the sense that _SYSTEMD_USER_UNIT is not set)
Hmm. Can we hack/inject this somehow with.... wait what *is* `_SYSTEMD_USER_UNIT` ? 

>  * All of the user units need to be turned into non-parameterized units
That sounds OK I guess. Will there still be a way to distinguish the different logs streams/status/etc?
Maybe we could *force* `_SYSTEMD_USER_UNIT` to look like it was parameterized ?

>  * The debugging / introspection story sucks
> ```
Hmmmm.... Maybe we could write a `tsystemctl` or `tjournalctl` command that knows how to talk to dbus directly and do the needful?

sargun:
So, what *should* happen is in the log, the following trusted fields should be set:
_SYSTEMD_UNIT should be set to something like `titus-executor@abc.service`
and
_SYSTEMD_USER_UNIT should be something like `titus-metadata-proxy.service`

Unfortunately, for whatever reason, the latter isn't being set correctly.

solarkennedy:
@sargun it has been a while, and I think we've been ok without this? (I haven't seen symptoms of system falling over)

sargun:
Yeah, true.

PR #598, opened 2021-04-13
Adjust titus-ssh to be compatible with scratch containers

solarkennedy:
This allows titus-ssh to work on minimal containers
(tested with scratch & alpine).

And in general, will allow ssh to work on any wacky
container our users can come up with, and still
have "the titus experience" where they can ssh
to it and get at least some sort of shell.

### DIFF
--- a/hack/images/titus-sshd/run-titus-sshd
+++ b/hack/images/titus-sshd/run-titus-sshd
@@ -1,0 +1,55 @@
+#!/titus/sshd/bin/busybox sh
+
+# For security and portability we make sure we only use the busybox utilities
+# and make no assumptions about the path and what other binaries
+# might be deployed
+cat="/titus/sshd/bin/busybox cat"
+grep="/titus/sshd/bin/busybox grep"
+mkdir="/titus/sshd/bin/busybox mkdir"
+touch="/titus/sshd/bin/busybox touch"
+
+log() {
+  echo "titus-sshd: $*" >&2
+}
+
+write_etc_profile() {
+$cat > /etc/profile <<'EOF'
+if [ -d /etc/profile.d ]; then
+  for i in /etc/profile.d/*.sh; do
+    if [ -r $i ]; then
+      . $i
+    fi
+  done
+  unset i
+fi
+EOF
+}
+
+setup_files() {
+  # These are the bare minimum set of files that
+  # must exist for ssh to work
+  [ -f /etc/passwd ] || ($touch /etc/passwd && log "setting up /etc/passwd for ssh to work")
+  [ -d /var/log ] || ($mkdir -p /var/log && log "creating /var/log for ssh to work")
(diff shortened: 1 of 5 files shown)


hack/images/titus-sshd/run-titus-sshd
```diff
@@ -0,0 +1,41 @@
+#!/titus/sshd/bin/busybox sh
+
+write_etc_profile() {
+cat > /etc/profile <<'EOF'
```
sargun, line 4: Dumb question, but these are all binaries sourced from the prefix, and not from the user container, right?

hack/images/titus-sshd/run-titus-sshd
```diff
@@ -0,0 +1,41 @@
...
+  done
+  unset i
+fi
+EOF
+}
+
+setup_files() {
+  # These are the bare minimum set of files that
+  # must exist for ssh to work
+  [ -f /etc/passwd ] || touch /etc/passwd
```
sargun, line 19: We should probably log if we had to do this.

sargun (approved):

hack/images/titus-sshd/run-titus-sshd
solarkennedy: Great question. Let me test how the path is constructed here. 
Busybox will put its binaries in the end of the path, so I think yea, I should use full explicit paths so we never accidentally run user code here.

hack/images/titus-sshd/run-titus-sshd
solarkennedy: great idea

hack/images/titus-sshd/run-titus-sshd
rgulewich: `/etc` is getting created by the executor, so we don't need to `mkdir -p` it, correct?

rgulewich (approved):

hack/images/titus-sshd/run-titus-sshd
solarkennedy: correct

hack/images/titus-sshd/run-titus-sshd
```diff
@@ -0,0 +1,55 @@
+#!/titus/sshd/bin/busybox sh
```
alekar, line 1: Maybe shell check this file?

alekar (approved):

hack/images/titus-sshd/run-titus-sshd
solarkennedy: I'll get this on the next PR and see if I can grow it to include any script.
I think I'll have to adjust the builder docker image to get shellcheck in it

solarkennedy:
I need to deploy this a little carefully, because I need to ensure + bump the titus-ssh image *first* before upgrading the executor.

PR #600, opened 2021-04-16
Implemented first MVP of multi-container user workloads (pods)

solarkennedy:
This is an MVP of multi-container workloads on Titus (pods).
It uses the "Country & Western" approach, which treats the
first container as the "main" one, which holds all the namespaces.

The lifecycle of a pod in this MVP is:

1. Start up, but pause (via tini) the main container
2. Start up the Titus system services (Country style)
3. Start up the Other containers (Western style)
4. Unpause the main container.

The main advantage to this design is that when there is only
1 container, step 3 is a noop. This provides backwards
compatibility with our existing workloads while we iterate
on the nuances and functionality of step 3 (ordering, volumes,
healthchecks, logs, etc)

----

With this change, the `Runner` struct in the executor
now holds the whole pod object. This allows the container
creation to have all the information it needs to properly
launch (Run) the main workload *and* all the sidecars.

I think eventually we'll need to tease out the `container`
object from the `Runner`, separating the settings that
are for the whole *task* (network stuff, global params),
and then have an *array* of container objects, instead
of the singleton, but I would like to do that refactor
on another pass, for when we need more direct access
to the extra container objects.

### DIFF
--- a/executor/runner/runner.go
+++ b/executor/runner/runner.go
@@ -16,6 +16,7 @@ import (
 	"github.com/Netflix/titus-executor/uploader"
 	multierror "github.com/hashicorp/go-multierror"
 	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 )
 
 // Task contains all information for running a task
@@ -37,6 +38,7 @@ type Runner struct {
 	metricsTagger tagger // the presence of tagger indicates extra Atlas tag is enabled
 	runtime       runtimeTypes.Runtime
 	config        config.Config
+	pod           *corev1.Pod
 
 	container runtimeTypes.Container
 	watcher   *filesystems.Watcher
@@ -66,6 +68,7 @@ func StartTaskWithRuntime(ctx context.Context, task Task, m metrics.Reporter, rp
 	if err != nil {
 		return nil, err
 	}
+	pod := task.Pod
 	runner := &Runner{
 		metrics:       m,
 		metricsTagger: metricsTagger,
@@ -74,6 +77,7 @@ func StartTaskWithRuntime(ctx context.Context, task Task, m metrics.Reporter, rp
 		UpdatesChan:   make(chan Update, 10),
 		StoppedChan:   make(chan struct{}),
 		container:     container,
+		pod:           pod,
 	}
 
 	rt, err := rp(ctx, runner.container, startTime)
@@ -106,7 +110,7 @@ func (r *Runner) startRunner(ctx context.Context, startTime time.Time) {
 	defer close(r.StoppedChan)
 
 	updateChan := make(chan update, 10)
-	go r.runContainer(ctx, startTime, updateChan)
+	go r.runMainContainerAndStartSidecars(ctx, startTime, updateChan)
 
 	var lastUpdate *update
 	for update := range updateChan {
@@ -164,11 +168,24 @@ func (r *Runner) prepareContainer(ctx context.Context) update {
 		return update{status: titusdriver.Lost, msg: err.Error()}
 	}
 
-	return update{status: titusdriver.Starting, msg: "starting"}
+	containerNames := getContainerNames(r.pod)
+	startingMsg := fmt.Sprintf("starting up %d containers: %s", len(containerNames), containerNames)
+	return update{status: titusdriver.Starting, msg: startingMsg}
+}
+
+func getContainerNames(pod *v1.Pod) []string {
+	if pod == nil || pod.Spec.Containers == nil {
+		return []string{"main"}
+	}
+	names := []string{}
+	for _, c := range pod.Spec.Containers {
+		names = append(names, c.Name)
+	}
+	return names
 }
 
 // This is just splitting the "run" part of the of the runner
-func (r *Runner) runContainer(ctx context.Context, startTime time.Time, updateChan chan update) {
+func (r *Runner) runMainContainerAndStartSidecars(ctx context.Context, startTime time.Time, updateChan chan update) {
 	defer close(updateChan)
 	select {
 	case <-r.killChan:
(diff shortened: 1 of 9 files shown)


solarkennedy:
Added support for putting labels on the running extra containers so they don't get detected as "disconnected".

Also added support for pulling the image before running it.

executor/runtime/docker/docker.go
```diff
@@ -1427,6 +1436,98 @@ func (r *DockerRuntime) statusMonitor(cancel context.CancelFunc, c runtimeTypes.
...
+	if len(otherUserContainers) == 0 {
+		return nil
+	}
+	l := log.WithField("taskID", r.c.TaskID())
+	l.Infof("Starting up %d other user containers", len(otherUserContainers))
+	// Currently the order here is hard-coded to be whatever order it is in the podspec
+	// with no health or startup probes
+	for i, v1Container := range otherUserContainers {
+		l.Debugf("Starting other user container %d/%d: %s", i+1, len(otherUserContainers), v1Container.Name)
+		err := pullWithRetries(ctx, r.metrics, r.client, v1Container.Image, doDockerPull)
```
alekar, line 1457: I think it's a better idea to pull all containers before we start them. Pulling here might cause unexpectedly long delays in startup.

executor/runtime/docker/docker.go
```diff
@@ -1427,6 +1436,98 @@ func (r *DockerRuntime) statusMonitor(cancel context.CancelFunc, c runtimeTypes.
...
+
+func (r *DockerRuntime) startOtherUserContainer(ctx context.Context, v1Container v1.Container, mainContainerID string) error {
+	l := log.WithField("taskID", r.c.TaskID())
+	containerName := r.c.TaskID() + v1Container.Name
+	dockerContainerConfig, dockerHostConfig, dockerNetworkConfig := k8sContainerToDockerConfigs(v1Container, mainContainerID)
+	l.WithFields(map[string]interface{}{
+		"dockerCfg": logger.ShouldJSON(ctx, *dockerContainerConfig),
+		"hostCfg":   logger.ShouldJSON(ctx, *dockerHostConfig),
+	}).Info("Creating other user container in docker")
+	containerCreateBody, err := r.client.ContainerCreate(ctx, dockerContainerConfig, dockerHostConfig, dockerNetworkConfig, containerName)
```
alekar, line 1477: Same comment as above. We can pull and create these containers in parallel (alongside/ like sidecars) then launch them here.

executor/runner/runner.go
```diff
@@ -232,7 +249,7 @@ func (r *Runner) maybeSetDefaultTags(ctx context.Context) {
 	}
 }
 
-func (r *Runner) monitorContainer(ctx context.Context, startTime time.Time, statusChan <-chan runtimeTypes.StatusMessage, updateChan chan update, details *runtimeTypes.Details) { // nolint: gocyclo
+func (r *Runner) monitorMainContainer(ctx context.Context, startTime time.Time, statusChan <-chan runtimeTypes.StatusMessage, updateChan chan update, details *runtimeTypes.Details) { // nolint: gocyclo
```
rgulewich, line 252: Good rename - thanks. :)

executor/runtime/docker/docker.go
```diff
@@ -1427,6 +1436,98 @@ func (r *DockerRuntime) statusMonitor(cancel context.CancelFunc, c runtimeTypes.
...
 
+// startOtherUserContainers launches the other user containers, only looking
+// any container objects in the pod *after* the first one, converting the
+// v1.Container spec into something docker can understand, and then
+// running that container.
+func (r *DockerRuntime) startOtherUserContainers(ctx context.Context, pod *v1.Pod, mainContainerID string) error {
+	if pod == nil || pod.Spec.Containers == nil {
+		return nil
+	}
+	otherUserContainers := pod.Spec.Containers[1:]
```
rgulewich, line 1447: I guess this is temporary until we pass in more information about which containers are "user" vs "sidecar", right?  It's ok for now, but as soon as we actually support this in the API, having the first container be the main one won't be something we can rely on.

executor/runtime/docker/docker.go
rgulewich: Hmm, yeah - after thinking about it a bit, that seems reasonable.  Starting shouldn't take long (since it's just launching tini at this point), so pulling images is going to be the long pole here.  The image pull seems like the most likely candidate for things to go wrong, so might as well fail up front before we start even one container.

executor/runtime/docker/docker.go
solarkennedy: Yes

executor/runtime/docker/docker.go
solarkennedy: Hmm. I'll see if I can parallelize this without loss of logging fidelity.

executor/runtime/docker/docker.go
solarkennedy: Yep. When your pod stuff hits in the API, we can look at the schema and not assume the first container is the main one.
(or something like that).

But all the other logic about pulling, creating will be the same. (should be)

*Then* lastly, when it comes to startup+healthchecks, that is where we'll care about which are user workloads vs sidecars, and we can be more precise about startup order.

rgulewich: It also occurs to me that you'll need to make sure that the pod-backed `Container` implementation (in `executor/runtime/types/pod.go` also picks the right user container. That code currently uses [GetUserContainer() from titus-kube-common](https://github.com/Netflix/titus-kube-common/blob/c6a533ee35ad49e3157579fe73197f4cc1639030/pod/containers.go#L7).

rgulewich (approved):

executor/runtime/docker/docker.go
solarkennedy: Wait, I thought you already did that? I rebased on master on your changes and I think that is in.

solarkennedy:
Rebased on master to get docker hub pull fixes.

PR #601, opened 2021-04-16
Ensure tini is always available to every titus container

solarkennedy:
I want to make sure that we can get "the titus experience",
especially when it comes to multi-container workloads (pods),
*even* on a mac laptop.

This means making sure we always have tini. Once we can safely
assume we always have tini, then we can remove remove codepaths
which cannot assume it to be there.

To do this, I bind mount tini from the host, instead of depending
on --init-path being set on the docker daemon. This works even if it
*is* set.

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -107,11 +107,11 @@ type DockerRuntime struct { // nolint: golint
 	client            *docker.Client
 	awsRegion         string
 	tiniSocketDir     string
-	tiniEnabled       bool
 	storageOptEnabled bool
 	pidCgroupPath     string
 	cfg               config.Config
 	dockerCfg         Config
+	defaultBindMounts []string
 
 	// cleanup callbacks that runtime implementations can register to do cleanup
 	cleanupFuncLock sync.Mutex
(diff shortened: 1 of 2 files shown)


executor/runtime/docker/docker.go
```diff
@@ -147,6 +147,10 @@ func NewDockerRuntime(ctx context.Context, m metrics.Reporter, dockerCfg Config,
 		return nil, err
 	}
 
+	// We bind-mount tini in as /sbin/docker-init to ensure we can always
+	// depend on it being there, regardless of the host docker configuration.
+	defaultBindMounts := []string{dockerCfg.tiniPath + ":/sbin/docker-init:ro"}
```
alekar, line 152: This should be mounted at `/dev/init`, no? Do you plan to still use docker to do the entry point rewriting? If you do, that is equivalent of forcing the flag on all the time, which might be easier?

solarkennedy: Older docker version use `/dev/init`, but newer ones (my laptop / devstack) use the new path.
It won't hurt on older versions.

I don't plan on doing the rewriting, and we don't need to with the `--init` flag (it does the rewrite for us).

alekar: Cool, this commit description confused me.

```
To do this, I bind mount tini from the host, instead of depending
on --init being set on the docker daemon. This works even if it
is set.```

alekar (approved):

executor/runtime/docker/docker.go
solarkennedy: Oh, yea I guess there are *two* flags here we are confusing.

One is `--init-path` on the docker daemon. This PR is to make it so we don't depend on that.

The second is `--init` on docker run. This bind mounts in (something) and rewrites entrypoint. We *do* still need this. Just with my PR, my bind mount wins and get used, regardless of `--init-path`.

rgulewich (approved):

PR #602, opened 2021-04-19
Added beta ephemeral storage support via titus-storage for GPU tasks

solarkennedy:
This adds Beta support for having a /ephemeral directory
mounted inside the container that represents scratch disk.

This is most often requested for GPU-based training workloads
(also the only place were we have ephemeral disks).

This is a simple implementation at first, with no regards
for quota or protection, just minimal isolation. This
feature will be considered "best effort" for our users.


cmd/titus-storage/ephemeral.go
```diff
@@ -0,0 +1,78 @@
...
+	ephemeralStorageRoot = "/ephemeral"
+	mountPathInContainer = "/ephemeral"
+)
+
+func ephemeralStorageIsAvailable() bool {
+	_, err := os.Stat(ephemeralStorageRoot)
+	if err == nil {
+		return true
+	}
+	if os.IsNotExist(err) {
```
sargun, line 22: why is this a special case?

cmd/titus-storage/ephemeral.go
```diff
@@ -0,0 +1,78 @@
...
+func ephemeralStorageStart(ctx context.Context, config MountConfig) error {
+	l := logger.GetLogger(ctx)
+	taskEphemeralStorageDir := getTaskEphemeralStorageDir(config.taskID)
+	err := os.MkdirAll(taskEphemeralStorageDir, os.FileMode(0777))
+	if err != nil {
+		return fmt.Errorf("Couldn't create directory %s for ephemeral storage: %w", taskEphemeralStorageDir, err)
+	}
+	// MkdirAll isn't guaranteed to actually chmod for you, so you must
+	// followup with your own chmod
+	_ = os.Chmod(taskEphemeralStorageDir, 0777)
```
sargun, line 59: Why do we ignore errors here?

cmd/titus-storage/ephemeral.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+	l.Infof("Mounted ephemeral storage %s into %s in the container", taskEphemeralStorageDir, mountPathInContainer)
+	return nil
+}
+
+func ephemeralStorageStop(ctx context.Context, config MountConfig) error {
+	l := logger.GetLogger(ctx)
+	taskEphemeralStorageDir := getTaskEphemeralStorageDir(config.taskID)
+	l.Infof("Cleaning up ephemeral storage %s", taskEphemeralStorageDir)
+	err := os.RemoveAll(taskEphemeralStorageDir)
```
sargun, line 72: return os.RemoveAll..

cmd/titus-storage/main.go
```diff
@@ -50,11 +50,21 @@ func main() {
 			l := logger.GetLogger(ctx)
 			command := args[0]
 			l.Infof("Running titus-storage with %s", command)
-			err := ebsRunner(ctx, command, mountConfig)
-			if err != nil {
-				l.Error(err)
+			if mountConfig.ebsVolumeID != "" {
+				err := ebsRunner(ctx, command, mountConfig)
+				if err != nil {
+					l.Error(err)
```
sargun, line 56: l.withError(err).Error()

cmd/titus-storage/main.go
```diff
@@ -50,11 +50,21 @@ func main() {
...
+					return err
+				}
 			}
-			return err
+			if ephemeralStorageIsAvailable() {
+				err := ephemeralStorageRunner(ctx, command, mountConfig)
+				if err != nil {
+					l.Error(err)
```
sargun, line 63: see above.

cmd/titus-storage/mount.go
```diff
@@ -44,10 +48,88 @@ func mountIt(ctx context.Context, dev string, fstype string, mountPoint string,
 	}
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
-	l.Printf("%s %s", strings.Join(cmd.Env, " "), mountCommand)
+	l.Printf("%s %s", strings.Join(cmd.Env, " "), mountBlockDeviceCommand)
 	return cmd.Run()
 }
 
+func bindMountInContainer(ctx context.Context, source string, target string, mountPerm string, pid1Dir string) error {
```
sargun, line 55: I strongly recommend you make this a struct because it'll be harder to keep track of args.

sargun (approved):

cmd/titus-storage/mount.go
solarkennedy: Great idea. Let me get https://github.com/Netflix/titus-executor/pull/586 merged and then I'll rebase and use the struct defined there.

rgulewich (approved):

rgulewich:
I freaked out a little at first that there were no quotas, but since this is a beta / best-effort thing, I managed to talk myself back down.  What's the behaviour going to be if more than one task on a GPU instance tries to use `/ephemeral`, @solarkennedy?  They could potentially stomp on each other's files and run each other out of disk space, right?  The second is ok, I think, but I'm wondering if we should actually do a mkdir of `/ephemeral/<task ID>` on the host, and mount *that* as `/ephemeral` in the container, to prevent users from accidentally stomping on each other's files.

If this problem is fixed by v2 / quota support, ignore me. :)

solarkennedy:
> I freaked out a little at first that there were no quotas, but since this is a beta / best-effort thing, I managed to talk myself back down. What's the behaviour going to be if more than one task on a GPU instance tries to use `/ephemeral`, @solarkennedy? They could potentially stomp on each other's files and run each other out of disk space, right? The second is ok, I think, but I'm wondering if we should actually do a mkdir of `/ephemeral/<task ID>` on the host, and mount _that_ as `/ephemeral` in the container, to prevent users from accidentally stomping on each other's files.
> 
> If this problem is fixed by v2 / quota support, ignore me. :)

Yep, the actual dir is

```return path.Join(ephemeralStorageRoot, "titus-task-storage", taskID)```

No stomping will be allowed. You are right though that one can use more than their fair share.


But, I was thinking the other day about how v2 might be implemented, and I actually don't think it will be harder then v1, is that I could do LVM. (on top of raid).

The upside to LVM is that it is super and instant to delete the lvm volume (no removing user files, which is dangerous), and LVM comes with free quota. I would give users `N gpus requested / VG total space`, where if you ask for 1/8 gpus, you get 1/8 of the VG to use. 

A little more restricting, but more fair and probably safer.
It is more complex, and a little harder to "see" what is happening (you can't just `ls /ephemeral`), but I think it is way worth it. 

Because this PR isn't even used yet, I think I'll actual go right for the v2 with LVM if you all agree.

rgulewich:
> Yep, the actual dir is
> 
> `return path.Join(ephemeralStorageRoot, "titus-task-storage", taskID)`

Ah, yeah - totally missed that.  Sorry.

> The upside to LVM is that it is super and instant to delete the lvm volume (no removing user files, which is dangerous), and LVM comes with free quota. I would give users `N gpus requested / VG total space`, where if you ask for 1/8 gpus, you get 1/8 of the VG to use.
> ... 
> Because this PR isn't even used yet, I think I'll actual go right for the v2 with LVM if you all agree.

That seems ok to me, but I haven't used LVM in production before (only on personal servers) - I'd like to know if @sargun is aware of any sharp edges / dragons / landmines (pick whichever metaphor is appropriate).

cmd/titus-storage/ephemeral.go
solarkennedy: Yea doesn't need to be

solarkennedy:
> > Yep, the actual dir is
> > `return path.Join(ephemeralStorageRoot, "titus-task-storage", taskID)`
> 
> Ah, yeah - totally missed that. Sorry.
> 
> > The upside to LVM is that it is super and instant to delete the lvm volume (no removing user files, which is dangerous), and LVM comes with free quota. I would give users `N gpus requested / VG total space`, where if you ask for 1/8 gpus, you get 1/8 of the VG to use.
> > ...
> > Because this PR isn't even used yet, I think I'll actual go right for the v2 with LVM if you all agree.
> 
> That seems ok to me, but I haven't used LVM in production before (only on personal servers) - I'd like to know if @sargun is aware of any sharp edges / dragons / landmines (pick whichever metaphor is appropriate).

Talking with sargun, it sounds like historically dynamic lvm like this is racy and can crash the kernel.
I'll stick with normal files for now, and if it ends up a bad idea, we can try lvm and see if it is any better, or we can try it with limited scope like on the p4s.

PR #604, opened 2021-04-20
Add injection code for IPv6 IMDS

sargun:


metadataserver/inject/inject_linux.go
```diff
@@ -0,0 +1,309 @@
...
+		file, err := listener.File()
+		if err != nil {
+			return fmt.Errorf("Could not get listener file: %w", err)
+		}
+
+		err = unix.Dup2(int(file.Fd()), 169)
+		if err != nil {
+			return fmt.Errorf("Could not clone FD %d into 169: %w", file.Fd(), err)
+		}*/
+
```
solarkennedy, line 85: Leftover?

metadataserver/inject/inject_linux.go
```diff
@@ -0,0 +1,309 @@
...
+	err = intermediateNSHandle.RouteAdd(&netlink.Route{
+		Dst:       localhostipnet,
+		LinkIndex: tocontainer.Attrs().Index,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("Could not add route to %v: %w", localhostipnet, err)
+	}
+
+	runtime.LockOSThread()
```
solarkennedy, line 253: Can you add a comment as to why this is so far down here? Is it only because you have to call it *right* before `netns.Set` ?

metadataserver/inject/inject_linux.go
```diff
@@ -0,0 +1,309 @@
...
+		return nil, fmt.Errorf("Could not set route_localnet in container NS: %w", err)
+	}
+
+	socket, err := unix.Socket(unix.AF_INET6, unix.SOCK_STREAM, 0)
+	/*
+		listener, err := net.ListenTCP("tcp", &net.TCPAddr{
+			Port: 80,
+			Zone: tocontainerInterfaceName,
+		}) // nolint: gosec
+	*/
```
solarkennedy, line 285: leftover? You have a different `Bind` later

metadataserver/inject/inject_linux.go
```diff
@@ -0,0 +1,309 @@
...
+	defer unix.Close(containerNSFD)
+	containerNS := netns.NsHandle(containerNSFD)
+
+	containerNSHandle, err := netlink.NewHandleAt(containerNS)
+	if err != nil {
+		return nil, fmt.Errorf("Could not get handle at intermediate ns: %w", err)
+	}
+	defer containerNSHandle.Delete()
+
+	err = intermediateNSHandle.LinkAdd(&netlink.Veth{
```
solarkennedy, line 138: Previously, in the C code, adding veth links (probably not this line of code) was done in a for loop 5 times, do you have more faith that in golang it will either work or not here?

(I may be commenting on the wrong line of code)

executor/standalone/standalone_test.go
```diff
@@ -696,11 +697,22 @@ func testMetadataProxyInjection(t *testing.T, jobID string) {
...
+	}
+	if !RunJobExpectingSuccess(t, ji) {
+		t.Fail()
+	}
+}
+func testMetadataProxyOnIPv6(t *testing.T, jobID string) {
+	ji := &JobInput{
+		ImageName:     ubuntu.name,
+		Version:       ubuntu.tag,
+		EntrypointOld: `/bin/bash -c 'curl -sf http://[fd00:ec2::254]/latest/meta-data/local-ipv4'`,
```
solarkennedy, line 715: Coool

solarkennedy: I look forward to when this test breaks in 2035 when we don't have `/local-ipv4` as route anymore, or it gives 404 or whatever :)

cmd/titus-inject-metadataproxy/main.go
```diff
@@ -0,0 +1,61 @@
...
+				return fmt.Errorf("Cannot parse log-level: %q", level)
+			}
+
+			return nil
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return inject.Inject(ctx, v.GetString("titus-pid-1-dir"), args)
+		},
+	}
+	rootCmd.Flags().String("titus-pid-1-dir", "", "Listening address")
```
solarkennedy, line 47: While it is nice that we have these two commands (the injector and the imds service itself) and it is nice to separate the concerns, now that they are in golang you *could* just have one command.

We would only need to do that if you thought it would actually simplify things, and not make things more complex. But lgtm like this.

solarkennedy (approved):
Excellent!

cmd/titus-inject-metadataproxy/main.go
sargun: We can't, because of threads we need to re-exec in order to complete the namespace transition.

metadataserver/inject/inject_linux.go
sargun: Yeah, need to remove the commented out code.

metadataserver/inject/inject_linux.go
rgulewich: The loop was to workaround a kernel race, which I haven't had time to track down. :(  So I'm guessing it could still happen.

PR #605, opened 2021-04-21
Added shellcheck capability to test suite

solarkennedy:
Probably going to have to fight CI a bit. Let's see...


alekar (approved):
That's great!

PR #606, opened 2021-04-21
Added titus-sshd to the CircleCI build steps

solarkennedy:
This makes it so we continuously build + push
the titus-sshd image, along with the other images
(builder, titus-logviewer, etc).

We haven't pushed a new image in 2 years it looks like:
https://hub.docker.com/r/titusoss/titus-sshd

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -132,6 +132,7 @@ workflows:
       - ci-builder
       - test-images
       - titus-logviewer
+      - titus-sshd
       - build:
           requires:
             - builder


sargun:
Can we do this only on master?

sargun:
Also, I see the build is failing. Do you want me to dive deeper into it?

solarkennedy:
> Also, I see the build is failing. Do you want me to dive deeper into it?

Let me give it a shot first. Not too surprising since it has been so long.

solarkennedy:
Closing, as there is no way this build is going to be something we can do in CI

PR #607, opened 2021-04-23
Set the apparmor profile if FUSE is requested

rgulewich:
This is feedback from the pod spec PR that I forgot to include: if FUSE is enabled, set the Apparmor profile correctly.  I've still allowed for overriding the profile with an attribute, but the plan is to not enable this in the TJC: this will be a feature only in `titus-standalone` (where I could see it being pretty handy).

### DIFF
--- a/executor/runtime/docker/capabilities.go
+++ b/executor/runtime/docker/capabilities.go
@@ -12,8 +12,9 @@ import (
 )
 
 const (
-	SYS_ADMIN = "SYS_ADMIN" // nolint: golint
-	NET_ADMIN = "NET_ADMIN" // nolint: golint
+	SYS_ADMIN              = "SYS_ADMIN" // nolint: golint
+	NET_ADMIN              = "NET_ADMIN" // nolint: golint
+	defaultApparmorProfile = "docker_titus"
 )
 
 func addAdditionalCapabilities(c runtimeTypes.Container, hostCfg *container.HostConfig) map[string]struct{} {
@@ -38,7 +39,7 @@ func addAdditionalCapabilities(c runtimeTypes.Container, hostCfg *container.Host
 func setupAdditionalCapabilities(c runtimeTypes.Container, hostCfg *container.HostConfig) error {
 	addedCapabilities := addAdditionalCapabilities(c, hostCfg)
 	seccompProfile := "default.json"
-	apparmorProfile := "docker_titus"
+	apparmorProfile := defaultApparmorProfile
 
 	if aProf := c.AppArmorProfile(); aProf != nil {
 		apparmorProfile = *aProf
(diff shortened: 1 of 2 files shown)


sargun (approved):

executor/runtime/docker/capabilities.go
```diff
@@ -54,6 +55,10 @@ func setupAdditionalCapabilities(c runtimeTypes.Container, hostCfg *container.Ho
 			PathInContainer:   fuseDev,
 			CgroupPermissions: "rmw",
 		})
+
+		if apparmorProfile == defaultApparmorProfile {
```
alekar, line 59: Is this if redundant?

rgulewich: If `c.AppArmorProfile()` is non-nil ([L44](https://github.com/Netflix/titus-executor/pull/607/files#diff-eda016907a75ec19335cdc60f120452ec1f19a7ea4d5b717d0699b438be3936eR44)), then `apparmorProfile` will be set to something that's not the default.  I only want to set it back to `docker_fuse` if `apparmorProfile` is nil - in other words, the user didn't explicitly choose to override the profile.

solarkennedy: But if someone does manually set `docker_titus` (a non nil setting), this code will *still* override to `docker_fuse`.
I think that is OK though, this will be such a rare thing.
Heck if the logic is here in the executor, I dare say we shouldn't allow users to set AA profiles in the first place?

solarkennedy (approved):

solarkennedy:
```time=2021-04-23T00:44:17Z level=warning msg=Error pulling image 'docker.io/titusoss/alpine:3.5', due to reason: Error response from daemon: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit``` blarg

rgulewich:
> `time=2021-04-23T00:44:17Z level=warning msg=Error pulling image 'docker.io/titusoss/alpine:3.5', due to reason: Error response from daemon: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit` blarg

I've been wondering if we should be caching these images somehow, or maybe building them from scratch in the tests?  The downside is adding a bunch of complexity for transient errors...

executor/runtime/docker/capabilities.go
rgulewich: We don't allow setting AA profiles via the API.  I considered ripping this code out altogether, then realized that I would probably want to use this functionality when iterating on AA changes with `titus-standalone` (which is currently the only way you could set this annotation, other than creating a pod in apiserver directly).

alekar (approved):

PR #608, opened 2021-04-23
PS-3406: Update sshd config to point to metatron host SSH certs.

JackOfMostTrades:
### Description of the Change

This will cause sshd to present host SSH certificates if available. Note that OpenSSH will silently ignore the HostCertificate directives if the files don't exist, so this is a pretty safe change. This update is consistent with the nflx-sshd-config sshd_config.


solarkennedy:
Cool. Yea I think we are ready to try this again!

PR #610, opened 2021-04-24
RFC: Use static busybox and static openssh binaries for titus-sshd

solarkennedy:
This is a radical departure from the way we build titus-sshd
today, but is faster and hopefully more sustainable?

It uses the sshd binary from the minos distro, which has a
library of static binaries:

https://github.com/minos-org/morpheus-ports/tree/master/openssh

The downside is that this is a step backwards in versions
6.1 compared to what we currently run, 7.7. And the build is
more opaque.

On the upside, the image is much smaller, because things are
statically linked instead of just relocated.

### DIFF
--- a/executor/runtime/docker/docker_ssh.go
+++ b/executor/runtime/docker/docker_ssh.go
@@ -23,13 +23,6 @@ Protocol 2
 # HostKeys for protocol version 2
 HostKey /titus/sshd/etc/ssh/ssh_host_rsa_key
 HostCertificate /run/metatron/certificates/ssh_host_rsa_key-cert.pub
-HostKey /titus/sshd/etc/ssh/ssh_host_ecdsa_key
-HostCertificate /run/metatron/certificates/ssh_host_ecdsa_key-cert.pub
-HostKey /titus/sshd/etc/ssh/ssh_host_ed25519_key
-HostCertificate /run/metatron/certificates/ssh_host_ed25519_key-cert.pub
-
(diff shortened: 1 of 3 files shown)


JackOfMostTrades:
Hmm, I like the idea in general, but wow 6.1 is _really_ old. Released 2012-08-29. There's been a handful of [security notices](http://www.openssh.com/security.html) since then, though probably nothing vital reading through it. Of course, OpenSSH depends on OpenSSL, so we'd also have to comb through those security advisories to see if there would be anything impacting (and there of course have been high severity CVEs in there since 2012).

If we wanted to go down this path I'd be more inclined to grab a recent release and see if we can cobble together a statically linked binary ourselves (or one that at the very least statically links some of the dependencies that bloat the image).

solarkennedy:
> Hmm, I like the idea in general, but wow 6.1 is _really_ old. Released 2012-08-29. There's been a handful of [security notices](http://www.openssh.com/security.html) since then, though probably nothing vital reading through it. Of course, OpenSSH depends on OpenSSL, so we'd also have to comb through those security advisories to see if there would be anything impacting (and there of course have been high severity CVEs in there since 2012).
> 
> If we wanted to go down this path I'd be more inclined to grab a recent release and see if we can cobble together a statically linked binary ourselves (or one that at the very least statically links some of the dependencies that bloat the image).

I agree, this is pretty much a non-starter here in this PR.

I also don't see how this thing was actually built (not surprising given its age, probably built by hand?) So who knows.

PR #614, opened 2021-04-27
Add validator

sargun:


solarkennedy:
Did you mean to put both these commits in the same PR?

vpc/service/validator.go
```diff
@@ -19,9 +21,94 @@ func (vpcService *vpcService) ValidateAllocationParameters(ctx context.Context,
...
+		validationFailures = append(validationFailures, &titus.ParametersValidationResponse_ValidationFailure{
+			FailureOneOf: &titus.ParametersValidationResponse_ValidationFailure_AccountIdUnsupported{
+				AccountIdUnsupported: &titus.ParametersValidationResponse_AccountIdUnsupported{
+					AccountId: req.AccountId,
+				},
+			},
+		})
+	}
+
+	rows, err := tx.QueryContext(ctx, `
```
joshi-keyur, line 62: Check `req.Subnets` has one or more values to see if this validation step is required?

sargun: If subnets is empty doesn't that fail validation?

vpc/service/validator.go
```diff
@@ -19,9 +21,94 @@ func (vpcService *vpcService) ValidateAllocationParameters(ctx context.Context,
...
+	if err != nil {
+		err = tracehelpers.WithGRPCStatusCode(fmt.Errorf("Could not start database transaction: %w", err), codes.Unknown)
+		tracehelpers.SetStatus(err, span)
+		return nil, err
+	}
 
-	err := status.Error(codes.Unimplemented, "Call not yet implemented")
-	tracehelpers.SetStatus(err, span)
-	return nil, err
+	defer func() {
```
joshi-keyur, line 38: @sargun can you explain why we are rolling back? IIUC we are only doing reads currently, and if we ever change this function to modify the DB, this could become a problem?

sargun: You should always have a rollback statement to avoid leaking transactions. Commit (for writes) should be explicit.

vpc/service/validator.go
```diff
@@ -19,9 +21,94 @@ func (vpcService *vpcService) ValidateAllocationParameters(ctx context.Context,
 	defer span.End()
 	log := ctxlogrus.Extract(ctx)
 	ctx = logger.WithLogger(ctx, log)
-	_ = ctx
+	span.AddAttributes(
+		trace.StringAttribute("account", req.AccountId),
+		trace.StringAttribute("subnet-ids", fmt.Sprintf("%v", req.Subnets)),
+		trace.StringAttribute("security-groups", fmt.Sprintf("%v", req.Subnets)),
```
joshi-keyur, line 27: `req.SecurityGroups`?

PR #615, opened 2021-04-27
Setup docker auth for dind standalone tests

solarkennedy:
This enables basic docker registry authentication.

At first my goal is to make the tests less flakey.
Maybe someday we will use auth on our real registries too.

This also includes changes to make the dind tests run on mac.

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -566,7 +566,7 @@ func (r *DockerRuntime) DockerPull(ctx context.Context, c runtimeTypes.Container
 	r.metrics.Counter("titus.executor.dockerImagePulls", 1, nil)
 	logger.Infof("DockerPull: pulling image")
 	pullStartTime := time.Now()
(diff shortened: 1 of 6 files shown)


executor/runtime/docker/docker.go
```diff
@@ -1782,7 +1782,11 @@ func (r *DockerRuntime) setupPostStartLogDirTiniHandleConnection2(parentCtx cont
 
 	if r.dockerCfg.bumpTiniSchedPriority {
 		group.Go(func() error {
-			return setupScheduler(cred)
+			err := setupScheduler(cred)
```
rgulewich, line 1785: Whoa, was this why it didn't work on Mac?

rgulewich (approved):

executor/runtime/docker/docker.go
solarkennedy: Not super sure. `sched_getscheduler` should be able to be run with the unconfined mode. It is also an older `4.19.76-linuxkit` kernel. `CONFIG_RT_GROUP_SCHED=y` on mine.

sargun (approved):

PR #616, opened 2021-04-28
Add security group table

sargun:
Reconcile SGs periodically.


solarkennedy:
How were we getting new SGs before?

And after we ship this, there will probably be a lot of old SGs deleted the first time this runs?

solarkennedy (approved):

solarkennedy:
This must be a precursor to something else that uses this new table, right?

sargun:
This is a one way sync from AWS to Titus. It is there to power more validation logic.

PR #617, opened 2021-04-28
Added titus-network mode support & v6, v4, and TSA awareness

solarkennedy:
This new network-mode is designed to be a "source of truth"
for what kind of networking to setup on a container.

Specifically I want some way to know for sure if TSA and VPCservice
are supposed to setup a ipv4 transition mechanism.

OR, in the future case where we know we want ipv6 only, vpcservice
needs to know that and forgo certain things.

Likewise, titus-inject-metadata will probably have to behave
differently. And also I even expect user applications to take advantage
of the `TITUS_NETWORK_MODE=IPV6_ONLY` style hints, so that they
don't have to scrape ifconfig or other weirdness, and can adjust
their behavior appropriately.

This just plumbs the setting through, no additional code is in place
yet to actually act on it.


cmd/titus-vpc-tool/setup_container.go
```diff
@@ -33,6 +33,7 @@ func setupContainercommand(ctx context.Context, v *pkgviper.Viper, iipGetter ins
 	}
 
 	cmd.Flags().Int("netns", 3, "The File Descriptor # of the network namespace to setup")
+	cmd.Flags().Int("network-mode", 0, "Type of networking to setup, uses the enum")
```
sargun, line 36: :-/. Why not make this a string?

executor/runtime/types/container.go
```diff
@@ -1301,3 +1312,19 @@ func isEFSID(FsID string) (bool, error) {
 	}
 	return matched, nil
 }
+
+func getHumanFriendlyNetworkMode(mode int) string {
```
sargun, line 1316: iirc the protobuf definitions have a helper that does this translation.

executor/runtime/types/pod.go
```diff
@@ -394,6 +395,16 @@ func (c *PodContainer) MetatronCreds() *titus.ContainerInfo_MetatronCreds {
 	return c.titusInfo.GetMetatronCreds()
 }
 
+func (c *PodContainer) NetworkMode() int {
+	// TODO: move to pod-common when stable
+	modeStr := c.pod.Annotations["titusParameter.agent.networkMode"]
+	mode, err := strconv.ParseInt(modeStr, 10, 32)
```
sargun, line 401: IIRC there is a https://golang.org/pkg/strconv/#Itoa function (Itoa)

sargun (changes requested):

cmd/titus-vpc-tool/setup_container.go
solarkennedy: I can do that.

Ooo yea looks like in the .pb.go it has all the mappers I want to make this sane. Great.

executor/runtime/types/container.go
```diff
@@ -1318,3 +1330,25 @@ func isEFSID(FsID string) (bool, error) {
 	}
 	return matched, nil
 }
+
+// GetHumanFriendlyNetworkMode uses the incoming network mode string
+// and mutates it a bit to be a environment-variable safe string.
+// In the unknown mode, however, we return an empty string for
+// the caller to *not* set the variable
+func GetHumanFriendlyNetworkMode(mode string) string {
```
sargun, line 1338: In the proto object, there are two maps that have this info, I think?
```

var NetworkConfiguration_NetworkMode_name = map[int32]string{
        0: "UnknownNetworkMode",
        1: "Ipv4Only",
        2: "Ipv6AndIpv4",
        3: "Ipv6AndIpv4Fallback",
        4: "Ipv6Only",
}

var NetworkConfiguration_NetworkMode_value = map[string]int32{
        "UnknownNetworkMode":  0,
        "Ipv4Only":            1,
        "Ipv6AndIpv4":         2,
        "Ipv6AndIpv4Fallback": 3,
        "Ipv6Only":            4,
}

```

Why not use those names?

solarkennedy: I don't know if we want those internal names to be exposed to the user, maybe.

Here is the doc with the spec I'm conforming to:
https://docs.google.com/document/d/1IJxY4LaiVYy5RqXDRtYSscvysae4bB7-3v4SSqvdNbE/edit

vpc/tool/assign3/assign_network.go
```diff
@@ -32,10 +32,12 @@ type Arguments struct {
 	Jumbo              bool
 	Bandwidth          uint64
 	Burst              bool
+	NetworkMode        string
```
sargun, line 35: The titus executor and titus-vpc-tool are (effectively) atomically upgraded. So, you should get rid of the `AssignIPv6Address` logic here.

solarkennedy: I don't quite understand.
You want me to remove the `AssignIPv6Address` cli flag?

executor/runtime/types/container.go
```diff
@@ -129,6 +129,7 @@ type TitusInfoContainer struct {
 	resources Resources
 
 	// VPC driver fields
+	networkMode        string
 	assignIPv6Address  bool
```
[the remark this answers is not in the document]
solarkennedy: We can't get rid of it, this bool-as-attribute is still being set in many places.

I *could* "Fake" the networkMode setting:

```
if mode == Unset {
  if assignV6:
    mode = Ipv4AndV6
  }
}
```

But I don't really want to.
I would rather the networkMode stay untouched throughout

executor/runtime/types/container.go
solarkennedy: Hmm. Like with another enum?

vpc/tool/assign3/assign_network.go
sargun: Yeah, isn't `AssignIPv6Address` going to be subsumed by network mode?

executor/runtime/types/container.go
sargun: or a map of nice-names to less-nice-names (and vice-versa)?

vpc/tool/assign3/assign_network.go
solarkennedy: Eventually yes, but network mode will be Unknown (`0`) for a long time, we still need this bool to be in place to send in this data when network mode is 0.

executor/runtime/types/container.go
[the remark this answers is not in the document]
sargun: I only see it being written to in boolPassthroughs. I'm suggesting that we replace the function here:
https://github.com/Netflix/titus-executor/blob/ba7a401b6e6820a8d1d030a8a5be0826e0cc34b5/executor/runtime/types/container.go#L524

and check if networkMode is set, and use that *first*, and then we add a TODO to remove this once we deprecate the old label.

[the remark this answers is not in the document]
solarkennedy: Got it

solarkennedy:
@sargun redone. `EffectiveNetworkMode()` is now the source of truth for what kind of network mode we want, and that is the only thing vpctool gets.

`EffectiveNetworkMode()` is computed based off of the input network mode and the attributes for backwards compat.

executor/runtime/docker/docker.go
```diff
@@ -350,6 +351,25 @@ func maybeAddOptimisticDad(sysctl map[string]string) {
...
+	case titus.NetworkConfiguration_Ipv6AndIpv4Fallback.String():
+		// IPv6 with fallback means we only want v6 resolvers, which reduces
+		// the burden on TSA for ipv4 udp traffic
+		return []string{"fd00:ec2::253"}
+	case titus.NetworkConfiguration_Ipv6Only.String():
+		// True ipv6 only means we really can only have a v6
+		// resolver
+		return []string{"fd00:ec2::253"}
+	default:
+		// Any other situation means we can return the classic v4 resolver
```
sargun, line 368: I'd // TODO this as maybe defaulting to V6 later, and treating V4 only as a specific case

executor/runtime/docker/docker.go
```diff
@@ -987,7 +996,7 @@ func (r *DockerRuntime) Prepare(ctx context.Context) error { // nolint: gocyclo
 				"TITUS_SECCOMP_AGENT_HANDLE_PERF_SYSCALLS": "true",
 			})
 		}
-		if r.c.SeccompAgentEnabledForNetSyscalls() {
+		if r.c.SeccompAgentEnabledForNetSyscalls() || r.c.EffectiveNetworkMode() == titus.NetworkConfiguration_Ipv6AndIpv4Fallback.String() {
```
sargun, line 999: Long term, should SeccompAgentEnabledForNetSyscalls go away?

executor/runtime/types/container.go
```diff
@@ -219,6 +223,16 @@ func NewTitusInfoContainer(taskID string, titusInfo *titus.ContainerInfo, resour
 		c.extraUserContainers, c.extraPlatformContainers = NewExtraContainersFromPod(*c.pod)
 	}
 
+	pConfig := podCommon.Config{}
+	c.podConfig = &pConfig
```
sargun, line 227: Any reason to break this into two separate lines?

executor/runtime/types/container.go
```diff
@@ -810,6 +826,14 @@ func (c *TitusInfoContainer) MetatronCreds() *titus.ContainerInfo_MetatronCreds
 	return c.titusInfo.GetMetatronCreds()
 }
 
+func (c *TitusInfoContainer) EffectiveNetworkMode() string {
```
sargun, line 829: I'd just add a comment, even though there is a more detailed comment on `computeEffectiveNetworkMode` of what this means.

executor/runtime/types/pod.go
```diff
@@ -549,6 +557,9 @@ func (c *PodContainer) Runtime() string {
 	return DefaultOciRuntime
 }
 
+// SeccompAgentEnabledForNetSyscalls only represents whether a user set the SeccompAgentEnabledForNetSyscalls
```
sargun, line 560: If so, why would we leave this here and not always use network mode?

go.mod
```diff
@@ -28,16 +28,19 @@ require (
 	github.com/docker/go-connections v0.4.0 // indirect
 	github.com/docker/go-units v0.4.0
 	github.com/ftrvxmtrx/fd v0.0.0-20150925145434-c6d800382fff
+	github.com/gogo/protobuf v1.3.2 // indirect
```
sargun, line 31: What's pulling this in?

sargun (approved):

executor/runtime/docker/docker.go
solarkennedy: Yes, I just didn't want to break existing stuff.

Oh I see, no you are right, EffectiveNetworkMode() knows about that already.

executor/runtime/types/container.go
solarkennedy: nope

executor/runtime/types/pod.go
solarkennedy: I gotta wait till we are not using the attribute everywhere?

I still need *some* function in here to tell me if the attribute is set.
Once spinnaker and stuff can set network mode, I'll depreciate usage of this attribute asap and remove these things.

go.mod
solarkennedy: My best guess is `github.com/Netflix/titus-kube-common` bump

We don't seem to be actually using it in the titus-executor codebase directly.
go mod tidy doesn't make it go away

solarkennedy: Oh, k8s itself?
```kubernetes/go.mod:	github.com/gogo/protobuf v1.3.2```

PR #618, opened 2021-04-29
Refactor: Launch all (main and other sidecars) in together

solarkennedy:
This is 'just' a refactor to make it so I have a
single function responsible for starting all
containers, the main one (via) tini included.

This will allow us to (in future PRs) be able
to do full ordering amongst all defined containers,
where "main" is just a special case (launch via tini
instead of launching via `ContainerStart`).

My end goal here is to have a function that can take
the whole pod, and report back the ordering required,
and then my `startAllUserContainers` function will
be able to start things in the right order (including the
main one).

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -1232,26 +1232,22 @@ func maybeConvertIntoBadEntryPointError(err error) error {
 	return err
 }
 
-func (r *DockerRuntime) waitForTini(ctx context.Context, listener *net.UnixListener, c runtimeTypes.Container) (string, error) {
+// setupTini connects to tini, and also sets up NFS mounts
+// it does *not* send the L (launch) signal to tini though
+func (r *DockerRuntime) setupTini(ctx context.Context, listener *net.UnixListener, c runtimeTypes.Container) (string, *net.UnixConn, error) {
 	// This can block for up to the full ctx timeout
 	logDir, containerCred, rootFile, unixConn, err := r.setupPostStartLogDirTini(ctx, listener, c)
 	if err != nil {
-		return logDir, err
+		return logDir, unixConn, err
 	}
 
 	if len(c.NFSMounts()) > 0 {
 		err = r.setupEFSMounts(ctx, c, rootFile, containerCred)
 		if err != nil {
-			return logDir, err
+			return logDir, unixConn, err
 		}
 	}
-
-	err = launchTini(unixConn)
-	if err != nil {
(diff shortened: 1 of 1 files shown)


solarkennedy:
Ok ready for review @rgulewich / @alekar .

On the one hand, I was hoping to make it so the C&W code never gets exercised if you only have 1 container.
This is still mostly true in this PR, except for the actual Starting bit.

If we want ordered startup *and* live with one foot in the old world (C&W ethos), then we have to have the start function deal with the fact that starting the one main container means telling tini to launch.

Longer term, I think we'll probably use tini for all containers, because we'll want things like stdout/err control, seccomp stuff, so maybe it won't be so weird, because then the "startInOrder" function really sends `L` to all the right sockets in the right order?

executor/runtime/docker/docker.go
```diff
@@ -1361,48 +1358,50 @@ func (r *DockerRuntime) statusMonitor(cancel context.CancelFunc, c runtimeTypes.
...
-	if len(otherUserContainers) == 0 {
-		return nil
+// startUserContainers launches all user containers (including the main one)
+// for containers other than the main one, it must pull and create them
+func (r *DockerRuntime) startUserContainers(ctx context.Context, pod *v1.Pod, mainContainerID string, tiniConn *net.UnixConn) error {
+	var otherUserContainers []v1.Container
+	if pod != nil && pod.Spec.Containers != nil {
+		otherUserContainers = pod.Spec.Containers[1:]
+	} else {
+		otherUserContainers = []v1.Container{}
```
rgulewich, line 1376: This is super nit-picky, and not totally necessary, *but* I know you're a Python person, and a PEP 20 enthusiast. So, in the interests of "simple is better" and "flat is better than nested" - maybe initialize `otherUserContainers` to an empty array, and you can remove this else?

rgulewich (approved):

rgulewich:
> Longer term, I think we'll probably use tini for all containers

Yeah, that seems totally reasonable to me, for the reasons you listed.  That should probably be our default, I would think.

solarkennedy:
> > Longer term, I think we'll probably use tini for all containers
> 
> Yeah, that seems totally reasonable to me, for the reasons you listed. That should probably be our default, I would think.

I thought it was! Then I found:
https://github.com/moby/moby/blob/9f2b33f75cb094917c69e03ba101bc0bee8673fb/daemon/oci_linux.go#L737
And realized it was not actually getting injected.
In the /logs PR I inject it myself, for better or worse.

PR #619, opened 2021-04-29
Added new system-service: container-tools

solarkennedy:
This is phase 2 of my container tools work, in an effort
to make containers easier to debug on titus.

This moves the `/titus/container-tools/` folder away from
a host bind-mount into a real sidecar container.

This container/image has statically linked tools to make
it easier for our users to debug their containers without
having to mutate their own, or bring heavyweight tools
with them.

This new image is lightweight (static binaries from scratch),
so I don't anticipate a lot of extra load from docker.
Additionally there is no systemd unit to start, so no additional
load from systemd.

### DIFF
--- a/executor/runtime/types/sidecars.go
+++ b/executor/runtime/types/sidecars.go
@@ -15,6 +15,7 @@ const (
 	SidecarTitusStorage         = "titus-storage"
 	SidecarSeccompAgent         = "seccomp-agent"
 	SidecarServiceMetadataProxy = "metadata-proxy"
+	SidecarContainerTools       = "container-tools"
 )
 
 var sideCars = []ServiceOpts{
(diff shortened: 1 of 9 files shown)


executor/runtime/types/sidecars.go
```diff
@@ -183,6 +193,10 @@ func shouldStartTitusStorage(cfg *config.Config, c Container) bool {
 	return c.EBSInfo().VolumeID != ""
 }
 
+func shouldStartContainerTools(cfg *config.Config, c Container) bool {
+	return cfg.ContainerToolsImage != ""
```
rgulewich, line 197: The idea is that eventually every container will get this, correct?

solarkennedy: Yep.

solarkennedy: But I figured we'll leave this in here, in an emergency we just set the FP to `""` and no container tools

rgulewich: Makes sense, thanks!

rgulewich (approved):

PR #621, opened 2021-04-30
Added basic ordering (class type) in multi-container workloads

solarkennedy:
This is the first part of class-based sidecar ordering in
Titus. It does not do "dependsOn" yet, only does two-layer
ordering between platform-defined and user-defined containers.
    
Currently just starts all platform-containers in parallel,
and then user-defined ones in parallel. dependsOn will be complex
and added in the next pass of code.
    
For this one, I wanted to check to make sure I was going in the
right direction.

### DIFF
--- a/executor/standalone/standalone_test.go
+++ b/executor/standalone/standalone_test.go
@@ -1304,3 +1304,53 @@ func TestBasicMultiContainer(t *testing.T) {
 		t.Fail()
 	}
 }
+
+func TestMultiContainerDoesPlatformFirst(t *testing.T) {
+	wrapTestStandalone(t)
+	skipIfNotPod(t)
+
+	// And for the main container, we use pgrep to ensure that our *user* sentinel container
+	// is in fact running along side us.
+	// It will only work if the user sentinel sidecar is seen running by the time we start.
+	testEntrypointOld := "pgrep -fx '/bin/sleep 430'"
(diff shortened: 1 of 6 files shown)


solarkennedy:
Woot! Finally green.
Next thing I really should do is improve the experience when sidecars fail, because right now there is very little feedback to the user, which is why it took so long for me to figure this one out.

executor/standalone/standalone_test.go
```diff
@@ -1304,3 +1300,50 @@ func TestBasicMultiContainer(t *testing.T) {
...
+			// running, only then will it keep running with its own sleep
+			{
+				Name:    "user-sentinel",
+				Image:   busybox.name + `:` + busybox.tag,
+				Command: []string{"/bin/sh", "-c"},
+				Args:    []string{"pgrep -fx '/bin/sleep 420' && /bin/sleep 430"},
+			},
+		},
+		ExtraAnnotations: map[string]string{
+			"type.pod.netflix.com/platform-sentinel": "sidecar",
```
rgulewich, line 1342: Might be nice to directly use the annotation name from `podCommon`?

rgulewich (approved):

executor/standalone/standalone_test.go
solarkennedy: Oh yea

PR #622, opened 2021-04-30
Remove epoll and use clone(CLONE_FILES) to send notify_fd to TSA

MyUmmaGumma:
Replace the epoll and wait for the first sendmsg interception mechanism with the clone with CLONE_FILES to share the notify_fd between parent and child and sequence sending the notify_fd after its been created in the parent but before seccomp policies are inherited by the child.

### DIFF
--- a/tini/src/seccomp_fd_notify.c
+++ b/tini/src/seccomp_fd_notify.c
@@ -49,7 +49,6 @@ typedef struct tsa_fds {
 } tsa_fds;
 
 pthread_mutex_t wait_to_send = PTHREAD_MUTEX_INITIALIZER;
-//int send_notify_fd = 0;
 pthread_cond_t ready_to_send = PTHREAD_COND_INITIALIZER;
 
 bool have_cap_sysadmin()
@@ -237,6 +236,7 @@ static int child_send_fd(void *arg)
 void maybe_setup_seccomp_notifer() {
 	char *socket_path;
 	const int STACK_SIZE = 64 * 1024;
+	const int MAX_ATTEMPTS = 10;
 	char *stack;
 	char *stack_top;
 	int notify_fd = -1;
@@ -278,21 +278,20 @@ void maybe_setup_seccomp_notifer() {
 	int sock_fd = -1;
 	// Sometimes things are not perfect, and the socket is not ready at first
 	// Instead of enforcing strict ordering, we can be defensive and retry.
-	int attempts = 10;
-	for (int i = 1; i <= attempts; i++) {
+	for (int i = 1; i <= MAX_ATTEMPTS; i++) {
 		sock_fd = socket(AF_UNIX, SOCK_STREAM, 0);
 		if (sock_fd != -1) {
 			break;
 		}
 		PRINT_INFO(
 			"Titus seccomp unix socket not ready on attempt %d/%d, Sleeping 1 second",
-			i, attempts);
+			i, MAX_ATTEMPTS);
 		sleep(1);
 	}
 	if (sock_fd == -1) {
 		PRINT_WARNING(
 			"Unable to open unix socket for seccomp handoff after %d attempts: %s",
-			attempts, strerror(errno));
+			MAX_ATTEMPTS, strerror(errno));
 		kill(child_pid, SIGKILL);
 		return;
 	}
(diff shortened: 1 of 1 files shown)


tini/src/seccomp_fd_notify.c
```diff
@@ -39,8 +43,13 @@
 		fprintf(stderr, "\n");                                         \
 	}
 
+typedef struct tsa_fds {
```
sargun, line 46: Why a typedef and not a plain struct?

tini/src/seccomp_fd_notify.c
```diff
@@ -39,8 +43,13 @@
...
+	int fd_to_tsa;
+	int notify_fd;
+} tsa_fds;
+
 pthread_mutex_t wait_to_send = PTHREAD_MUTEX_INITIALIZER;
-int send_notify_fd = 0;
+//int send_notify_fd = 0;
```
sargun, line 52: You can just delete this line.

tini/src/seccomp_fd_notify.c
```diff
@@ -217,213 +226,130 @@ struct sock_filter net_perf_filter[] = {
...
+		PRINT_WARNING("Failure to setup child for clone");
+		return;
 	}
-
-	epfd = epoll_create(EPOLL_IGNORED_VAL);
-	if (epfd == -1) {
-		PRINT_WARNING("epoll_create error");
-		return NULL;
+	stack_top = stack + STACK_SIZE;
+	if (CHILD_SIG != 0 && CHILD_SIG != SIGCHLD) {
```
sargun, line 274: Why do we ignore that here? Or setup an ignore for the child signal?

tini/src/seccomp_fd_notify.c
```diff
@@ -217,213 +226,130 @@ struct sock_filter net_perf_filter[] = {
...
+	if (!socket_path) {
+		PRINT_WARNING("TITUS_SECCOMP_NOTIFY_SOCK_PATH not defined.");
+		kill(child_pid, SIGKILL);
+		return;
+	}
+	int sock_fd = -1;
+	// Sometimes things are not perfect, and the socket is not ready at first
+	// Instead of enforcing strict ordering, we can be defensive and retry.
+	int attempts = 10;
+	for (int i = 1; i <= attempts; i++) {
```
sargun, line 297: MAX_ATTEMPTS could be defined right?

tini/src/seccomp_fd_notify.c
```diff
@@ -217,213 +226,130 @@ struct sock_filter net_perf_filter[] = {
...
-		notify_fd = install_notify_filter();
-		if (notify_fd == -1) {
-			close(sock_fd);
-			return;
+	struct sockaddr_un addr = { 0 };
+	addr.sun_family = AF_UNIX;
+	strncpy(addr.sun_path, socket_path, sizeof(addr.sun_path) - 1);
+	int result = -1;
+	for (int i = 1; i <= attempts; i++) {
+		result = connect(sock_fd, (struct sockaddr *)&addr,
```
sargun, line 320: This might be a dumb question, but why isn't this done in the child?

tini/src/seccomp_fd_notify.c
MyUmmaGumma: just to avoid writing the `struct` in `struct tsa_fds`

tini/src/seccomp_fd_notify.c
MyUmmaGumma: The intention is to ignore everything that isn't a SIGCHILD as the other signals might be to terminate the process. We don't ignore SIGCHILD so that we can wait in the parent.

tini/src/seccomp_fd_notify.c
MyUmmaGumma: Ah, it could be  - just wrote the code out trying to isolate the child to do the single task of just sending over the fd (or call sendmsg) only and the parent setting up the `fds` in the  `struct tsa_fds`.

sargun:
LGTM. The one thing you might want to look at is pr_set_pdeathsig in order to make sure that the child terminates with the parent.

MyUmmaGumma:
Based on @sargun's suggestion I added the PDEATH_SIG to handle the case when the parent might exit before the child and to prevent the child from becoming a zombie process.

PR #623, opened 2021-05-01
Use LVM for ephemeral storage

solarkennedy:
Added LVM-based ephemeral storage
  
This is a redo of the bind-mount based ephemeral storage, but with LVM

* Safer in the sense that we never see the user's files or have to clean them up
* "Quota" is enabled, they only get the slice of the ephemeral drives that
  proportional to the number of gpus they asked for
* No hacky bind mount injection thing ("only" the standard block device injection like ebs)

It does assume that all titus agents have a LVM vg in place called "ephemeral" (done).

All this ephemeral storage stuff is *best effort* and will not block a task
from launching.

----

In two commits, the first one is not interesting and just me removing old code (bind mount stuff)
https://github.com/Netflix/titus-executor/pull/623/commits/96708e238568b0bdf385d3a73c131bcf23ef3d9d
has the lvm parts.


solarkennedy:
Woot, green build on standalone. Some day I would like to make it easier to get standalone tests easy to run on mac. :(

Ready for real review.

sargun (approved):

PR #624, opened 2021-05-03
Misc logging and status updates

solarkennedy:
When trying to debug why a container didn't start
correctly, I took it as an opportunity to do a
refresh of all the (semi) cryptic `creating` and
`got msg` log messages.

My main philosophy here is to remove the "cool story"
logs, and hopefully make it easier for users and operators
to know the "why" for what is happening.

----

Also, now that we are entering the pod world,
I think it will become more important that our logs say "which"
container is having things done to it.

### DIFF
--- a/executor/runner/runner.go
+++ b/executor/runner/runner.go
@@ -168,8 +168,7 @@ func (r *Runner) prepareContainer(ctx context.Context) update {
 		return update{status: titusdriver.Lost, msg: err.Error()}
 	}
 
-	containerNames := getContainerNames(r.pod)
-	startingMsg := fmt.Sprintf("starting up %d containers: %s", len(containerNames), containerNames)
+	startingMsg := fmt.Sprintf("main containter (%s) created", r.container.TaskID())
 	return update{status: titusdriver.Starting, msg: startingMsg}
 }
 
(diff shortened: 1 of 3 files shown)


sargun:
These go back to control plane right? Should we make Sure they aren't parsing this?

corindwyer:
These all look fine to me. I didn't see any reason that actually gets used.

sargun (approved):

solarkennedy:
How worried should I be about adding these Status updates? 
I forget these end up being pod updates, and that is how they get to the control plane.

Should I just keep an eye out on etcd stuff as I roll this out?
Or should I tried to go on a diet and try to *remove* some updates?

alekar:
I think there is sufficient reason to be worried. But I doubt etcd will have an issue.

My concern is we might be breaking some platforms like metaflow which might be parsing these status message.

solarkennedy:
> I think there is sufficient reason to be worried. But I doubt etcd will have an issue.
> 
> My concern is we might be breaking some platforms like metaflow which might be parsing these status message.

Ack. I got pvolpe to check if this will break metaflow, it will not.
I do intend to change these messages more as we move more towards pods, so I guess let's break them now with benign changes

PR #625, opened 2021-05-03
Fix for Tini failure to launch

MyUmmaGumma:
Separate out the Networking and perf call handling. Catch first sendmsg only if network sys calls need to be intercepted.

Tini logs:
```
Networking system calls will be intercepted
Listening to networking syscall notifications on 6..
Can send the notify fd now!
  fd=6; events: EPOLLIN
Intercepted sendmsg - passthrough!
Stopped polling notifyfd on Tini, over to TSA!
Sent the notify fd to the seccomp agent socket at /titus-executor-sockets/titus-seccomp-agent.sock
```

### DIFF
--- a/tini/src/seccomp_fd_notify.c
+++ b/tini/src/seccomp_fd_notify.c
@@ -130,7 +130,7 @@ static int install_notify_filter(void) {
 		BPF_STMT(BPF_RET | BPF_K, SECCOMP_RET_ALLOW),
 	};
 
-struct sock_filter net_perf_filter[] = {
+struct sock_filter net_filter[] = {
 		/* X86_64_CHECK_ARCH_AND_LOAD_SYSCALL_NR */
 		BPF_STMT(BPF_LD | BPF_W | BPF_ABS, (offsetof(struct seccomp_data, arch))),
 		BPF_JUMP(BPF_JMP | BPF_JEQ | BPF_K, AUDIT_ARCH_X86_64, 0, 2),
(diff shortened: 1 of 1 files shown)


tini/src/seccomp_fd_notify.c
```diff
@@ -400,30 +385,31 @@ void maybe_setup_seccomp_notifer() {
...
-		/* 
-		   We need Tini to be ready to intercept and pass through the first 
-		   sendmsg() system call before sending the notify fd. 
-		   So, we wait until the polling has been set up.
-		*/
-		pthread_create(&thread_id, NULL, &catch_send_fd, (void *)&notify_fd);
-		pthread_mutex_lock(&wait_to_send);
-		while(send_notify_fd == 0) {
-			pthread_cond_wait(&ready_to_send, &wait_to_send);
+		char *is_handle_net = getenv("TITUS_SECCOMP_AGENT_HANDLE_NET_SYSCALLS");
```
solarkennedy, line 388: Maybe longer term we can have an enum for the types, just kinda awkward when we only have 2.
*then* it can be a param to `install_notify_filter` and it can know which type

tini/src/seccomp_fd_notify.c
```diff
@@ -150,21 +150,6 @@ struct sock_filter net_perf_filter[] = {
 		/* Trap connect */
 		BPF_JUMP(BPF_JMP | BPF_JEQ | BPF_K, __NR_connect, 0, 1),
 		BPF_STMT(BPF_RET + BPF_K, SECCOMP_RET_USER_NOTIF),
-		/* Trap perf-related syscalls */
```
solarkennedy, line 153: This is ok in the short term, but eventually we are going to want perf calls on things handling `net` too.

solarkennedy (approved):

tini/src/seccomp_fd_notify.c
MyUmmaGumma: Agreed. Probably cleanest with the clone() mechanism to send the notify fd.

tini/src/seccomp_fd_notify.c
MyUmmaGumma: agreed.

PR #626, opened 2021-05-04
Handle only one v4 and one v6 ip in a pod update

solarkennedy:
New version of k8s (1.19+) are picky about pods
having multiple ip addresses in their pod status.

Specifically, they only allow one of each.

This change ensures we only update pods with one of
each to be compliant, especially in the case
where we have EIPs.

----

Does anyone know what else uses this field?
I'm guessing spinnaker somehow uses this eventually?

### DIFF
--- a/vk/backend/backend.go
+++ b/vk/backend/backend.go
@@ -336,12 +336,27 @@ func (b *Backend) handleUpdate(ctx context.Context, update runner.Update) {
 	b.pod.Status.Message = update.Mesg
 	if update.Details != nil {
 		b.pod.Status.PodIP = update.Details.NetworkConfiguration.IPAddress
-		b.pod.Status.PodIPs = []v1.PodIP{
-			{IP: update.Details.NetworkConfiguration.IPAddress},
+
+		// We only get to have 1 IPv4 and 1 IPv6 address in this array
+		// https://github.com/kubernetes/kubernetes/blob/31030820be979ea0b2c39e08eb18fddd71f353ed/pkg/apis/core/validation/validation.go#L3289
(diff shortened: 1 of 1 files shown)


vk/backend/backend.go
```diff
@@ -336,12 +336,27 @@ func (b *Backend) handleUpdate(ctx context.Context, update runner.Update) {
...
+			b.pod.Status.PodIPs = []v1.PodIP{
+				{IP: update.Details.NetworkConfiguration.IPAddress},
+			}
+		}
+
+		// And now, V6, if we have one
+		// TODO: append a non-eip v6 if we can determine that we have one
+		if update.Details.NetworkConfiguration.EniIPv6Address != "" {
+			b.pod.Status.PodIPs = append(b.pod.Status.PodIPs, v1.PodIP{
+				IP: update.Details.NetworkConfiguration.EniIPv6Address,
```
solarkennedy, line 358: Oh crap I mixed up ENI and EIP here. this wrong.

solarkennedy: Wait no, I'm right, sorry. We don't have ipv6 EIPs yet.
What we are calling `EniIPv6Address` is the normal ip on a container.

sargun:
I think Spinnaker uses this. IIRC this factors into Sonar as well.

sargun (approved):

rgulewich (approved):

corindwyer:
Yes, the current integration uses annotations to get ip information.
https://github.com/Netflix/titus-control-plane/blob/caee1e51be0bacf14ea7d63ef15814ed1c024c65/titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeUtil.java#L117

PR #627, opened 2021-05-04
Move jumbo, and burst logic server side

sargun:
This starts the VPC Service side of work to move jumbo and
burst server side.


sargun:
This doesn't have any of the client-side code yet. That's a WIP.

vpc/proto/vpc.proto
```diff
@@ -126,6 +126,8 @@ message AssignIPRequestV3 {
     bool idempotent = 13;
     // Jumbo: I want "default routes" with an MTU of "Jumbo"
     bool jumbo = 14;
+    bool burst = 15;
+    uint64 bandwidth = 16;
```
solarkennedy, line 130: what do you think about encoding the units in the name?

vpc/proto/vpc.proto
```diff
@@ -126,6 +126,8 @@ message AssignIPRequestV3 {
     bool idempotent = 13;
     // Jumbo: I want "default routes" with an MTU of "Jumbo"
     bool jumbo = 14;
+    bool burst = 15;
```
solarkennedy, line 129: how about `burstable_bandwidth` ? Or a docstring for that this does?

vpc/proto/vpc.proto
```diff
@@ -148,6 +150,10 @@ message AssignIPResponseV3 {
         uint32 mtu = 2;
         Family family = 3;
     }
+    message Bandwidth {
+        uint64 bandwidth = 1;
+        uint64 burst = 2;
```
solarkennedy, line 155: Is this `bandwidth_limit` or `bandwidth_max` or something?

vpc/proto/vpc.proto
sargun: This is actually a good question on what the unit actually is...

vpc/proto/vpc.proto
sargun: What about "ceil"?

vpc/proto/vpc.proto
rgulewich: Big +1 on including the units in the name.

rgulewich (approved):

rgulewich:
LGTM, but I don't think I understand why we're moving it server-side.

sargun:
1. centralized > decentralized at sufficient scale
2. For benoit A/B test

PR #628, opened 2021-05-06
Do shell splitting on container's Command when an annotation is set

rgulewich:
The previous pod spec implementation assumed that we could get the TJC to do the docker shell splitting and then pass down the result in the container's Command and Args. Unfortunately, since the jobs are signed with the original Command and Args, this breaks Metatron. The executor will now parse Command when an annotation is set.

See https://github.com/Netflix/titus-kube-common/pull/15 for the kube-common changes

### DIFF
--- a/executor/runtime/types/pod.go
+++ b/executor/runtime/types/pod.go
@@ -755,6 +755,10 @@ func (c *PodContainer) extractServiceMesh() error {
 	return nil
 }
 
+// Optionally do shell splitting on the container's Command if the appropriate annotation is set. We do this
+// because lots of jobs still depend on this behaviour. Unfortunately, this parsing can't be done in
+// the TJC and passed down to the executor, because Titus clients sign jobs with the original unparsed Command
+// and Args. If the executor reports something different than what was in the signature, this breaks Metatron.
 func (c *PodContainer) parsePodCommandAndArgs() error {
 	uc := podCommon.GetUserContainer(c.pod)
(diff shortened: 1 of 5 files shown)


fabiokung (approved):

solarkennedy:
> LGTM, caveat that I don't know if this needs to be backwards compatible. Does it?

This should be ok because this is only with v1 style pod containers, so it shouldn't require backwards compat in that sense.
I presume that in the control plan if v1; then don't parse entrypoint?

executor/runtime/types/pod.go
```diff
@@ -746,6 +755,34 @@ func (c *PodContainer) extractServiceMesh() error {
 	return nil
 }
 
+func (c *PodContainer) parsePodCommandAndArgs() error {
```
solarkennedy, line 758: Can you add a docstring for why we do this?
I think your commit message has most of it

solarkennedy (approved):

alekar (approved):

corindwyer (approved):

executor/runtime/types/pod.go
rgulewich: Good idea - done.

rgulewich:
> This should be ok because this is only with v1 style pod containers, so it shouldn't require backwards compat in that sense.
> I presume that in the control plan if v1; then don't parse entrypoint?

Yeah, that's correct - this only happens for `v1` pods.

PR #630, opened 2021-05-07
Disable DAD on IMDS

sargun:
We use IPv4 "via" feature. If the IPv6 address on the to-intermediate
namespace link is not setup in time, then we run into an awkward situation
where the nexthop cannot be resolved.

### DIFF
--- a/metadataserver/inject/inject_linux.go
+++ b/metadataserver/inject/inject_linux.go
@@ -281,6 +281,13 @@ done:
 		return nil, fmt.Errorf("Could not set rp_filter in container NS: %w", err)
 	}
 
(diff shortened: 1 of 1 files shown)


solarkennedy (approved):

solarkennedy:
I don't understand, how do we use ipv4 (169.x) "via" ipv6 DAD?

sargun:
So, we started using a new linux Kernel routing feature called "via" for nexthops as part of this PR. See: https://tools.ietf.org/html/rfc5549

It basically uses an IPv6 nexthop for resolving the nexthop for the v4 destination, meaning p2p router links (effectively what the veth is), don't have to be addressed.

solarkennedy:
Huh. Shipit!

PR #632, opened 2021-05-10
Added basic pod /logs for C&W

solarkennedy:
This change uses /logs from the 'main' container and shares
it with other user containers.

The main benefit here is the /logs is consistent, and it
consumes the storage space from the main container. It is all
backwards compatible with all the other expectations around
how /logs behaves.

If we like this approach (bind mount from the main container),
it can be extended to other volumes, like /run.

Additionally, this PR includes injecting `tini` into the other
user containers. This is a bit hard, because docker won't do it
for us, even with `Init: true`.
https://github.com/moby/moby/blob/9f2b33f75cb094917c69e03ba101bc0bee8673fb/daemon/oci_linux.go#L737

This can be improved, but at least will now get files like:

    /logs/stdout
    /logs/stderr
    /logs/stdout-nginx
    /logs/stderr-nginx

Which is the killer feature of the /logs shared volume.

----

What do you all think of this approach?

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -34,6 +34,7 @@ import (
 	"github.com/docker/docker/api/types/container"
 	"github.com/docker/docker/api/types/events"
 	"github.com/docker/docker/api/types/filters"
+	"github.com/docker/docker/api/types/mount"
 	"github.com/docker/docker/api/types/network"
 	docker "github.com/docker/docker/client"
 	"github.com/docker/go-units"
@@ -1272,7 +1273,14 @@ func (r *DockerRuntime) Start(parentCtx context.Context, pod *v1.Pod) (string, *
 	}
 	go r.statusMonitor(eventCancel, r.c, eventChan, eventErrChan, statusMessageChan)
 
-	err = r.startUserContainers(ctx, pod, r.c.ID(), tiniConn)
+	inspectOutput, err := r.client.ContainerInspect(ctx, r.c.ID())
+	if err != nil {
+		eventCancel()
+		err = fmt.Errorf("container prestart error inspecting main container: %w", err)
+		return "", nil, statusMessageChan, err
+	}
+	mainContainerRoot := inspectOutput.GraphDriver.Data["MergedDir"]
+	err = r.startUserContainers(ctx, pod, r.c.ID(), tiniConn, mainContainerRoot)
 	if err != nil {
 		eventCancel()
 		return "", nil, statusMessageChan, err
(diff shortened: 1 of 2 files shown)


executor/runtime/docker/docker.go
```diff
@@ -1497,23 +1504,54 @@ func (r *DockerRuntime) createOtherUserContainer(ctx context.Context, v1Containe
...
 	// These labels are needed for titus-node-problem-detector
 	// to know that this container is actually part of the "main" one.
 	labels := map[string]string{
 		models.ExecutorPidLabel: fmt.Sprintf("%d", os.Getpid()),
 		models.TaskIDLabel:      mainContainerID,
 	}
+	mounts := []mount.Mount{
+		{
+			Type:     "bind",
+			Source:   mainContainerRoot + "/logs",
```
rgulewich, line 1517: Interesting - is this how Docker does it, under the covers?

Is the main container's `/logs` guaranteed to exist at this point?  Or I guess it doesn't matter because the main container will start first, and we'll create `/logs` anyway?

And just to confirm: if the user has `/logs` in their container pointing to `/path/to/somewhere/on/the/host`, we'll stomp over it, right?

rgulewich (approved):

executor/runtime/docker/docker.go
solarkennedy: Haha no, docker 

> Interesting - is this how Docker does it, under the covers?
Haha no. Well sorta.
Normally volumes are their own separate thing under /var/lib/docker/volumes, and then *that* is bind mounted into the container.

Here I'm just picking a super specific host bind mount that *happens* to be the root dir of another container (the main one).

This is not something docker can normally support because it can't make assumptions about the storage driver.
We on titus can though, because we are using overlay. If we change that to use something else, then this trick will break.

Amazingly, this trick does work on docker-for-mac. I have no idea how though (closed source).

> 
> Is the main container's `/logs` guaranteed to exist at this point? Or I guess it doesn't matter because the main container will start first, and we'll create `/logs` anyway?
> 

Yes, I'm relying on the main container to be ready first, it gets /logs from the pushEnvironment function.


> And just to confirm: if the user has `/logs` in their container pointing to `/path/to/somewhere/on/the/host`, we'll stomp over it, right?

If it is a dir, tar won't care. If it is a file called 'logs' in '/', then tar will overwrite it because we use `AllowOverwriteDirWithFile: true` in our tar command.

solarkennedy:
Coming back to this, messy rebase. Let's see if the tests are still green...

PR #634, opened 2021-05-11
Refactor standalone tests to be able to be run one at a time

solarkennedy:
This does a few things:

1. Runs standalone tests *twice* in the script instead of twice
   in the golang runner. I don't like this too much, but this
   split is temporary anyway and will go away once we are using
   v1 schema.
2. Removes the injected jobID var in each test. This was also
   kinda a function of the WithPod or WithCinfo. Again, this was
   temp code anyway, so I'm not too mornful about dropping it.
3. Put the parallizing code *in* each test. This kinda sucks, but
   it is only one line for each test, instead of having to maintain
   the big master list of tests at the top.
4. Makes all tests "public", so they can be run individually.
   This is the big deal, it means I can click on one in my IDE
   and run it. An iteration cycle of seconds instead of many minutes.

This is still not quite enough to make it all work on MacOSX, but
it is a prereq.

### DIFF
--- a/executor/standalone/standalone_test.go
+++ b/executor/standalone/standalone_test.go
@@ -5,8 +5,6 @@ import (
 	"flag"
 	"fmt"
 	"os"
-	"reflect"
-	"runtime"
 	"strconv"
 	"strings"
 	"testing"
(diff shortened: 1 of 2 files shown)


alekar (approved):

executor/standalone/standalone_test.go
```diff
@@ -262,20 +156,22 @@ func dockerPull(t *testing.T, imgName string, imgDigest string) (*dockerTypes.Im
...
-func testSimpleJob(t *testing.T, jobID string, usePodSpec bool) {
+func TestSimpleJob(t *testing.T) {
+	wrapTestStandalone(t)
 	ji := &JobInput{
 		ImageName:     alpine.name,
 		Version:       alpine.tag,
 		EntrypointOld: "echo Hello Titus",
-		JobID:         jobID,
-		UsePodSpec:    usePodSpec,
+		JobID:         t.Name(),
```
rgulewich, line 165: What do you think about having a wrapper function here to create the job ID?  So something like:
```
JobID: generateJobID(t.Name)
```

Having a unique job ID (between both the cinfo and pod spec tests) makes it much easier to debug tests when they fail in Buildkite - you can search for the job ID in both the build logs and the journald logs and not have to think too hard about which job you're actually looking at.  This becomes particularly difficult when all of the tests are running in parallel.

rgulewich (approved):

executor/standalone/standalone_test.go
solarkennedy: Oh yea of course. I was too focused on the *test* name, you are right the jobid is a great place to stick in that suffix.

solarkennedy: Added. We can have our cake (single test with a click) and eat it too (context-aware jobids in the test output)!

rgulewich: Wooo, thank-you! 🎉

PR #637, opened 2021-05-14
If pod has user env var index annotation set, don't require ContainerInfo

rgulewich:
This removes the last hard dependency on the ContainerInfo annotation: for generating the Metatron task-identity endpoint.  If the TJC sets the annotation indicating what index the user env vars start at, the executor has enough info to reconstruct ContainerInfo on its own, so let's do that instead.

kube-common PR: https://github.com/Netflix/titus-kube-common/pull/17

### DIFF
--- a/executor/runtime/types/pod_test.go
+++ b/executor/runtime/types/pod_test.go
@@ -28,7 +28,13 @@ var (
 	stringNil           *string
 	imageName           = "titusoss/alpine"
 	imageFullWithLatest = "docker.io/titusoss/alpine:latest"
+	testAppDetail       = "appDetail"
+	testAppName         = "appName"
+	testAppOwner        = "user@example.com"
+	testAppStack        = "appStack"
+	testAppSeq          = "appSeq"
 	testDigest          = "sha256:58e1a1bb75db1b5a24a462dd5e2915277ea06438c3f105138f97eb53149673c4"
+	testJobID           = "jobid"
 )
 
 func addPodAnnotations(pod *corev1.Pod, annotations map[string]string) {
@@ -92,8 +98,6 @@ func TestNewPodContainer(t *testing.T) {
 	imgName := "titusoss/alpine"
 	imgDigest := "sha256:58e1a1bb75db1b5a24a462dd5e2915277ea06438c3f105138f97eb53149673c4"
 	expectedImage := "docker.io/" + imgName + "@" + imgDigest
-	expAppName := "appName"
-	expAppOwner := "user@example.com"
 	expBwLimit := int64(128 * units.MB)
 	expCapabilities := &corev1.Capabilities{
 		Add:  []corev1.Capability{"NET_ADMIN"},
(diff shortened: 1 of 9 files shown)


alekar:
I am not a fan of overloading annotations. Can you consider another annotation `com.netflix.titus/useContainerInfo`?

alekar (approved):

rgulewich:
> I am not a fan of overloading annotations. Can you consider another annotation `com.netflix.titus/useContainerInfo`?

Yeah, fair point.  We could - but the idea is that in 2 months the original ContainerInfo annotation will be gone. :)

vk/backend/backend.go
```diff
@@ -143,20 +143,29 @@ func getContainerInfo(pod *v1.Pod) (string, error) {
 }
 
 func NewBackend(ctx context.Context, rp runtimeTypes.ContainerRuntimeProvider, pod *v1.Pod, cfg *config.Config, m metrics.Reporter) (*Backend, error) {
-	containerInfoStr, err := getContainerInfo(pod)
+	var containerInfo titus.ContainerInfo
```
solarkennedy, line 146: In the case of schema v1+, this containerInfo variable looks like it will remain empty, and passed to the backend.
I was trying to figure out, "if this variable is empty, then does the backend object even need it? And is it still empty by the time we write out the json file for the imds to use?

And I think I answer my on question with: Yes, it is ok to be empty, because when we create the container object, we don't even look at the containerinfo part if it is v1+.

And yes, I'm confusing the original pod.json that vk uses with the `containerConfigFile` json, which is written down later (which will our cinfo, no matter what, for imds to use).

executor/runtime/types/pod_test.go
```diff
@@ -1300,5 +1306,175 @@ func TestNewPodContainerEntrypointShellParsing(t *testing.T) {
...
+			Command:    expArgs,
+		},
+		TitusProvidedEnv: map[string]string{
+			"FROM_TITUS_1": "T1",
+			"FROM_TITUS_2": "T2",
+		},
+		UserProvidedEnv: map[string]string{
+			"FROM_USER_1": "U1",
+			"FROM_USER_2": "U2",
+		},
```
solarkennedy, line 1443: Great!

solarkennedy (approved):
With this change, we need to be super careful when mutating the pod to not break metatron.
Previously, it was impossible to break metatron because it *only* cared about the cinfo.

Maybe with the webhook we can provide a `addEnvVarToUserContainer()` func that does it "correctly" by prepending the var and bumping the index?

Because these tests are great, they can't "see" mutations from the pod by other means (webhook), and therefore we'll have no warning if someone ships a change that tacks on a "foo=bar" and breaks all metatron signatures because they forgot to update the index.

vk/backend/backend.go
rgulewich: > And I think I answer my on question with: Yes, it is ok to be empty, because when we create the container object, we don't even look at the containerinfo part if it is v1+.

Yeah, that's correct.  I _think_ you're telling me that I should add a comment here, which I totally will. :)

solarkennedy: Hehe. Sure. All this should go away in <6 months (please), so don't sweat it.

rgulewich:
> Maybe with the webhook we can provide a `addEnvVarToUserContainer()` func that does it "correctly" by prepending the var and bumping the index?

Yeah, great idea.  We're going to need something like that, for sure.

rgulewich:
After talking to @corindwyer, I made the following changes:

- If the ContainerInfo annotation is set, use it
- If not, require that the env var index annotation is set, and return an error if it's not

This should be fine, since the v1 pod code isn't running anywhere in prod yet. The first released version of the TJC with the code will include the env var annotation.

sargun (approved):

solarkennedy (approved):

PR #638, opened 2021-05-14
Enabled standalone tests runnable from mac (darwin)

solarkennedy:
This makes it so "unsupported" platforms (darwin) just are noops
instead of raising an unsupported error.

I think this is fine, I don't think anyone should expect a full
feature-set running Titus on darwin :)

But, this enables the majority of our standalone tests to actually
work on a mac. The few that don't I added a skipper and a comment
as to why they don't work, because they magically may start working
in the future.

It does mean that I had to make the standalone tests *opt out*
instead of opt in. This would only affect someone who is running
`go test` manually instead of `make test`, which is probably nobody.

But, we've finally made it. With this PR I can run standalone tests
in a few seconds with a click of a button from my IDE.

The full test suite takes 25 seconds on my laptop:

```
Running tool: /usr/local/opt/go/libexec/bin/go test -timeout 30s -coverprofile=/var/folders/qd/20hc8yr51nlg1x6n7ppjwzjh0000gn/T/vscode-goVhHHGp/go-code-cover github.com/Netflix/titus-executor/executor/standalone

ok  	github.com/Netflix/titus-executor/executor/standalone	25.023s	coverage: [no statements]
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,7 +51,7 @@ TEST_DIRS = $(shell go list -f 'TEST-{{.ImportPath}}' ./...)
 .PHONY: $(TEST_DIRS)
 $(TEST_DIRS): | $(clean)
 	$(eval import_path := $(subst TEST-,,$@))
-	CGO_ENABLED=0 go test -o test-darwin/$(import_path).test -c $(import_path)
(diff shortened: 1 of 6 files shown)


solarkennedy:
Bump?

rgulewich (approved):

PR #639, opened 2021-05-18
Remove v1 VPC Tool parts

sargun:
This removes the multi-generational VPC logic. It plumbs through an "Assignment" object, which then encapsulates (pointer receiver method) the logic to unroll the requisite information (container ENI, IPs, etc..). 

Still needs to be tested in staging, but I wanted to get a general overview of what people thought.


cmd/titus-vpc-tool/setup_container.go
```diff
@@ -20,10 +18,7 @@ func setupContainercommand(ctx context.Context, v *pkgviper.Viper, iipGetter ins
 			netns := v.GetInt("netns")
 			bandwidth := v.GetInt64("bandwidth")
 			burst := v.GetBool("burst")
-			jumbo := v.GetBool("jumbo")
 			switch strings.ToLower(v.GetString(generationFlagName)) {
-			case "v1":
-				return container.SetupContainer(ctx, iipGetter(), netns, uint64(bandwidth), burst, jumbo)
 			case "v2", "v3":
```
solarkennedy, line 22: remove v2 also if you want to get rid of both?

executor/runtime/docker/docker.go
```diff
@@ -653,7 +658,7 @@ func prepareNetworkDriver(parentCtx context.Context, cfg Config, c runtimeTypes.
 	// We intentionally don't use context here, because context only KILLs.
```
solarkennedy, line 658: Can you update this comment?

executor/runtime/docker/docker.go
```diff
@@ -665,126 +670,46 @@ func prepareNetworkDriver(parentCtx context.Context, cfg Config, c runtimeTypes.
...
-			case <-ctx.Done():
-				return errors.Wrap(ctx.Err(), "Error waiting for v1 assignment command to exit")
-			}
-			if err2 != nil {
-				log.WithError(err2).Error("Received error on termination of assignment command")
-				return errors.Wrap(err2, "Could not unassign task IP address")
-			}
-			return nil
-		}, nil
-	case vpcTypes.V3:
```
solarkennedy, line 766: Is your intent to remove *all* case statements when it comes to instance generation?
Or do you want to leave them in because we know we'll get a v4 some day?
Because in previous parts of the code you just took out the v1/v2 and wasn't sure if you intended to go "all the way" and remove all the case stuff.

executor/runtime/types/container.go
```diff
@@ -1185,15 +1196,6 @@ func populateContainerEnv(c Container, config config.Config, userEnv map[string]
 			env[metadataserverTypes.NetflixIPv6sEnvVarName] = vpcAllocation.IPV6Address.Address.Address
 		}
 
-		if vpcAllocation.ElasticAddress != nil {
-			env[metadataserverTypes.EC2PublicIPv4EnvVarName] = vpcAllocation.ElasticAddress.Ip
```
solarkennedy, line 1189: Is this env being set through some other means?

executor/runtime/docker/docker.go
```diff
@@ -1704,13 +1710,13 @@ func (r *DockerRuntime) setupPostStartLogDirTiniHandleConnection2(parentCtx cont
 		return fmt.Errorf("error mounting proc pid1 in titus init: %w", err)
 	}
 
-	if r.cfg.UseNewNetworkDriver && c.VPCAllocation().IPV4Address != nil {
+	if r.cfg.UseNewNetworkDriver && c.VPCAllocation() != nil {
```
solarkennedy, line 1713: Can UseNewNetworkDriver go away next? Are we not using that everywhere?

executor/runtime/docker/docker.go
```diff
@@ -1811,7 +1819,7 @@ func setupNetworking(burst bool, c runtimeTypes.Container, cred ucred) (cleanupF
 	}
 	defer shouldClose(netnsFile)
 
-	setupCommand := exec.Command(vpcToolPath(), setupNetworkingArgs(burst, c)...) // nolint: gosec
+	setupCommand := exec.CommandContext(ctx, vpcToolPath(), setupNetworkingArgs(burst, c)...) // nolint: gosec
```
solarkennedy, line 1822: Same deal, do we no longer have cleanup concerns like we did previously wrt context and command execution?

executor/runtime/docker/docker.go
```diff
@@ -1828,119 +1836,31 @@ func setupNetworking(burst bool, c runtimeTypes.Container, cred ucred) (cleanupF
...
-	default:
-		err = fmt.Errorf("Unknown generation: %s", g)
-		killCh <- struct{}{}
-		log.WithError(err).Error("Received allocation with unknown generation")
-		return nil, err
+	f2, err := os.Open(netnsPath)
+	if err != nil {
+		return nil, errors.Wrap(err, "Unable to open container network namespace file")
 	}
+	return func() error {
```
solarkennedy, line 1857: This is interesting. Is this supposed to be called if anything goes wrong in the above?
If yes, I don't see how that is possible with when we return normal errors above, and the only thing I see deferred is the contex cancel function.

executor/runtime/types/container.go
```diff
@@ -1185,17 +1183,25 @@ func populateContainerEnv(c Container, config config.Config, userEnv map[string]
...
+	if a := vpcAllocation.ElasticAddress(); a != nil {
+		env[metadataserverTypes.EC2PublicIPv4EnvVarName] = a.Ip
+		env[metadataserverTypes.EC2PublicIPv4sEnvVarName] = a.Ip
+	}
 
+	if a := vpcAllocation.ContainerENI(); a != nil {
+		env["EC2_VPC_ID"] = a.VpcId
+		env["EC2_INTERFACE_ID"] = a.NetworkInterfaceId
+		env["EC2_SUBNET_ID"] = a.SubnetId
 	}
```
solarkennedy: Ah this makes more sense to me

solarkennedy (approved):

cmd/titus-vpc-tool/setup_container.go
sargun: V2 was already partially removed.

cmd/titus-vpc-tool/setup_instance.go
```diff
@@ -17,24 +15,15 @@ func setupInstanceCommand(ctx context.Context, v *pkgviper.Viper, iipGetter inst
...
 			}
 			switch strings.ToLower(v.GetString(generationFlagName)) {
-			case "v1":
-				return setup.Setup(ctx,
-					iipGetter(),
-					locker,
-					conn,
-					v.GetString(interaceSubnet),
-					v.GetString(interfaceAccount))
-			case "v2":
```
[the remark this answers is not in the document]
sargun: V2 was already partially removed.

executor/runtime/docker/docker.go
sargun: So, the idea its to move away from the switch statements in throughout the code, and add a pointer method on the Assignment struct that gives the specific info the user may be interested in.

executor/runtime/types/container.go
sargun: See line 1197

executor/runtime/docker/docker.go
sargun: We apparently have it disabled when we run testing.

executor/runtime/docker/docker.go
sargun: "Different" concerns.

sargun:
@rgulewich I removed a bunch of "new" protobuf stuff you pulled in. Please take a look.

rgulewich (approved):

solarkennedy (approved):

PR #640, opened 2021-05-20
Fix the titus-sshd docker build for 2021

solarkennedy:
I think the gentoo approach is OK. It does give us
the benefit of being able to precisely control how
sshd is built, *and* keep all the features of a modern
ssh daemon (port forwarding, ssh CA auth, etc).

The downside of course is... gentoo and build times.

But this does work, and maybe it is just worth it to
fight the build once every couple of years.

----

Holy shit it works.

I still have a lot of extra junk in the dockerfile that is "temporary", so I'm going to remove that one piece at a time, `make itest` each time to ensure it *still* works.

### DIFF
--- a/hack/images/titus-sshd/Dockerfile
+++ b/hack/images/titus-sshd/Dockerfile
@@ -11,10 +11,9 @@ RUN emerge --sync
 RUN echo 'MAKEOPTS="-j32"' >> /etc/portage/make.conf
 RUN echo 'USE="-pam -gdbm -berkdb -bindist"' >> /etc/portage/make.conf
 # This is because Docker builds don't have CAP_SYS_PTRACE and cannot unshare
-RUN echo 'FEATURES="-sandbox -usersandbox -ipc-sandbox -network-sandbox -pid-sandbox"' >> /etc/portage/make.conf
-#TEMP
-ENV FEATURES="nostrip"
-RUN emerge -C openssh
+# Note: For some reason we cannot strip our binaries. Not sure why...
+RUN echo 'FEATURES="-sandbox -usersandbox -ipc-sandbox -network-sandbox -pid-sandbox nostrip"' >> /etc/portage/make.conf
+RUN emerge --unmerge openssh
 # We want to re-emerge openssl with the new USE flags, because the old USE flags can confuse portage
 RUN emerge --oneshot openssl
 # you need nss on the host to build nss in the prefix
--- a/hack/images/titus-sshd/Dockerfile
+++ b/hack/images/titus-sshd/Dockerfile
@@ -1,3 +1,10 @@
+# Welcome to the titus-sshd Dockerfile!
+#
+# This image uses gentoo to get fine-grained control over the
+# sshd binary and to relocate it to /titus/sshd, *even though*
+# it is still dynamically linked!
+#
+# First we setup a gentoo environment that can run *in* docker:
 FROM gentoo/stage3-amd64 as builder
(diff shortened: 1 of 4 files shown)


solarkennedy:
Again, just pushing this up here in case my computer explodes. `make itest` does work on this image this very second. I'll ask the gentoo team to not change anything :)

hack/images/titus-sshd/Dockerfile
```diff
@@ -3,35 +3,48 @@ RUN emerge-webrsync
...
 RUN echo 'MAKEOPTS="-j32"' >> /etc/portage/make.conf
 RUN echo 'USE="-pam -gdbm -berkdb -bindist"' >> /etc/portage/make.conf
-# This is because Docker builds don't have CAP_SYS_PTRACE
-RUN echo 'FEATURES="-sandbox -usersandbox"' >> /etc/portage/make.conf
-RUN echo "=sys-auth/libnss-compat-1.2 ~amd64" >> /etc/portage/package.accept_keywords
+# This is because Docker builds don't have CAP_SYS_PTRACE and cannot unshare
+RUN echo 'FEATURES="-sandbox -usersandbox -ipc-sandbox -network-sandbox -pid-sandbox"' >> /etc/portage/make.conf
+
+#TEMP
+ENV FEATURES="nostrip"
```
sargun, line 10: This is probably good to leave.

hack/images/titus-sshd/Dockerfile
```diff
@@ -3,35 +3,48 @@ RUN emerge-webrsync
 RUN emerge --sync
 RUN echo 'MAKEOPTS="-j32"' >> /etc/portage/make.conf
 RUN echo 'USE="-pam -gdbm -berkdb -bindist"' >> /etc/portage/make.conf
-# This is because Docker builds don't have CAP_SYS_PTRACE
-RUN echo 'FEATURES="-sandbox -usersandbox"' >> /etc/portage/make.conf
-RUN echo "=sys-auth/libnss-compat-1.2 ~amd64" >> /etc/portage/package.accept_keywords
+# This is because Docker builds don't have CAP_SYS_PTRACE and cannot unshare
+RUN echo 'FEATURES="-sandbox -usersandbox -ipc-sandbox -network-sandbox -pid-sandbox"' >> /etc/portage/make.conf
```
sargun, line 7: It probably makes sense to combine all these writes to make.conf?

hack/images/titus-sshd/Dockerfile
sargun, line 10: This is probably good to leave, or write to make.conf

sargun (approved):

hack/images/titus-sshd/Dockerfile
solarkennedy: Haha. *THIS* turned out to be the linchpin.
Without nostrip, the relocation fails and it segfaults!!!

hack/images/titus-sshd/Dockerfile
solarkennedy: Eh, I would like to just append the stuff we want and take the rest of make.conf... from stock?

solarkennedy:
Ok. Green build and I'm happy with the state of the dockerfile.

If shipit I'll merge, push to docker hub, and carefully upgrade titus-sshd via FP across our infra.

This brings us from `OpenSSH_7.7p1` > `OpenSSH_8.1p1`. (cc @JackOfMostTrades )

PR #641, opened 2021-05-20
Fix JSON encoding for teardown

sargun:


executor/runtime/docker/docker.go
```diff
@@ -1883,7 +1882,10 @@ func teardownCommand(netnsFile *os.File, allocation vpcapi.Assignment) error {
 		return errors.Wrap(err, "Could not start teardown command")
 	}
 
-	err = encoder.Encode(allocation)
+	marshaler := jsonpb.Marshaler{
+		Indent: "\t",
```
solarkennedy, line 1886: Wait what? What json decoder is caring if we use tabs or spaces???

sargun: It was jsonpb vs. encoding/json. I *thought* protobuf was supposed to be compatible, but apparently not with 1.3?

PR #642, opened 2021-05-21
Minor changes to make logs easier to understand in standlone tests

solarkennedy:
Lots of stuff doesn't work in standalone, and isn't supposed to.
These are a few changes that I think are safe to make the
tests less noisy, hopefully making it easier to spot and understand
the "real" error and not the "normal" errors that fly by all the time.


rgulewich (approved):

sargun (approved):

rgulewich:
This is super helpful - thanks for making this change!

PR #643, opened 2021-05-24
Fix Availability Zone Reconcilation Loop

sargun:
This fixes the availability zone reconcilation loop to iterate through
all accounts (both Trunk ENIs and Branch ENIs). We specific detachment
function requires some information it derives from the trunk ENI account,
because the disassociation operation happens in that account.

### DIFF
--- a/vpc/service/gc_v3.go
+++ b/vpc/service/gc_v3.go
@@ -39,6 +39,51 @@ const (
 	associated = "associated"
 )
 
+// getAllRegionAccounts gets regions accounts of the trunk ENI ("accounts"), as well as the
+// branch ENI accounts
+func (vpcService *vpcService) getAllRegionAccounts(ctx context.Context) ([]keyedItem, error) {
+	tx, err := vpcService.db.BeginTx(ctx, &sql.TxOptions{
+		ReadOnly: true,
+	})
+	if err != nil {
+		logger.G(ctx).WithError(err).Error("Could not start database transaction")
+		return nil, err
+	}
+	defer func() {
+		_ = tx.Rollback()
+	}()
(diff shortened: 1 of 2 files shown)


vpc/service/gc_v3.go
```diff
@@ -39,6 +39,51 @@ const (
...
+		ReadOnly: true,
+	})
+	if err != nil {
+		logger.G(ctx).WithError(err).Error("Could not start database transaction")
+		return nil, err
+	}
+	defer func() {
+		_ = tx.Rollback()
+	}()
+	rows, err := tx.QueryContext(ctx, "SELECT region, account_id FROM accounts")
```
solarkennedy, line 55: Although this would no longer be in a transaction, you could call `getRegionAccounts` and append instead of duplicate

vpc/service/service.go
```diff
@@ -434,7 +434,7 @@ func (vpcService *vpcService) getTaskLoops() []taskLoop {
 		},
 		{
 			taskName:   "availability_zone",
-			itemLister: vpcService.getRegionAccounts,
+			itemLister: vpcService.getAllRegionAccounts,
```
solarkennedy, line 437: Is there any other worker that needs to do "All" region accounts instead of just "RegionAccounts" ? Like is the bug in any other woker?

solarkennedy (approved):

vpc/service/service.go
sargun: I looked through the other workers, and it doesn't seem so.

sargun: A bunch of the other workers use region accounts, but they either are trunk eni specific or branch eni specific, and don't do both.

PR #644, opened 2021-05-24
Added the first multi-container workload standalone test

solarkennedy:
This adds a few more docker-for-mac workarounds, and then our
first multi-container workload test. This test should continue
to work regardless of the backend, but currently works
using the C&W implementation.

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -241,7 +241,14 @@ func (r *DockerRuntime) registerRuntimeCleanup(callback cleanupFunc) {
 
 func setupLoggingInfra(dockerRuntime *DockerRuntime) error {
 	var err error
-	dockerRuntime.tiniSocketDir, err = ioutil.TempDir("/var/tmp", "titus-executor-sockets")
+	var tmpDir = "/var/tmp"
+	if runtime.GOOS == "darwin" { //nolint:goconst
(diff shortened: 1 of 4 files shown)


executor/runtime/docker/docker.go
```diff
@@ -1279,7 +1293,7 @@ func (r *DockerRuntime) Start(parentCtx context.Context, pod *v1.Pod) (string, *
 		err = fmt.Errorf("container prestart error inspecting main container: %w", err)
 		return "", nil, statusMessageChan, err
 	}
-	mainContainerRoot := inspectOutput.GraphDriver.Data["MergedDir"]
+	mainContainerRoot := getMainContainerRoot(inspectOutput)
```
rgulewich, line 1296: Nice.

rgulewich (approved):

solarkennedy:
Well it is "good" that this test is failing for some reason in buildkite, probably docker-in-docker nonsense. 
This will take me a while to iterate ... unless I can ssh to one of the buildkite nodes maybe...

solarkennedy:
Alright I narrowed this down to the fact that we use the `vfs` driver in our standalone tests due to docker-in-docker.
I've made the shared volumes stuff gracefully degrade in this case (the test still should pass because we are not making any assertions about shared volumes).

PR #645, opened 2021-05-25
Use workload, not app, in v1 podspec annotations

rgulewich:
This is the executor side of https://github.com/Netflix/titus-kube-common/pull/18 - making the v1 pod spec annotations use "workload" instead of "app".

### DIFF
--- a/executor/runtime/types/pod.go
+++ b/executor/runtime/types/pod.go
@@ -475,6 +475,11 @@ func (c *PodContainer) LogUploadThresholdTime() *time.Duration {
 }
 
 func (c *PodContainer) MetatronCreds() *titus.ContainerInfo_MetatronCreds {
+	if c.titusInfo != nil {
+		// ContainerInfo was passed in the attribute: return its copy of the creds
+		return c.titusInfo.MetatronCreds
+	}
+
 	if c.podConfig.WorkloadMetadata == nil || c.podConfig.WorkloadMetadataSig == nil {
 		return nil
 	}
--- a/executor/runtime/types/pod.go
+++ b/executor/runtime/types/pod.go
@@ -176,8 +176,8 @@ func (c *PodContainer) AppArmorProfile() *string {
 }
 
 func (c *PodContainer) AppName() string {
-	if c.podConfig.AppName != nil {
-		return *c.podConfig.AppName
+	if c.podConfig.WorkloadName != nil {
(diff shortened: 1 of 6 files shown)


executor/standalone/standalone_test.go
```diff
@@ -1065,9 +1065,18 @@ func TestMetatron(t *testing.T) {
 		Version:         userSet.tag,
 		MetatronEnabled: true,
 		// The metatron test image writes out the task identity retrieved from the metadata service to `/task-identity`
-		EntrypointOld: fmt.Sprintf("/bin/bash -c \"cat /task-identity; grep %s /task-identity && grep jobAcceptedTimestampMs /task-identity | grep -E '[\\d+]'\"", t.Name()),
-		JobID:         generateJobID(t.Name()),
-		UsePodSpec:    shouldUsePodspecInTest,
+		EntrypointOld: strings.Join([]string{
```
rgulewich, line 1068: This was code that I didn't push as part of debugging metatron issues in https://github.com/Netflix/titus-executor/pull/637 - it's still useful for figuring out what went wrong, so I'll include it here.

corindwyer (approved):

executor/runtime/types/pod.go
```diff
@@ -475,13 +475,18 @@ func (c *PodContainer) LogUploadThresholdTime() *time.Duration {
 }
 
 func (c *PodContainer) MetatronCreds() *titus.ContainerInfo_MetatronCreds {
-	if c.podConfig.AppMetadata == nil || c.podConfig.AppMetadataSig == nil {
+	if c.titusInfo != nil {
```
rgulewich, line 478: Stashed changes that I forgot to push as part of the last PR. :(

executor/standalone/standalone_test.go
solarkennedy: If you want you could use a backtick in the string to make it a true golang string literal and avoid quoting escapes, or indenting, or breaking it up at all (if you are OK hard-coding TestMetatron).

https://play.golang.org/p/SGFvIC02Oah

executor/runtime/types/pod.go
solarkennedy: Hmm. I kinda expected to see this code on only the cinfo type implementation, but I guess we are getting to the point where *both* need it as we make the transition?

solarkennedy (approved):

PR #649, opened 2021-06-03
Move around security group checking logic

sargun:
This does a bunch of refactoring related to where security groups logic
is handled. Because the transition entry will be its own entry in
assignments, we need to change where the logic is, if we're to try
to do work across two discrete transactions (i.e. generate two
assignments, and populate them vs. trying to populate two assignments
in the same code). By shifting the logic into the DB of doing
verification at "commit" time, we simultaneously insert two
assignments in the same transaction without problem.


vpc/service/db/migrations/30_branch_eni_sg_check.up.sql
```diff
@@ -0,0 +1,46 @@
...
+CREATE CONSTRAINT TRIGGER sg_exists_on_update AFTER UPDATE OF security_groups ON branch_enis DEFERRABLE
+    FOR EACH ROW EXECUTE PROCEDURE check_sg_exists();
+
+CREATE CONSTRAINT TRIGGER sg_exists_on_insert AFTER INSERT ON branch_enis DEFERRABLE
+    FOR EACH ROW EXECUTE PROCEDURE check_sg_exists();
+
+
+CREATE OR REPLACE FUNCTION check_sg_modifications() RETURNS trigger
+    LANGUAGE plpgsql AS
+$$
```
solarkennedy, line 28: missing `;`? jk I have no idea when you need semicolons here, or even what `$$` means :)

vpc/service/reconcile_branch_enis.go
```diff
@@ -351,43 +351,6 @@ retry:
 
 	span.AddAttributes(trace.StringAttribute("networkInterfaceSecurityGroups", fmt.Sprintf("%+v", networkInterfaceSecurityGroups)))
 
-	_, err = fastTx.ExecContext(ctx, "UPDATE branch_enis SET mac = $1 WHERE branch_eni = $2 AND mac IS NULL",
```
solarkennedy, line 354: Why is it that we dont need `mac` anymore?

solarkennedy (approved):

vpc/service/reconcile_branch_enis.go
sargun: Mac was a column that was added a while ago. The original table didn't have MAC in it. The reconciler was supposed to "fix" that -- it's been in prod long enough where that should be set now.

Now, ENIs should have MAC addresses set when inserted into the DB.

PR #652, opened 2021-06-08
Block access to IPv6 IMDS on AWS-facing network

sargun:


solarkennedy:
Can you say more in the description about what this is/does?

solarkennedy (approved):

PR #655, opened 2021-06-09
Added an FSlock on titus-storage

solarkennedy:
This adds a fs lock on titus-storage, to avoid race condition when
multiple mounts/attaches are happening on an agent.

This was recommened in the beginning, so here it is after seeing the
race conditions in the wild.

I tried really hard to add sleep and debugs to see this in action,
but for some reason I never see `/tmp/titus-storage` on my agent,
even though I'm certain the right code is running (prints).
Is there something I'm missing?

### DIFF
--- a/cmd/titus-storage/main.go
+++ b/cmd/titus-storage/main.go
@@ -3,9 +3,13 @@ package main
 import (
 	"context"
 	"os"
+	"path/filepath"
+	"time"
 
+	"github.com/Netflix/titus-executor/fslocker"
 	"github.com/Netflix/titus-executor/logger"
 	"github.com/Netflix/titus-executor/utils/log"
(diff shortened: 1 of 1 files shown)


cmd/titus-storage/main.go
```diff
@@ -118,3 +127,21 @@ func newMountConfigFromViper(v *viper.Viper) MountConfig {
 		pid1Dir:       v.GetString(titusPid1DirFlagName),
 	}
 }
+
+func getExclusiveLock(ctx context.Context) (*fslocker.ExclusiveLock, error) {
+	stateDir := "/tmp/titus-storage"
```
sargun, line 132: Please use /run.

sargun (approved):

andrew-leung (approved):

PR #656, opened 2021-06-10
Statically wire up the entire IMDS, and handle link watch

sargun:
There were a bunch of race conditions in the IMDS due to how certain
behaviours were being handled in the Linux kernel (for example, veth
bringup was / is async). This in turn, makes sure that's all okay.

Beyond that, it does static programming of the neighbor table,
so we don't have to take any chances on neighbor discovery
"acting weird".

### DIFF
--- a/metadataserver/inject/inject_linux.go
+++ b/metadataserver/inject/inject_linux.go
@@ -14,7 +14,6 @@ import (
 	"time"
 
 	"github.com/Netflix/titus-executor/logger"
-
 	"github.com/vishvananda/netlink"
 	"github.com/vishvananda/netns"
 	"golang.org/x/sys/unix"
@@ -203,7 +202,19 @@ func setupNamespaces(ctx context.Context, pid1dir string) (*int, error) { // nol
 		return nil, fmt.Errorf("Could not set %q link up: %w", tocontainerInterfaceName, err)
 	}
 
-	// Make sure LO is up. This should be a noop:
+	// We need to do this because even though we set the links in the Linux kernel to up, the process of
+	// "the other side" recognizing that the other side is up is handled through an asynchronous process
+	// called link_watch.
+
+	// For example, let's say we have veth0 <-> veth1
+	// And I set veth0 up. Veth1 still thinks the layer 1 interface is down, until link watch fires
+	// at which point it marks layer 1 as up, and subsequently this triggers a bunch of setup.
+	err = waitForInterfacesUp(ctx, containerNSHandle, intermediateNSHandle)
+	if err != nil {
+		return nil, fmt.Errorf("Could not wait for netns to come up: %w", err)
+	}
+
+	// Make sure LO is up in the container. This should be a noop:
 	lo, err := containerNSHandle.LinkByName("lo")
 	if err != nil {
 		return nil, fmt.Errorf("Could not get container lo: %w", err)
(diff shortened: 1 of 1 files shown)


metadataserver/inject/inject_linux.go
```diff
@@ -338,3 +364,59 @@ done:
...
+			toIMDSLink, err = containerNSHandle.LinkByName(toimdsInterfaceName)
+			if err != nil {
+				return fmt.Errorf("Could not get %s: %w", toimdsInterfaceName, err)
+			}
+
+			toContainerLink, err = intermediateNSHandle.LinkByName(tocontainerInterfaceName)
+			if err != nil {
+				return fmt.Errorf("Could not get %s: %w", tocontainerInterfaceName, err)
+			}
+
```
solarkennedy, line 411: Do we have to continuously poll to get these? 
By the time we run this wait function, is it not safe to assume the Links themselves exist?

metadataserver/inject/inject_linux.go
```diff
@@ -338,3 +364,59 @@ done:
...
+	defer t.Stop()
+	for {
+		select {
+		case <-ctx.Done():
+			return nil, fmt.Errorf("Context complete before able to get LL addrs: %w", ctx.Err())
+		case <-t.C:
+			addrs, err := handle.AddrList(link, netlink.FAMILY_V6)
+			if err != nil {
+				return nil, fmt.Errorf("Could not list V6 addrs on %q: %w", link.Attrs().Name, err)
+			}
```
solarkennedy, line 379: Is this function specific to ipv6? 
If so, can you rename this function to `getLinkLocalIPv6Addr` maybe?
I just worry sometimes when I see `addr.IP` and making a mistake thinking that it is v4 when it is actually v6.

metadataserver/inject/inject_linux.go
```diff
@@ -326,6 +328,30 @@ done:
...
+		LinkIndex:    tocontainer.Attrs().Index,
+		Family:       netlink.FAMILY_V6,
+		State:        netlink.NUD_PERMANENT,
+		Type:         unix.RTN_UNICAST,
+		IP:           metadataserviceLinkLocal.IP,
+		HardwareAddr: metadataservice.Attrs().HardwareAddr,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("Cannot insert neighbor entry %w from imds NS to container NS", err)
+	}
```
solarkennedy, line 353: I look forward to when we can do this same trick for the ipv6 router or gateway or whatever.

solarkennedy (approved):

metadataserver/inject/inject_linux.go
sargun: In order to get the current _status_ of the link, we have to refetch the data. Netlink also has a "monitor" API, but the golang library doesn't suppot it as far as I can tell.

metadataserver/inject/inject_linux.go
sargun: Right, that makes sense.

metadataserver/inject/inject_linux.go
sargun: You mean on eth0? Or...?

solarkennedy: Yes, something like:

```
containerNSHandle.NeighSet(&netlink.Neigh{
  LinkIndex:    eth0.Attrs().Index,	
  Family:       netlink.FAMILY_V6,
  State:        netlink.NUD_PERMANENT,	
  Type:         unix.RTN_UNICAST,	
  IP:           ourIPv6GatewayIP.IP,	
  HardwareAddr: whateverOurIPv6RouterIsStaticallyGivenToUsFromEC2.Attrs().HardwareAddr,
}
```

And then never deal with neighbor discovery again?

sargun: Getting the actual MAC of the phantom router is non-trivial :(.

PR #657, opened 2021-06-10
Use modern busybox instead of alpine because pgrep works

solarkennedy:
cherry-picking a fix to reduce the diff in my other PR.

----

I'm banging my head on why tests fail on buildkite but not on my laptop.

I want to make sure this change has nothing to do with it.

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -1523,10 +1523,15 @@ func (r *DockerRuntime) k8sContainerToDockerConfigs(v1Container v1.Container, ma
 		l.Info("no mainContainerRoot available, volumes will not be sharable between containers")
 	}
 
-	// TODO: Use the same env logic as the main container
-	baseEnv := []string{
-		"TITUS_REDIRECT_STDERR=/logs/stderr-" + v1Container.Name,
-		"TITUS_REDIRECT_STDOUT=/logs/stdout-" + v1Container.Name,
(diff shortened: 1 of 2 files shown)


solarkennedy:
> I suppose you don't plan to merge this, since branch builds should give you the information you need?

I was going to go and merge if it is green, yes.

solarkennedy:
Well, I didn't expect this to fail, but it did. I think it is just revealing an underlying bug or difference between buildkite and my laptop (where it passes).

Here I see 
```
Jun 10 21:11:37 41339826f554 89ad544d5134[46]: [FATAL tini (26)] Failed to open stdout redirect path: No such file or directory
```

And:
```
time=2021-06-10T21:11:36Z level=info msg=no mainContainerRoot available, volumes will not be sharable between containers taskID=Titus-90130-TestBasicMultiContainer-0-2
```

Which indicates two bugs I can fix:
1. We have overlay2 on in these buildkite nodes.... eh but still docker-in-docker, I guess that still wont work :(
2. Tini should only redirect logs if we do have the /logs shared volume available (which I guess we won't with docker-in-docker). I didn't realize that was a fatal error for tini.

alekar:
Can we mirror the image in titusoss before we use it in tests?

alekar (approved):

solarkennedy:
> Can we mirror the image in titusoss before we use it in tests?

Can do

PR #658, opened 2021-06-15
Added base environment support to country & western pods

solarkennedy:
You could argue that some (not all) of this could be injected in the webhook.
One downside so that approach is that it would be split: we would
*still* have to have this `Env()` function in the executor to actually
compute the whole env (docker image vars, vpctool vars, etc). The other
downside is that populating the entire env in the pod would make our pod objects
*even* bigger, duplicating the arrays N times for N containers.

Some of these variables are not known to the webhook until the last
second, and they are not "container-specific" you might say.

When a user specifies an env variable for the whole job, they probably
expect it to show up on each container. This PR implements that.

It also starts a new variable, `TITUS_CONTAINER_NAME`, which we did have
on kubelet, so I'm bringing it back.

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -472,7 +472,7 @@ func (r *DockerRuntime) mainContainerDockerConfig(c runtimeTypes.Container, bind
 
 	// This must go after all setup
 	containerCfg.Env = c.SortedEnvArray()
-	containerCfg.Env = append(containerCfg.Env, "TITUS_CONTAINER_NAME="+c.ID())
+	containerCfg.Env = append(containerCfg.Env, "TITUS_CONTAINER_NAME="+c.TaskID())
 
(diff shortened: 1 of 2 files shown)


rgulewich (approved):

solarkennedy:
Rebased.

Just to reiterate, this is a temp solution here to get *some* environment variables in the C&W containers.

When we have control-plane API support, we'll get some variables injected into our env for free (like `NETFLIX_ACCOUNT_ID`), Then we'll still have to have some notion of "late" env variables injected at the last minute by the executor (hopefully only the ones we really have to vpc stuff).

In other ways, this naive approach is "wrong" because it will inject variables that come from Image I think?

*And*, combine this with @rgulewich 's change to use the v1 pod spec and put system and user vars in the pod `env`, this may be a hint that this is going to get more complicated :)

I would like to revisit this approach after we have just pod spec stuff, no cinfo.

PR #659, opened 2021-06-15
Added multi-container event handling for country & western pods

solarkennedy:
This changes makes the 'watchForEvents' function that listens for docker
events and responds to them "multi-container aware".

That is, it now responds to arbitrary container ids and container names.

In this way, we can get task messages for any container, not just the
main one. If a sidecar ooms, dies, or whatever, the whole task will finish,
and log appropriately.

Note that this branch doesn't include the platform-sidecar distinctions.
I can rebase on that when it is in master, as we *dont* want an entire
task to finish just because a platform sidecar dies.

Future work might include only *optionally* tearing down the ask
if a sidecar container dies, allowing for custom lifecycles on a
per-container basis, but for now we treat all user containers the same.

----

I don't know how important the cancel function stuff is. 
Seems to work fine without it?


solarkennedy:
Now that I have a struct to hold sidecar info, I'm going to discard this and try again.

PR #661, opened 2021-06-16
Use podEnv whenever there is at least 1 container

solarkennedy:
Now that we can have multiple containers in the pod,
we should always return the podEnv in that case.

I dare say we should never return containerInfoEnv?

----

The cinfo env was subtly different and wrong (didn't have `NETFLIX_ACCOUNT_ID`).


rgulewich (approved):

rgulewich:
> I dare say we should never return containerInfoEnv?

As soon as we've moved to the full pod-based implementation, `containerInfoEnv` will go away, so this works for the time being.

PR #662, opened 2021-06-17
Refactor: Promote user and platform containers as real attributes of the runtime struct

solarkennedy:
In some sense, this is much cleaner than constantly parsing the pod object
all the time.

It does solidify the difference between user/platform containers,
hopefully that is a good thing.

Someday, the bit `Container` object can be broken down in to the job stuff
and the `UserContainers` can have all user containers, not just the
extra ones!

The reason for this refactor is to make it easier to store more methods
and attributes into the ExtraContainer struct to make it easy to do
bookkeeping with them. You can see the first thing that becomes
easier is tracking the container id for each one. I no longer need
and extra map hanging around just for that, and I can iterate over
the real objects instead of their names. Good stuff.

### DIFF
--- a/executor/runtime/types/container.go
+++ b/executor/runtime/types/container.go
@@ -148,6 +148,10 @@ type TitusInfoContainer struct {
 	logUploaderConfig           uploader.Config
 
 	pod *corev1.Pod
+	// extraUserContainers stores and array of metadata about all the non-main user containers
+	extraUserContainers []*ExtraContainer
+	// extraUserContainers stores and array of metadata about all the platform-defined containers
+	extraPlatformContainers []*ExtraContainer
 
 	// GPU devices
 	gpuInfo GPUContainer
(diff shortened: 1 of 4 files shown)


executor/runtime/types/container.go
```diff
@@ -148,6 +148,10 @@ type TitusInfoContainer struct {
 	logUploaderConfig           uploader.Config
 
 	pod *corev1.Pod
+	// extraUserContainers stores and array of metadata about all the non-main user containers
+	extraUserContainers []*ExtraContainer
+	// extraUserContainers stores and array of metadata about all the platform-defined containers
```
solarkennedy, line 153: Hmm, do you think maybe I should do a map here so I can more naturally iterate over these by name, and look up one of their healths by name?

rgulewich: Maybe, though I'd assume YAGNI for now and refactor that as part of the healthcheck work, if it makes sense then.

rgulewich (approved):

PR #663, opened 2021-06-17
Pod spec v1 fixes for failing integration tests

rgulewich:
This includes a bunch of fixes found through the integration tests:

- S3 log uploaders weren't getting set correctly due to a bug in `urfave/cli` with parsing string slices. This was fixed by upgrading to a newer version of the library
- Docker image parsing wasn't working correctly when the registry contained a port
- Removed fallbacks to ContainerInfo, since it's no longer being fully populated for v1 pods
- Changed how system vs user env vars are being determined for the Metatron task identity:
    - Use a list of the system env vars passed from the TJC
    - VK adds a bunch of env vars to the list, so remove them, too
- The TJC is setting `terminationGracePeriodSeconds` to 10min. In cases where the user workload has exited, or the container didn't start, stop the task immediately, rather than in 10 min. This was causing a weird issue where the IP address was getting freed in vpcservice before the container actually exited. Further containers coming up on the box would get duplicate IP errors.
- Don't override the process that gets written to the container config file unless it's unset
- Make the executor backend enable debug logging when the `DEBUG` env var is set
- More debug logging


executor/runtime/docker/docker.go
```diff
@@ -2108,7 +2109,7 @@ func (r *DockerRuntime) setupGPU(ctx context.Context) error {
 }
 
 // Kill uses the Docker API to terminate a container and notifies the VPC driver to tear down its networking
-func (r *DockerRuntime) Kill(ctx context.Context) error { // nolint: gocyclo
+func (r *DockerRuntime) Kill(ctx context.Context, wasKilled bool) error { // nolint: gocyclo
 	logger.G(ctx).Debug("Shutting down main container")
```
solarkennedy, line 2113: Can you do:
```
if wasKilled {
  logger.G(ctx).Debug("Shutting down main container because we were asked to stop from the API")
} else {
  logger.G(ctx).Debug("Shutting down main container because it finished or died")
}
```
There have been a number of times where I which I had that distinction in the logs, and thanks to you, now we can know

solarkennedy (approved):
I don't understand, were these tests failing on master?

cmd/titus-executor-backend/main.go
```diff
@@ -87,6 +87,11 @@ func main() {
 	defer m.Flush()
 
 	app.Action = func(c *cli.Context) error {
+		if mainCfg.debug {
```
solarkennedy, line 90: Huh. I thought for sure debug logs were enabled in the *standalone* tests, is this a different class of debugs or a different class of integration tests that you are refering to?

rgulewich: A bit of a naming error on my part - this is fixes for titus-functional-tests that run against real Titus stacks, not the integration tests in this repo.  Sorry for the confusion. :\

executor/runtime/docker/docker.go
rgulewich: Good idea - will do.

PR #664, opened 2021-06-18
Fix the titus-sshd aa profile to allow for scratch containers

solarkennedy:
For scratch containers, the run-titus-sshd touches

* /etc/passwd
* /etc/profile

(and a few other files, but those are the sensitive ones)
For sshd to startup properly.

The AA profile was blocking those, and whatever testing I was
doing before where it works must have had AA off or in complain mode.

### DIFF
--- a/root/etc/apparmor.d/docker_fuse
+++ b/root/etc/apparmor.d/docker_fuse
@@ -43,9 +43,29 @@ profile docker_fuse flags=(attach_disconnected,mediate_deleted) {
 
   deny /titus/** wl,
 
-/titus/sshd/usr/sbin/run-titus-sshd Cx,
+  # Special sub-profile just for the init script for sshd
+  /titus/sshd/run-titus-sshd Cx,
+  profile run-sshd /titus/sshd/run-titus-sshd {
+    # We trust everything under /titus
+    # No writes though...
+    /titus/** rmlk,
+    /titus/sshd/usr/sbin/sshd ix,
 
-  profile sshd /titus/sshd/usr/sbin/run-titus-sshd {
+    # You're allowed to do anything with yourself
+    @{PROC}/@{pid}/** rwmlk,
+    @{PROC}/self/** rwmlk,
+
+    /dev/{null,ptmx,tty,urandom,log} rwmlk,
+    /dev/pts/* rwmlk,
+
+    /etc/{gai.conf,group,nsswitch.conf,shadow} rmkl,
+    # The ssh startup script does edit these files if they need modification for sshd to start
+    /etc/{passwd,profile} rwmlk,
+  }
+
+  # Once we are running, this is the profile that the sshd binary itself gets
+  /titus/sshd/usr/sbin/sshd Cx,
+  profile sshd /titus/sshd/usr/sbin/sshd {
     signal (send,receive) peer=@{profile_name},
 
     network,
--- a/root/etc/apparmor.d/docker_fuse
+++ b/root/etc/apparmor.d/docker_fuse
@@ -43,29 +43,35 @@ profile docker_fuse flags=(attach_disconnected,mediate_deleted) {
 
   deny /titus/** wl,
 
+  /titus/sshd/usr/sbin/sshd Cx -> &docker_titus//exec-sshd,
+  /titus/sshd/run-titus-sshd Cx -> &docker_titus//sshd,
+
   # Special sub-profile just for the init script for sshd
-  /titus/sshd/run-titus-sshd Cx,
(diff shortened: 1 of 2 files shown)


sargun:
So, are any of these files read by the relocated userspace without a user logging in? My fear is that a corrupted version of one of these files could be used to trigger an exploit in sshd.

root/etc/apparmor.d/docker_fuse
```diff
@@ -71,7 +71,9 @@ profile docker_fuse flags=(attach_disconnected,mediate_deleted) {
     /dev/{null,ptmx,tty,urandom,log} rwmlk,
     /dev/pts/* rwmlk,
 
-    /etc/{gai.conf,group,nsswitch.conf,passwd,shadow} rmkl,
+    /etc/{gai.conf,group,nsswitch.conf,shadow} rmkl,
+    # The ssh startup script does edit these files if they need modification for sshd to start
```
sargun, line 75: Is there any way that we can have another profile for the init script?

sargun: Like, if the init script is just another binary path, we should be able to do:

```
/titus/sshd/usr/sbin/init-titus-sshd {
... Stuff it does ...
}
```

Right?

sargun (changes requested):

root/etc/apparmor.d/docker_fuse
solarkennedy: That is a great idea.
I can experiment with this, but is it just based on the path of the binary? does `exec` get a new profile based on the new path, or does it keep the old profile it used to have?

But yea, I think ideally I can have a second profile just for the init script, and let the next profile when `exec sshd` happens and let that take over.

sargun: It should automatically switch profiles on exec based on path unless you're doing stuff with custom hats.

root/etc/apparmor.d/docker_fuse
```diff
@@ -43,9 +43,29 @@ profile docker_fuse flags=(attach_disconnected,mediate_deleted) {
...
+    /dev/pts/* rwmlk,
+
+    /etc/{gai.conf,group,nsswitch.conf,shadow} rmkl,
+    # The ssh startup script does edit these files if they need modification for sshd to start
+    /etc/{passwd,profile} rwmlk,
+  }
 
-  profile sshd /titus/sshd/usr/sbin/run-titus-sshd {
+  # Once we are running, this is the profile that the sshd binary itself gets
+  /titus/sshd/usr/sbin/sshd Cx,
```
sargun, line 67: Are there any things we can remove from here?

sargun (approved):

root/etc/apparmor.d/docker_fuse
solarkennedy: Not sure. 
How about I'll deploy this profile on my dev stack, ensure aa is enabled, start with a *blank* profile, try to start a scratch container, ensure it doesnt work, and then slowly add lines until it does work?

In theory the scratch container requires the most out of this `/titus/sshd/usr/sbin/run-titus-sshd` profile, so it should be exercising everything.

solarkennedy:
@sargun can you look one more time ? I split out a distinct profile for the init script, started with nothing, and added in lines until the denies stopped. 

I added explicit transitions because I couldn't figure out how to depend on the binary name.

Here is what PS shows when it is done:
```
docker_titus (enforce)          10000    1704005  0.0  0.0   8804  1020 ?        Ss   20:11   0:00 /apps/titus-executor/bin/titus-nsenter /titus/sshd/run-titus-sshd -D -e
docker_titus//&docker_titus//exec-sshd (enforce) 10000 1704012 0.0  0.0 6648 5008 ? S 20:11   0:00 sshd: /titus/sshd/usr/sbin/sshd -D -e [listener] 0 of 10-100 startups
docker_titus (enforce)          10000    1704037  0.0  0.0 840804 11364 ?        Ssl  20:11   0:00 /titus/spectatord/spectatord
docker_titus (enforce)          10000    1704038  0.0  0.0  90672  7152 ?        Ssl  20:11   0:00 /titus/atlas-titus-agent/atlas-titus-agent -t nf.subnet=subnet-04fca5dd4bbf221cb,nf.vpc=vpc-46f5a223
docker_titus (enforce)          10000    1704158  0.0  0.0   2984   996 ?        Sl   20:11   0:00 /pause
docker_titus (enforce)          10000    1704176  0.0  0.0 107528  4912 ?        Ssl  20:11   0:00 bin/adminlogs
```

sargun (approved):

PR #665, opened 2021-06-18
Up the EBS attach/detach timeout from 20 to 60 seconds

solarkennedy:
I think we just need to give it more time to reduce
flakiness in our tests and the user experience.


cmd/titus-storage/ebs.go
```diff
@@ -130,8 +130,8 @@ func ebsStop(ctx context.Context, ec2Client *ec2.EC2, c MountConfig, ec2Instance
 	// *Sometimes* ebsStop gets called *defore* a container dies, but *after* a *different* container
 	// comes up, see it is attached, and mounts. In this case we need to still panic here, but it is "ok"
 	// In that situation we should not continue to detach a volume out from underneath it
-	l.Printf("Waiting up to %d seconds for %s to be not in use...", 30, actualDeviceName)
-	err = waitForDeviceToBeNotInUse(actualDeviceName, 30)
+	l.Printf("Waiting up to %d seconds for %s to be not in use...", 60, actualDeviceName)
```
andrew-leung, line 133: Not sure if this is wired up but a configurable property would provide some future proofing.

andrew-leung (approved):

PR #666, opened 2021-06-18
Added v0 of container healthchecks

solarkennedy:
Surprising to me, users can add healthchecks to their docker *images*
today! We just don't do anything except report it on the updates channel.

This PR adds support for accepting liveness probes from the pod spec
and converting them to use docker healthchecks on the backend.

This is a feature we get "for free" because of the C&W design (using
docker natively as the executor).

This currently doesn't help the user very much, because I don't have
the wiring in place to see health events. I'm waiting on my big
refactor PR to plumb that through.

However, soon we'll need to address a big elephant, which is "what"
to do with this data. Surprising to me (TIL), healthchecks don't
have anything to do with restart policy!

We will have to implement our own connector, like kubelet does
to kill the container once it goes unhealthy, and let the restart
policy take effect.
https://github.com/moby/moby/issues/28400

But that can be in v1 (this is v0).

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -1604,16 +1604,18 @@ func (r *DockerRuntime) k8sContainerToDockerConfigs(v1Container v1.Container, ma
 	// to ignore whatever entrypoing is on the *image*. We want the normal docker behavior here,
 	// but we *also* want tini.
 	dockerEntrypoint := append([]string{"/sbin/docker-init", "-s", "--"}, v1Container.Command...)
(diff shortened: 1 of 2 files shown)


solarkennedy:
This PR is safe to merge, but it won't have any real effect on anything until we can easily expose or kill the container.

For that (v1), I would like to block on https://github.com/Netflix/titus-executor/pull/670 so that we have a more formal place to put extra sidecar data and status.

But this PR can be merged in now, and at least the health data of a sidecar will be available in the raw titus-executor logs.

rgulewich (approved):

PR #667, opened 2021-06-18
Bumped titus-api-definitions to include the new network mode api and update to latest protogen

solarkennedy:
In 2020 google changed the golang protobuf stuff https://github.com/protocolbuffers/protobuf-go#historical-legacy:

> It has been superseded by the google.golang.org/protobuf module, which contains an updated and simplified API, support for protobuf reflection, and many other improvements. We recommend that new code use the google.golang.org/protobuf module.

And we've been kinda fighting it for a while. I think my version of protoc got updated and I figured, lets move forward instead of downgrading (my laptop)?

So I fought the compile for a couple of hours.
Mostly reading
https://developers.google.com/protocol-buffers/docs/reference/go-generated#invocation

And understanding the relationship between protos and go modules.
Here are the big changes:

1. grpc files are created separately, so 2 protoc invocations
2. Specifying the package in the .proto file is deprecated and discouraged (which is why I use the command line args so much)
3. New Unimplemented stuff are semi-required, but encouraged.

And then on the titus side:

1. I got rid of the makefile magic. I thought it was just too much and I couldn't figure out what was going on.
2. Updated all the go.mod references to use the new compiler stuff
3. Updated the go.mod reference to pick up the latest titus api definitions

Its kinda a lot of stuff at once, but in the end there should be no functional changes to titus stuff.

I might be able to split this apart into separate PRs if you are worried about the big diff, but in the end the "Big" change is the upgrade to protoc, go-proto-gen, and all the changes surrounding that to make it happen.


solarkennedy:
I'm going to discard this for now.

Someday in the future we will have to upgrade protobuf-gen-go and move to the new golang/protobuf away from the old github.com/protobuf.

PR #671, opened 2021-06-24
Added new Pod task identity endpoint to the titus-imds

solarkennedy:
This new taskPodIdentity endpoint is very similar to the
existing taskIdentity endpoint, except instead of providing
containerinfo, it provides a full pod. This will provide
for attesting task identity based on containers (plural)
instead of containers.

Additionally, by providing the whole pod, we can just let
Security (metatron) do whatever they want based on the pod
object, no changes needed on our side.

The downside is that the pod object isn't quite as structured
when it comes to the metadata that metatron wants. Any consumer
of this pod will have to parse the annotation strings.

Another potential security issue here is that we are now exposing
the pod object to the pod processes! It may be a slippery slope
if we make this a public api to other consumers.

Another potential issue is that this pod object is loaded on titus-imds
start, and frozen. We don't reload the json, even though things like
podstatus do get updated in that file. They may be confusing?

Until this is finalized, I made the loading of this object not-fatal
to the imds. If there is any error, the show will go on.


metadataserver/identity.go
```diff
@@ -41,6 +43,43 @@ func (ms *MetadataServer) generateTaskIdentity() *titus.TaskIdentity {
...
+func (ms *MetadataServer) generateTaskPodIdentity() (*titus.TaskPodIdentity, error) {
+	s := int64(time.Now().Second())     // from 'int'
+	n := int32(time.Now().Nanosecond()) // from 'int'
+	now := &timestamp.Timestamp{Seconds: s, Nanos: n}
+	taskStatus := titus.TaskInfo_RUNNING
+
+	if ms.pod == nil {
+		return nil, fmt.Errorf("No pod object available to marshal as a task identity")
+	}
+	podData, err := proto.Marshal(ms.pod)
```
JackOfMostTrades, line 55: Was this the bit that was supposed to be the k8s wire format rather than the proto encoded?

solarkennedy: Wait, I thought proto-encoded *was* the wire format?

JackOfMostTrades: Nope, they've got some whole envelope scheme for their wire format with a magic header: https://kubernetes.io/docs/reference/using-api/api-concepts/#protobuf-encoding

From their docs:

> If you want to store or interact with proto-formatted Kubernetes API objects, we recommend using the "official" serialization stack in k8s.io/apimachinery. Directly serializing these types to proto will not result in data that matches the wire format...

solarkennedy: Oh wow. I missed that on the first read.

Gosh, it is just the two of us here, we are not building an api for the whole world.
I don't think we have any need for that wrapping.
Can we get away with just one layer of serialization? We don't need to literally match the wire format between kubelet and api server here.

JackOfMostTrades: Yea, I mean, I guess it goes against that k8s recommendation of not using the messages directly, but I kind of feel like it doesn't really apply here? But in that case this serialization to bytes seems unnecessary; should the `TaskPodIdentity` message just embed the `Pod` field as the k8s `Pod` message rather than bytes?

solarkennedy: That part no, I don't think so, because it would mean... we would have to bump the software any time there is a protobuf change in the pod spec (which we don't control).

But then again, we kinda do. The pod object we are serializing in the first place comes from our golang k8s client libs. 

I'm not super sure. I just don't have that much experience with protobufs to be sure if this is a good idea or now.
Maybe @sargun might now what is best here?

Certainly putting the pod as bytes in this field I think is "safe". The pod will change, and when it does, we won't have to rebuild any of *our* protos at all. Even you won't either. Your side will have to upgrade to later version of k8s client libs as we upgrade (+/- one version as we go), but that is inevitable if you are sucking in pod objects.

Double wrapping the pod object in another proto? That seems not needed for us here. Heck we are *already* doing that with this task ident proto!

JackOfMostTrades: Making the field `bytes` or `Pod` yields the same thing on the wire so I suppose it doesn't actually make a difference to how we'll consume it. It would just save you a step serializing to bytes explicitly.

sargun: So, I strongly recommend not using proto, as the proto serialization / deserialization isn't 100% compatible with the Kubernetes serialization / deserialization (for example, IIRC, you can have defaults in the kubernetes spec). 

I think (and I say this with a sense of depression), the easiest way to do this is to turn TaskPodIdentity into a CRD, and register it, and use the Kubernetes serialization libraries.

solarkennedy: Waaaaa?

This sounds way overkill. We don't wire-protocol compatibility here, we are not talking to the api server. 
Heck I could just `cat pod.json` and stick that in the `bytes`! All we really need is some way for someone to see what the current pod object is.

sargun: You definitely can serialize the pod object to json (or literally cat pod.json), and stuff it into a string / bytes. I think that's a good idea.

cmd/titus-metadata-service/main.go
```diff
@@ -91,6 +93,30 @@ func readTaskConfigFile(taskID string) (*titus.ContainerInfo, error) {
...
 
+func readTaskPodFile(taskID string) (*corev1.Pod, error) {
+	if taskID == "" {
+		log.Errorf("task ID is empty: can't read pod config file")
+		return nil, fmt.Errorf("task ID env var unset: %s", taskInstanceIDEnvVar)
+	}
+
+	// This filename is from VK, which is /run/titus-executor/$namespace__$podname/pod.json
+	// We only use the default namespace, so we hardcode it here.
+	confFile := filepath.Join("/run/titus-executor/default__", taskID, "pod.json")
```
rgulewich, line 104: It would be kinda nice if this path was abstracted behind a function somehow, or a constant somewhere - wouldn't this also be used in the titus-vk implementation?

metadataserver/identity.go
rgulewich: JSON-ified pod stuffed into bytes seems reasonable, as long as @JackOfMostTrades is cool with that.

JackOfMostTrades: As long as I can parse it back out easily enough in Java I'm fine with that. Hopefully using some k8s library so that I can get it into a POJO. I'm guessing that this would just match the proto JSON format?

rgulewich (approved):

metadataserver/identity.go
solarkennedy: `POJO` ?

JackOfMostTrades: [Plain old java object](https://en.wikipedia.org/wiki/Plain_old_Java_object#JavaBeans)

cmd/titus-metadata-service/main.go
solarkennedy: I agree.

There *is* some code in VK that does this calculation. The only weirdness is that it assumes you *already* have the pod object. Here we only know the taskid. Eh.

@sargun would it be ok to import executorapi in here from VK? Or would it be better just to hard code this path?

metadataserver/identity.go
```diff
@@ -52,7 +52,7 @@ func (ms *MetadataServer) generateTaskPodIdentity() (*titus.TaskPodIdentity, err
 	if ms.pod == nil {
 		return nil, fmt.Errorf("No pod object available to marshal as a task identity")
 	}
-	podData, err := proto.Marshal(ms.pod)
+	podData, err := json.Marshal(ms.pod)
```
JackOfMostTrades, line 55: Do you need to use [protojson](https://pkg.go.dev/google.golang.org/protobuf/encoding/protojson) instead of json?

> This package produces a different output than the standard "encoding/json" package, which does not operate correctly on protocol buffer messages.

solarkennedy: Gosh. I don't know.
That isn't how this pod object is written down on disk in the first place, it is just normal json?

protojson.marshal() requires a proto message in the first place, which I don't have.

I can skip the marshaling step entirely and give you the *literal* bytes on disk, but again, that is just normal json, not protojson.

JackOfMostTrades: /me so confused

If `ms.pod` isn't a proto message how would `proto.Marshal(ms.pod)` have worked?

solarkennedy: ms.pod is just a golang object of the pod struct.

`proto.Marshal(ms.pod)` works and turns it into bytes for transport. It somehow knows how to work with any interface (struct)

`json.Marshal(ms.pod)` works because the struct has `json:` annotations so it knows how to make a json byte string.

Keep in mind the json I'm working with on disk is just plain old json:
https://github.com/Netflix/titus-executor/blob/6c1a85add201991ef72ca7e97fda9dc485e77ad1/vk/backend/backend.go#L275

I'm using plain json load to read it in the first place. It can't have that much secret sauce?

metadataserver/identity.go
sargun: This approach sounds good to me.

metadataserver/identity.go
solarkennedy: Got confirmation from @sargun that it won't be possible to use jsonpb in this case, because we don't have the original proto.

PR #672, opened 2021-06-28
Make prefix length of IPv6 addresses the same

sargun:
Right now, there's a bug where if we assign an address using
new IPv6 addresses on an interface, we install a /64, versus
a /128.


solarkennedy:
Huh, is it really that easy?

solarkennedy (approved):

PR #673, opened 2021-06-28
Refactor logic around log / filesystem watcher shutdown

sargun:
1. This changes where the final upload occurs, from a call stack
   perspective. This is so the Zipkin traces look "right".
2. This removes the duplicated upload logic from the final
   upload function, again, so we can get a per-file breakdown
   of uploads.

### DIFF
--- a/executor/runner/runner.go
+++ b/executor/runner/runner.go
@@ -326,6 +326,7 @@ func (r *Runner) doShutdown(ctx context.Context, lastUpdate update) { // nolint:
 
 	ctx, span := trace.StartSpan(ctx, "doShutdown")
 	defer span.End()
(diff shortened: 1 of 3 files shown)


executor/runner/runner.go
```diff
@@ -326,6 +326,7 @@ func (r *Runner) doShutdown(ctx context.Context, lastUpdate update) { // nolint:
 
 	ctx, span := trace.StartSpan(ctx, "doShutdown")
 	defer span.End()
+	span.AddAttributes(trace.BoolAttribute("wasKiled", r.wasKilled()))
```
solarkennedy, line 329: thank you for this, it is helpful

alekar (approved):

PR #674, opened 2021-06-30
No reset generation

sargun:


solarkennedy (approved):
Do you think "3" is the last generation here? You don't think we'll need room for 4?

sargun:
Hopefully we wont need 4....but I'm not ruling it out. The thing is all new trunk ENIs have generation set at the time of creation.

PR #675, opened 2021-06-30
When IPv6 is enabled, use IPv6 AWS DNS server

sargun:


executor/runtime/docker/docker.go
```diff
@@ -391,6 +391,14 @@ func (r *DockerRuntime) mainContainerDockerConfig(c runtimeTypes.Container, bind
 		Runtime: c.Runtime(),
 	}
 
+	if r.c.AssignIPv6Address() {
+		if r.c.SeccompAgentEnabledForNetSyscalls() {
+			hostCfg.DNS = []string{"fd00:ec2::253"}
+		} else {
+			hostCfg.DNS = []string{"fd00:ec2::253", "169.254.169.253"}
```
solarkennedy, line 398: This will impact existing dual-stack titus apps in production, which is fine?

solarkennedy (approved):

executor/runtime/docker/docker.go
sargun: It's fine. I think.

PR #678, opened 2021-07-07
Transition NS Work

sargun:
This is the first commit of adding the transition namespace related code.

I'm going to wait for @solarkennedy to merge https://github.com/Netflix/titus-executor/pull/617 and then merge this after. Right now, this is somewhat of a placeholder PR.


sargun:
Service side TODOs so far:
* Finish server side GC / freeing of transition namespaces
* Make sure GC code doesn't accidentally clean up transition namespaces in soft GC mode

vpc/proto/vpc.proto
```diff
@@ -157,6 +159,13 @@ message AssignIPResponseV3 {
         // burst is the HTB class's ceil in bits per second
         uint64 burst = 2;
     }
+    message TransitionAssignment {
+        string assignmentId = 1;
+        // TODO: Add mark.
```
solarkennedy, line 164: What kind of mark?

vpc/service/assign_addresses_v3.go
```diff
@@ -668,45 +679,64 @@ func addDefaultRoute(ctx context.Context, jumbo bool, ass *vpcapi.AssignIPRespon
 			},
 		}
 	}
+
+	if ass.TransitionAssignment != nil {
+		// Transition assignment takes a slightly conservative stance
+		ass.TransitionAssignment.Routes = []*vpcapi.AssignIPResponseV3_Route{
+			{
+				Destination: "0.0.0.0/0",
+				Mtu:         1500,
```
solarkennedy, line 688: Can we just do the same jumbo logic with v4 as we do with v6?

Or does the fact that it is shared mean we must stick to a common denominator ?

vpc/service/assign_addresses_v3.go
```diff
@@ -719,21 +749,25 @@ func (vpcService *vpcService) assignIPsToENI(ctx context.Context, req *vpcapi.As
 		_ = tx.Rollback()
 	}()
 
-	err = lockAssignment(ctx, tx, ass.assignmentID)
+	_, err = lockAssignment(ctx, tx, ass.assignmentID)
```
solarkennedy, line 752: It's weird to me that `lockAssignment` now returns and assignment, but we don't care about it?
Unless it is a transition assignment, then we do? What is up with that?

vpc/service/db/migrations/31_transition_assignments.up.sql
```diff
@@ -0,0 +1,22 @@
+START TRANSACTION;
+alter table assignments
+    add is_transition_assignment bool default false not null;
+
+alter table assignments
+    add transition_assignment bigint;
+
+-- This prevents GC of assignment, but if the association is deleted
```
solarkennedy, line 8: Heh, I don't know enough sql to know if this is a good idea.

solarkennedy (commented):
Ready to rebase

vpc/service/assign_addresses_v3.go
sargun: lock assignment always queried the assignment. It also keeps a pessimistic lock on it for non-serializable transactions. This prevents two people from trying to assign IPs to it. It returns a "special" form of assignment (in case there's a race here). I don't really want to create *another* function that does the same thing (locks the assignment), so it seems like a good place to do two birds with one stone.

vpc/proto/vpc.proto
sargun: I decided the best way to handle bandwidth limiting would be to set SO_MARK on the sockets. The TSA would need to know on a container-by-container basis the mark to use when it injects the socket. 

This is a longer-term: TODO.

vpc/service/assign_addresses_v3.go
sargun: The latter.

vpc/service/assign_addresses_v3.go
```diff
@@ -668,45 +679,64 @@ func addDefaultRoute(ctx context.Context, jumbo bool, ass *vpcapi.AssignIPRespon
...
+		ass.TransitionAssignment.Routes = []*vpcapi.AssignIPResponseV3_Route{
+			{
+				Destination: "0.0.0.0/0",
+				Mtu:         1500,
+				Family:      vpcapi.AssignIPResponseV3_Route_IPv4,
+			},
+			{
+				Destination: "::/0",
+				Mtu:         1500,
+				Family:      vpcapi.AssignIPResponseV3_Route_IPv6,
```
solarkennedy, line 694: Why do we need v6 routes on the transition interface?

vpc/service/assign_addresses_v3.go
```diff
@@ -786,10 +823,41 @@ func (vpcService *vpcService) assignIPsToENI(ctx context.Context, req *vpcapi.As
...
+		resp.Ipv4Address, err = assignArbitraryIPv4AddressV3(ctx, tx, iface, maxIPAddresses, ass.subnet.cidr, ass.branch, ass.branchENISession)
+		if err != nil {
+			span.SetStatus(traceStatusFromError(err))
+			return nil, err
+		}
+	case *vpcapi.AssignIPRequestV3_TransitionRequested:
+		resp.TransitionAssignment = &vpcapi.AssignIPResponseV3_TransitionAssignment{
+			AssignmentId: transitionAss.assignmentID,
+		}
+		if transitionAss.ipv4addr.Valid {
```
solarkennedy, line 835: What is this case here, that the transition interface already has an IP?

vpc/tool/assign3/assign_network.go
```diff
@@ -179,6 +182,8 @@ func doAllocateNetwork(ctx context.Context, instanceIdentityProvider identity.In
 		}
 	} else if shouldAssignV4 {
 		assignIPRequest.Ipv4 = &vpcapi.AssignIPRequestV3_Ipv4AddressRequested{Ipv4AddressRequested: true}
+	} else if args.NetworkMode == titus.NetworkConfiguration_Ipv6AndIpv4Fallback.String() {
```
solarkennedy, line 185: woot!

solarkennedy (approved):

vpc/service/assign_addresses_v3.go
sargun: Right, because we can race between "associating" the assignment with a transition NS (and allocating that transition ns), and assigning an IP to it. If two assignments simultaneously generate a lock to the same assignment NS and one adds the IP, the other doesn't need to readd the IP.

vpc/service/assign_addresses_v3.go
sargun: This is laziness :(

PR #679, opened 2021-07-07
Rip out / deprecate old non-soft GC code

sargun:
This removes all code related to the old hard-gc bit.

It also fixes protobuf compilation of vpc.proto


vpc/proto/vpc.proto
```diff
@@ -226,11 +226,12 @@ message GCRequestV3 {
...
     repeated string runningTaskIDs = 2;
     // Soft indicates that do not perform the actual GC, only return the task IDs that would be GC'd
-    bool soft = 3;
+    bool soft = 3 [deprecated = true];
 }
 
 message GCResponseV3 {
-    repeated string removedAssignments = 1;
+    // assignmentsToRemove returns which assignments *should be* removed.
+    repeated string assignmentsToRemove = 1;
```
solarkennedy, line 234: Whoa I didn't know you could rename a field like this.
I guess if the number stays the same it is OK?

solarkennedy (approved):

vpc/proto/vpc.proto
sargun: You can even do stuff like go from uint32 -> uint64. Protobuf is wild.

PR #682, opened 2021-07-08
Mount /run/metatron for containers as a shared tmpfs, so that all containers in a pod can use it

solarkennedy:
This is to enable a shared /run for all containers in a pod,
mostly for sharing metatron credentials.

The thing is, '/run' is too much. We'll have all sorts of conflicts
over pid files, locks, logs, etc if we share the whole thing.

This change *only* shares /run/metatron to the other containers.

I was hoping that I could implement Country&Western in such a way
that the existing behavior of the main container remains unchanged,
but the tmpfs stuff is unavoidable. Hopefully when we get to the "stage1" approach,
this runtime dir will just be reused.


alekar (approved):

solarkennedy:
Tested this a bit in real life, added the explicit `/run/metatron` mount (not *all* of /run).
Also, because it is VK creating this tmpfs (and not docker), that means *vk* must clean up!

A new failure mode I could be potentially introducing would be failing to clean up the tmpfs when done.
If someone kill -9's VK or something, these tmpfs will be left behind :(
Still though, no sensitive material will be left on disk.

solarkennedy:
Per @JackOfMostTrades 's guidance, I've added `/metatron` (static certs) to also be shared. It doesn't have anything sensitive, so it doesn't need to live in a tmpfs, and can be shared via the normal means.

alekar (approved):

executor/runtime/docker/docker.go
```diff
@@ -1645,23 +1708,35 @@ func (r *DockerRuntime) k8sContainerToDockerConfigs(v1Container v1.Container, ma
...
-			{
-				Type:     "bind",
-				Source:   mainContainerRoot + "/logs",
-				Target:   "/logs",
-				ReadOnly: false,
-			},
-			{
-				Type:     "bind",
-				Source:   mainContainerRoot + "/run",
-				Target:   "/run",
```
alekar, line 1658: @solarkennedy do we not need this anymore? It seems like we do?
```
	// Always setup tmpfs: it's needed to ensure Metatron credentials don't persist across reboots and for SystemD to work
	hostCfg.Tmpfs = map[string]string{
		"/run": "rw,exec,size=" + defaultRunTmpFsSize,
	}```

solarkennedy:
Let me test this a little more thoroughly on my dev stack first.

solarkennedy:
Confirmed, this works fine. For whatever reason we've been able to get metatron+systemd working since forever.

solarkennedy:
If for some reason this starts causing trouble with leftover junk, we can instead change to depend on the /run tmpfs of the host (because these end up in /run/titus-executor anyway) if we wanted to, and clean up via systemd instead of via titus-executor cleanup hooks.

PR #683, opened 2021-07-09
Made `/titus/run/pod.json` the "Downward API"

solarkennedy:
We need to be very sure this is the API we want to support,
it is hard to take back once we do it.

I do like the idea of files though.

This change adds `/titus/run/pod.json`, as a bind mount
to the container. It does get updated as VK changes it,
which is useful for processes inside that need to ask
"are we healthy?".

----

While this "works", the file cannot be read because outside the container it is 600 because we use `tempFile`:

```
// The file's permissions will be 0600 by default. You can change these by
// explicitly calling Chmod on the returned PendingFile.
```

If shipit, I'll make a PR to VK to do 644.

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -13,6 +13,7 @@ import (
 	"net"
 	"os"
 	"os/exec"
+	"path"
 	"path/filepath"
 	"regexp"
 	"runtime"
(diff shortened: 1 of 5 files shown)


alekar:
I think this file should be `0400` for all processes in the pod, since it is being written by the executor?

solarkennedy:
> I think this file should be `0400` for all processes in the pod, since it is being written by the executor?

I can do that, I'll have to do an explicit `0400` and then `chown 10000`. Apparmor should also prevent it from being written to, and the bind mount is RO.

solarkennedy:
> I think this file should be `0400` for all processes in the pod, since it is being written by the executor?

Wait, can you clarify, when you say "should", do you mean you think it needs to be due to security desires? (even though that may not be the current way it works)
Or when you say "should", do you mean "this is the current way I believe it works" ?

alekar:
This:

> Wait, can you clarify, when you say "should", do you mean you think it needs to be due to security desires? (even though that may not be the current way it works)

solarkennedy:
Added https://github.com/Netflix-Skunkworks/titus-virtual-kubelet/pull/179 to tighten the permissions on the pod.json file.

alekar:
👍

PR #684, opened 2021-07-10
Refactor: Always assume we have a pod object to work with

solarkennedy:
In practice, we always have a pod. Long gone are the days
where we don't have a pod object at all.

But, the difference is whether we have a v0 pod or a v1 pod.

In this refactor, I removed all places where a pod was nil, which is
not a realistic reflection of the code we run in real life.

And in some places I added a "v0" pod in its place. I'm worried that
all the places where I *needed* a v0 pod indicates that we have a bug in our
podContainer implementation.

These todos can be addressed on the next pass.

Rob did a great job of doing cinfo/pod tests on the standalone stuff,
but this is in the types test area, which doesn't have the duplication.

### DIFF
--- a/executor/runner/runner.go
+++ b/executor/runner/runner.go
@@ -178,9 +178,6 @@ func (r *Runner) prepareContainer(ctx context.Context) update {
 }
 
 func getContainerNames(pod *v1.Pod) []string {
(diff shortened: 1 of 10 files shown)


solarkennedy:
This is ready for review, tests pass.

alekar (approved):

PR #686, opened 2021-07-11
Remove need for IAM Proxy

sargun:
This adds the ability for the metadataserver to run without a backing
server. If the environment variables are specified for AZ, region, and
AZ ID, then it'll use those as the "source of truth".

### DIFF
--- a/metadataserver/server.go
+++ b/metadataserver/server.go
@@ -144,6 +144,10 @@ func NewMetaDataServer(ctx context.Context, config types.MetadataServerConfigura
 		signer:                    config.Signer,
 		tokenRequired:             config.RequireToken,
 		xForwardedForBlockingMode: config.XFordwardedForBlockingMode,
+
+		region:             config.Region,
+		availabilityZone:   config.AvailabilityZone,
+		availabilityZoneID: config.AvailabilityZoneID,
 	}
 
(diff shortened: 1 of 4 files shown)


metadataserver/server.go
```diff
@@ -146,6 +146,10 @@ func NewMetaDataServer(ctx context.Context, config types.MetadataServerConfigura
 		signer:                    config.Signer,
 		tokenRequired:             config.RequireToken,
 		xForwardedForBlockingMode: config.XFordwardedForBlockingMode,
+
+		region:             config.Region,
+		availabilityZone:   config.AvailabilityZone,
+		availabilityZoneID: config.AvailabilityZoneID,
```
solarkennedy, line 152: I don't see this ever used?

solarkennedy (approved):

metadataserver/server.go
sargun: Here: https://github.com/Netflix/titus-executor/blob/124f2faba23f6702a123ad44cc0b8fb43273e3cc/metadataserver/server.go#L419-L474

solarkennedy:
This is good stuff. I vote we make this thing *always* use vars instead of proxying locally.

PR #689, opened 2021-07-12
Added basic SSL authn/z on the titus IAM service

solarkennedy:
This change addes CN-based authn/z for the Titus iam service.

It makes it so that the client (titus-imds) must have a valid
`titus-agent` certificate to connect to the titus iam service.

Another PR for the titus-agent will be made to set the right
netflix config to point this code at the actual location
of our certs/keys/cas.

### DIFF
--- a/root/lib/systemd/system/titus-iam-service.service
+++ b/root/lib/systemd/system/titus-iam-service.service
@@ -4,6 +4,7 @@ Conflicts=halt.target shutdown.target sigpwr.target
 
 [Service]
 ExecStart=/apps/titus-executor/bin/titus-iam-service
+EnvironmentFile=-/etc/titus-executor/iam-service.env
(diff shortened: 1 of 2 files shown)


root/lib/systemd/system/titus-iam-service.service
```diff
@@ -4,6 +4,7 @@ Conflicts=halt.target shutdown.target sigpwr.target
 
 [Service]
 ExecStart=/apps/titus-executor/bin/titus-iam-service
+EnvironmentFile=-/etc/titus-executor/iam-service.env
```
gabrielhartmann, line 7: What's using this?

gabrielhartmann (approved):

solarkennedy:
Discarding as we've decided to go with the CA approach

PR #692, opened 2021-07-14
Added docker event handling for multi-container pods (C&W)

solarkennedy:
This wires together the docker events with multi-containers.

Previously, we only looked at the 'main' container for any event.
This meant that if a sidecar died, nobody would know, nothing would
happen.

Now, all docker events for all containers are treated equally.
If one container fails, the whole task will fail.

Long term we will make this more nuanced and configurable,
especially for platform vs user sidecars.


executor/runtime/docker/docker.go
```diff
@@ -1408,9 +1401,30 @@ func (r *DockerRuntime) Start(parentCtx context.Context, pod *v1.Pod) (string, *
 		return "", nil, statusMessageChan, err
 	}
 
+	for _, c := range r.c.ExtraPlatformContainers() {
+		entry.Debugf("Adding status mmonitor for %s container (cid %s)", c.Name, c.ID)
+		eventChan, eventErrChan := r.getDockerEventsChannelsForContainer(eventCtx, c.ID)
+		go r.statusMonitor(eventCancel, c.ID, eventChan, eventErrChan, statusMessageChan)
```
sargun, line 1407: Shouldn't we keep track of these using something like a group?

executor/runtime/docker/docker.go
```diff
@@ -1436,10 +1464,13 @@ func (r *DockerRuntime) statusMonitor(cancel context.CancelFunc, c runtimeTypes.
 		// 4. Else, keep sending messages until we bail
 		select {
 		case err := <-errChan:
+			if err.Error() == "context canceled" {
```
sargun, line 1467: errors.is(err, context.Canceled)

executor/runtime/docker/docker.go
```diff
@@ -1408,9 +1401,30 @@ func (r *DockerRuntime) Start(parentCtx context.Context, pod *v1.Pod) (string, *
...
 	}
 
+	for _, c := range r.c.ExtraPlatformContainers() {
+		entry.Debugf("Adding status mmonitor for %s container (cid %s)", c.Name, c.ID)
+		eventChan, eventErrChan := r.getDockerEventsChannelsForContainer(eventCtx, c.ID)
+		go r.statusMonitor(eventCancel, c.ID, eventChan, eventErrChan, statusMessageChan)
+	}
+	for _, c := range r.c.ExtraUserContainers() {
+		entry.Debugf("Adding status mmonitor for %s container (cid %s)", c.Name, c.ID)
+		eventChan, eventErrChan := r.getDockerEventsChannelsForContainer(eventCtx, c.ID)
```
sargun, line 1411: Is there going to be many getDockerEventsChannelsForContainers? Should we instead have one channel, and multiplex it? Anyone know about the internals of how Docker events work? CC: @alekar

solarkennedy: Great question. 
The docs do say that having multiple filters on a stream (multiple container ids for us) is treated like an OR.
Having a single stream is probably cheaper for docker to do, I think that is a good idea.

The code that has to read the events has to decode "which" container emitted it anyway, might as well be a single event stream.

solarkennedy: Yea I'm going to go with the 1 channel + 1 docker event stream approach

executor/runtime/docker/docker.go
solarkennedy: Not super sure. We don't want to wait for all monitors to finish, in fact we kinda want to fail as soon as any one of them finish and send along the status update.

solarkennedy:
I'm compromising here with a middle ground approach:

* One docker event stream for the main container - the reason we need this is because we know the docker containerid way way early in the lifecycle, and I like the idea of watching for docker events for it way early, before we start any of the sidecars.
* One docker event stream for all other containers - We can start this stream after we `create` all other containers. We don't even make a second stream if we only have one container!

executor/runtime/docker/docker.go
```diff
@@ -1501,37 +1543,57 @@ func getMainContainerRoot(inspectOutput types.ContainerJSON) string {
...
+	}
+	for _, c := range r.c.ExtraUserContainers() {
+		ids = append(ids, c.Status.ContainerID)
+	}
+	return ids
+}
 
-	// This context should be tied to the lifetime of the container -- it will get significantly less broken
-	// when we tear out the launchguard code
+func isClosed(ch <-chan runtimeTypes.StatusMessage) bool {
```
solarkennedy, line 1557: This is wrong.
I'm not using this channel correctly (reading from it when testing if it is closed).

I'll redo.

executor/runtime/docker/docker.go
```diff
@@ -1484,15 +1478,60 @@ func (r *DockerRuntime) Start(parentCtx context.Context, pod *v1.Pod) (string, *
...
 	}
 
 	return logDir, details, statusMessageChan, err
 }
 
+func (r *DockerRuntime) getDockerEventsChannelsForContainers(eventCtx context.Context, containerIDs []string) (<-chan events.Message, <-chan error) {
+	filters := filters.NewArgs()
+	for _, containerID := range containerIDs {
+		filters.Add("container", containerID)
+		filters.Add("type", "container")
```
alekar, line 1527: Shouldn't this be outside the loop?

alekar (approved):

executor/runtime/docker/docker.go
solarkennedy: I don't think so. Seems to work correctly while testing. 
Per the docs (at least with cli):
https://docs.docker.com/engine/reference/commandline/events/#limiting-filtering-and-formatting-the-output
Not sure if it is redundant or not. The tests should tell though, let me see for sure.

solarkennedy:
Rebased on master and made a better test that looks for the *exact* exit code of the extra container.

PR #696, opened 2021-07-21
Fix docker_fuse apparmor profile

solarkennedy:
When I copy/pasted the docker_titus profile, I didn't update
the strings that referenced profile names to whatever the "self" was.
(in this case, `docker_fuse` and not `docker_titus`)

### DIFF
--- a/root/etc/apparmor.d/docker_fuse
+++ b/root/etc/apparmor.d/docker_fuse
@@ -43,8 +43,8 @@ profile docker_fuse flags=(attach_disconnected,mediate_deleted) {
 
   deny /titus/** wl,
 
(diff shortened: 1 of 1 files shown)


root/etc/apparmor.d/docker_fuse
```diff
@@ -98,5 +98,4 @@ profile docker_fuse flags=(attach_disconnected,mediate_deleted) {
 
     /etc/{gai.conf,group,nsswitch.conf,passwd,shadow} rmkl,
   }
-}
-
+}
```
sargun, line 101: End with newline please.

sargun (approved):

PR #700, opened 2021-07-22
Add per-job ssh policy principal to ssh config when an ssh policy is defined in metatron app metadata

JackOfMostTrades:
When an SSH policy is set in the app metadata, add the corresponding principal to the SSH trusted principals configuration.

I think this should be safe since since it only applies if the new metadata key is set and appends to existing principals.

### DIFF
--- a/executor/runtime/docker/docker_ssh.go
+++ b/executor/runtime/docker/docker_ssh.go
@@ -135,7 +135,7 @@ func addContainerSSHDConfigWithData(c runtimeTypes.Container, tw *tar.Writer, cf
 		log.Warn("The NETFLIX_ENVIRONMENT variable is not set. SSH access to the container may not be available!")
 	}
 
-	sshPolicyHash, err := getContainerSshPolicyHash(c)
+	sshPolicyHash, err := getContainerSSHPolicyHash(c)
 	if err != nil {
 		log.Warnf("Failed to get container SSH policy hash: %v", err)
 	}
@@ -169,7 +169,7 @@ func addContainerSSHDConfigWithData(c runtimeTypes.Container, tw *tar.Writer, cf
 	return nil
 }
 
-func getContainerSshPolicyHash(c runtimeTypes.Container) (string, error) {
+func getContainerSSHPolicyHash(c runtimeTypes.Container) (string, error) {
 	metatronCreds := c.MetatronCreds()
 	if metatronCreds == nil {
 		return "", nil
--- a/executor/runtime/docker/docker_ssh.go
+++ b/executor/runtime/docker/docker_ssh.go
@@ -2,8 +2,11 @@ package docker
 
 import (
 	"archive/tar"
+	"crypto/sha256"
+	"encoding/base64"
 	"fmt"
 	"io/ioutil"
+	"net/url"
 	"strings"
 
 	"github.com/Netflix/titus-executor/config"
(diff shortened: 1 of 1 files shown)


executor/runtime/docker/docker_ssh.go
```diff
@@ -157,3 +168,28 @@ func addContainerSSHDConfigWithData(c runtimeTypes.Container, tw *tar.Writer, cf
...
+	sshPolicy := params.Get("sshPolicy")
+	if sshPolicy == "" {
+		return "", nil
+	}
+	h := sha256.New()
+	_, err = h.Write([]byte(sshPolicy))
+	if err != nil {
+		return "", fmt.Errorf("Failed to hash ssh policy: %v", err)
+	}
+	return base64.StdEncoding.EncodeToString(h.Sum(nil)), nil
```
solarkennedy, line 194: Wait, why do we need to base64 a shasum? It should be ascii-safe?

executor/runtime/docker/docker_ssh.go
```diff
@@ -157,3 +168,28 @@ func addContainerSSHDConfigWithData(c runtimeTypes.Container, tw *tar.Writer, cf
...
+func getContainerSSHPolicyHash(c runtimeTypes.Container) (string, error) {
+	metatronCreds := c.MetatronCreds()
+	if metatronCreds == nil {
+		return "", nil
+	}
+	appMetadata := metatronCreds.AppMetadata
+	if appMetadata == nil && *appMetadata == "" {
+		return "", nil
+	}
+	params, err := url.ParseQuery(*appMetadata)
```
solarkennedy, line 181: Whoa, the appmetadata is a url query string?

executor/runtime/docker/docker_ssh.go
```diff
@@ -140,6 +148,9 @@ func addContainerSSHDConfigWithData(c runtimeTypes.Container, tw *tar.Writer, cf
 		}
 		if containerEnv != "" {
 			lines = append(lines, fmt.Sprintf("~v3:titus:%s:%s:%s:%s:%s:%s", username, c.AppName(), containerEnv, c.TaskID(), c.JobGroupStack(), c.JobGroupDetail()))
+			if sshPolicyHash != "" {
+				lines = append(lines, fmt.Sprintf("~v3:instance:%s:%s:%s", username, c.TaskID(), sshPolicyHash))
```
solarkennedy, line 152: If this becomes widespread, we might want something to make it easier for you and me to debug, like:
    
    policyComment := fmt.Sprintf("# This next principal allows %s", sshPolicy)
    policyPrincipal :=  fmt.Sprintf("~v3:instance:%s:%s:%s", username, c.TaskID(), sshPolicyHash)
    lines = append(lines, policyComment, policyPrincipal)

executor/runtime/docker/docker_ssh.go
JackOfMostTrades: Yea, this is being added to the authorized principals file. I could make this a bit more clear and return the []byte hash here and do the encoding when it's inserted into the authorized principals.

solarkennedy (approved):

executor/runtime/docker/docker_ssh.go
JackOfMostTrades: Fair enough. I'll make that update.

executor/runtime/docker/docker_ssh.go
JackOfMostTrades: Yup. Probably would have suggested json if we were doing this again from scratch, but it's a simple enough way to encode a bunch of key/value pairs.

solarkennedy (approved):

russell-lewis (approved):

PR #702, opened 2021-08-03
Add support for IPv6 NFS mounts

MyUmmaGumma:
Fix for TITUS-5627

Lookup for valid v4 and v6 and then do getaddrinfo and parse results

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -1860,7 +1860,7 @@ func (r *DockerRuntime) setupEFSMounts(parentCtx context.Context, c runtimeTypes
 		mountOptions := append(
 			baseMountOptions,
 			fmt.Sprintf("fsc=%s", c.TaskID()),
-			fmt.Sprintf("source=%s:%s", nfs.Server, nfs.ServerPath),
+			fmt.Sprintf("source=[%s]:%s", nfs.Server, nfs.ServerPath),
 		)
 		cmd.Env = []string{
 			fmt.Sprintf("MOUNT_TARGET=%s", nfs.MountPoint),
--- a/mount/titus-mount-nfs.c
+++ b/mount/titus-mount-nfs.c
@@ -29,52 +29,66 @@ char nfs4[] = "nfs4";
 		}                                                              \
 	} while (0)
 
-static int dns_lookup(const char *hostname, struct sockaddr_in *addr)
-{
-	struct hostent *hp;
-	addr->sin_family = AF_INET;
+#define IP_ADDRESS_LEN 256
 
-	if (inet_aton(hostname, &addr->sin_addr)) {
-		fprintf(stderr,
-			"titus-mount-nfs: %s is already an IP. Not doing a DNS lookup\n",
-			hostname);
-		return 0;
-	}
-	fprintf(stderr,
-		"titus-mount-nfs: Decoding and resolving dns hostname for %s\n",
-		hostname);
-	hp = gethostbyname(hostname);
-	if (hp == NULL) {
-		int err_ret = h_errno;
-		fprintf(stderr, "titus-mount-nfs: can't get address for %s: %s\n",
-			hostname, hstrerror(err_ret));
-		return -1;
-	}
-	if (hp->h_length > (int)sizeof(struct in_addr)) {
-		fprintf(stderr, "titus-mount-nfs: got bad hp->h_length");
-		return -1;
-	}
-	memcpy(&addr->sin_addr, hp->h_addr, hp->h_length);
+static int add_addr_options(const char *hostname, char *final_options)
+{
+	strcat(final_options, ",addr=");
+	strcat(final_options, hostname);
 	return 0;
 }
 
-static void compose_final_options(const char *nfs_mount_hostname,
-				  char *final_options)
+/*
+ * Perform a DNS lookup and add to the DNS options. Prefer v6 over v4 
+ * (rely on /etc/gai.conf)
+ */
+static int hostname_to_ip_option(const char *hostname, char *final_options)
 {
-	/* For NFS, we must do the dns resolution *here* while we are inside net ns */
-	struct sockaddr_in server_addr;
-	char *ip_string;
+	struct addrinfo *result, *iter;
+	int error;
+	unsigned char buf[sizeof(struct in6_addr)];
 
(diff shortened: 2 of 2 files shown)


mount/titus-mount-nfs.c
```diff
@@ -29,52 +29,66 @@ char nfs4[] = "nfs4";
...
-		fprintf(stderr, "titus-mount-nfs: can't get address for %s: %s\n",
-			hostname, hstrerror(err_ret));
-		return -1;
-	}
-	if (hp->h_length > (int)sizeof(struct in_addr)) {
-		fprintf(stderr, "titus-mount-nfs: got bad hp->h_length");
-		return -1;
-	}
-	memcpy(&addr->sin_addr, hp->h_addr, hp->h_length);
+static int add_addr_options(const char *hostname, char *final_options)
```
sargun, line 34: Any reason this isn't a void?

mount/titus-mount-nfs.c
```diff
@@ -29,52 +29,66 @@ char nfs4[] = "nfs4";
...
+ * (rely on /etc/gai.conf)
+ */
+static int hostname_to_ip_option(const char *hostname, char *final_options)
 {
-	/* For NFS, we must do the dns resolution *here* while we are inside net ns */
-	struct sockaddr_in server_addr;
-	char *ip_string;
+	struct addrinfo *result, *iter;
+	int error;
+	unsigned char buf[sizeof(struct in6_addr)];
```
sargun, line 49: Shouldn't this be buf[INET6_ADDRSTRLEN]?

mount/titus-mount-nfs.c
```diff
@@ -308,7 +322,11 @@ int main(int argc, char *argv[])
 	/* Now we can switch net/mount namespaces so we can lookup the ip and eventually mount */
 	switch_namespaces(pidfd);
 	fprintf(stderr, "titus-mount-nfs: user-inputed options: %s\n", options);
-	compose_final_options(nfs_mount_hostname, final_options);
+	res = hostname_to_ip_option(nfs_mount_hostname, final_options);
+	if (res != 0) {
+		fprintf(stderr, "Error resolving hostname to IP address on mount");
```
sargun, line 327: Add \n at the end.

mount/titus-mount-nfs.c
```diff
@@ -29,52 +29,66 @@ char nfs4[] = "nfs4";
...
+	} 
+
+	error = inet_pton(AF_INET6, hostname, buf);
+	if (error == 1) {
+		fprintf(stderr, "Valid IPv6 address %s - No DNS needed", hostname);
+		add_addr_options(hostname, final_options);
+		return 0;
+	} 
+
+	fprintf(stderr, "Now trying to do getaddrinfo %s: ", gai_strerror(error));
```
sargun, line 65: I think that this should be:

`fprintf(stderr, "Now trying to do getaddrinfo: %s\n", hostname)`

mount/titus-mount-nfs.c
```diff
@@ -29,52 +29,66 @@ char nfs4[] = "nfs4";
...
+	error = getaddrinfo(hostname, NULL, NULL, &result);
+	if (error) {
+		fprintf(stderr, "Error trying to resolve %s: ", gai_strerror(error));
```
sargun, line 68: fprintf(stderr, "Error trying to resolve %s: %s", hostname, gai_strerror(error));

executor/runtime/docker/docker.go
```diff
@@ -1860,7 +1860,7 @@ func (r *DockerRuntime) setupEFSMounts(parentCtx context.Context, c runtimeTypes
 		mountOptions := append(
 			baseMountOptions,
 			fmt.Sprintf("fsc=%s", c.TaskID()),
-			fmt.Sprintf("source=%s:%s", nfs.Server, nfs.ServerPath),
+			fmt.Sprintf("source=[%s]:%s", nfs.Server, nfs.ServerPath),
```
solarkennedy, line 1863: Is putting brackets on the outside here safe to do, safe to do, even if it is a regular hostname or ipv4?

mount/titus-mount-nfs.c
MyUmmaGumma: Fixed.

mount/titus-mount-nfs.c
MyUmmaGumma: Fixed.

mount/titus-mount-nfs.c
MyUmmaGumma: Fixed.

mount/titus-mount-nfs.c
MyUmmaGumma: removed.

mount/titus-mount-nfs.c
MyUmmaGumma: Fixed.

executor/runtime/docker/docker.go
MyUmmaGumma: Yes, I tested the IPv4 and seemed to work fine. The unit tests tested against the EFS DNS endpoint which also worked.

mount/titus-mount-nfs.c
```diff
@@ -29,52 +29,64 @@ char nfs4[] = "nfs4";
...
+		return -1;
+	} 
+
+	for (iter = result; iter != NULL; iter = iter->ai_next) {
+		char ipaddress[IP_ADDRESS_LEN];
+		void *ptr_to_ip;
+		inet_ntop(iter->ai_family, iter->ai_addr->sa_data, ipaddress, IP_ADDRESS_LEN);
+		switch (iter->ai_family) {
+			case AF_INET:
+				ptr_to_ip = &((struct sockaddr_in *)iter->ai_addr)->sin_addr;
```
solarkennedy, line 76: Is this safe? What if the container doesn't have a v6 address?

MyUmmaGumma: Code wise its safe but as you pointed out for a Ipv4Only container, this code would prefer resolving to a v6 address and the connection will run into problems. That scenario will require thought where we might want to retry mounting onto the IPv4 resolved address were the Ipv6 mount fail due to network unreachability.

sargun:
Don't all your fprintf's need a \n at the end?

sargun (approved):

PR #703, opened 2021-08-04
Make sure /titus/etc/ssh (dirs) have perms 755

sargun:
If /titus, or /titus/etc, or /titus/etc/ssh have world
writable permissions, SSH refuses to use them for authz.

In newer versions of Docker, if the directory is not explicitly
part of the tarball, it create it with 777 perms.

### DIFF
--- a/executor/runtime/docker/docker_ssh.go
+++ b/executor/runtime/docker/docker_ssh.go
@@ -103,6 +103,14 @@ func addContainerSSHDConfig(c runtimeTypes.Container, tw *tar.Writer, cfg config
 }
 
 func addContainerSSHDConfigWithData(c runtimeTypes.Container, tw *tar.Writer, cfg config.Config, caData []byte) error {
+	if err := tw.WriteHeader(&tar.Header{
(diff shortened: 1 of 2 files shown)


executor/runtime/docker/docker_ssh.go
```diff
@@ -103,6 +103,14 @@ func addContainerSSHDConfig(c runtimeTypes.Container, tw *tar.Writer, cfg config
 }
 
 func addContainerSSHDConfigWithData(c runtimeTypes.Container, tw *tar.Writer, cfg config.Config, caData []byte) error {
+	if err := tw.WriteHeader(&tar.Header{
+		Name:     "titus/etc/ssh",
```
JackOfMostTrades, line 107: Probably doesn't matter, but the `Name:` in the old code below has a leading `/` but `Name:` here does not.

sargun: The "/" gets stripped off by the Tar implementation.

alekar (approved):

JackOfMostTrades (approved):

PR #707, opened 2021-08-12
Support concurrent file uploads with deduplication logic

MyUmmaGumma:
Container shutdown is slow because of serialized file uploads. 
Add logic to make them concurrent with a limit of 8 go routines per upload. Also, add some deduplication logic to avoid duplicate file uploads.

The changes to upload.go have not been tested as we don't use the directory upload logic yet. I've left similar changes to limit concurrent uploads for whenever we decide to use that logic.

I ran a cursory test to check the number of worker threads when the worker threads exceed the files to be uploaded. I haven't run a test to see if the [slow container shutdown](https://jira.netflix.net/browse/TITUS-4144?filter=-1) is solved with this change.

### DIFF
--- a/filesystems/watcher.go
+++ b/filesystems/watcher.go
@@ -23,7 +23,6 @@ import (
 	"github.com/Netflix/titus-executor/filesystems/xattr"
 	"github.com/Netflix/titus-executor/uploader"
 	"github.com/Netflix/titus-executor/vpc/tracehelpers"
-	ccache "github.com/karlseguin/ccache/v2"
 	log "github.com/sirupsen/logrus"
 	"go.opencensus.io/trace"
 	"golang.org/x/sync/errgroup"
@@ -39,8 +38,7 @@ const (
 	// StdioAttr is the attribute that states this is a stdio, rotatable file
 	StdioAttr                    = "user.stdio"
 	waitForDieAfterContextCancel = time.Minute
-	concurrentUploaders          = 3
-	dedupCacheExpirationTime     = time.Second * 10
+	concurrentUploaders          = 8
 )
 
 var (
@@ -281,7 +279,7 @@ func (w *Watcher) traditionalRotate(ctx context.Context) {
 	logFileList, err := buildFileListInDir(w.config.localDir, true, w.config.uploadThresholdTime)
 	if err == nil {
 		for _, logFile := range logFileList {
-			_ = w.uploadLogfile(ctx, logFile, false, nil)
+			_ = w.uploadLogfile(ctx, logFile, false)
 		}
 	} else {
 		log.Error(err)
@@ -565,7 +563,7 @@ func parsecurrentOffsetBytes(name string, currentOffsetBytes []byte) int64 {
 	return currentOffsetInt
 }
 
-func (w *Watcher) concurrentUploadLogFile(ctx context.Context, logFileList []string, dedupCache *ccache.Cache) error {
+func (w *Watcher) concurrentUploadLogFile(ctx context.Context, logFileList []string) error {
 	// How many workers do we need
 	uploadWorkers := concurrentUploaders
 	if len(logFileList) < uploadWorkers {
@@ -585,7 +583,7 @@ func (w *Watcher) concurrentUploadLogFile(ctx context.Context, logFileList []str
 		g.Go(func() error {
 			defer sem.Release(1)
 			if !CheckFileForStdio(fileToUploadCopy) {
-				if err := w.uploadLogfile(ctx, fileToUploadCopy, true, dedupCache); err != nil {
+				if err := w.uploadLogfile(ctx, fileToUploadCopy, true); err != nil {
 					return err
 				}
 				log.Error("Uploaded concurrently ", fileToUploadCopy)
@@ -604,7 +602,7 @@ func (w *Watcher) concurrentUploadLogFile(ctx context.Context, logFileList []str
 
 // uploadAllLogFiles is called to upload all of the files in the directories
 // being watched.
-func (w *Watcher) uploadAllLogFiles(ctx context.Context, dedupCache *ccache.Cache) error {
+func (w *Watcher) uploadAllLogFiles(ctx context.Context) error {
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
@@ -619,7 +617,17 @@ func (w *Watcher) uploadAllLogFiles(ctx context.Context, dedupCache *ccache.Cach
 		return err
 	}
 
-	err = w.concurrentUploadLogFile(ctx, logFileList, dedupCache)
+	//Remove duplicate entries in the list of files to upload
+	filenames := make(map[string]bool)
+	uniqueFiles := []string{}
+	for _, filename := range logFileList {
+		if _, exists := filenames[filename]; !exists {
+			filenames[filename] = true
+			uniqueFiles = append(uniqueFiles, filename)
+		}
+	}
+
+	err = w.concurrentUploadLogFile(ctx, uniqueFiles)
 	tracehelpers.SetStatus(err, span)
 
 	return err
@@ -633,7 +641,7 @@ type uploaded struct {
 // uploadLogFile is called to upload a single log file while the
 // task is running. Currently, no error is returned to the caller,
 // it is just logged.
-func (w *Watcher) uploadLogfile(ctx context.Context, fileToUpload string, finalization bool, dedupCache *ccache.Cache) error {
+func (w *Watcher) uploadLogfile(ctx context.Context, fileToUpload string, finalization bool) error {
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
@@ -666,26 +674,6 @@ func (w *Watcher) uploadLogfile(ctx context.Context, fileToUpload string, finali
 		tracehelpers.SetStatus(err, span)
 		log.WithError(err).Error("Error uploading")
 		return err
-	}
-
-	if dedupCache != nil {
-		//Check if we uploaded this file before
-		elem := dedupCache.Get(fileToUpload)
-		if elem == nil {
-			dedupCache.Set(fileToUpload, uploaded{stat.ModTime(), stat.Size()}, dedupCacheExpirationTime)
-		} else {
-			prev := elem.Value().(*uploaded)
-			if prev.size < stat.Size() || prev.mtime.Before(stat.ModTime()) {
-				prev.size = stat.Size()
-				prev.mtime = stat.ModTime()
-				dedupCache.Set(fileToUpload, prev, dedupCacheExpirationTime)
-			} else {
-				err = fmt.Errorf("Duplicate file upload for %q", fileToUpload)
-				tracehelpers.SetStatus(err, span)
-				log.WithError(err).Error("Error uploading")
-				return err
-			}
-		}
 	}
 
 	err = w.uploader.Upload(ctx, fileToUpload, path.Join(w.config.uploadDir, remoteFilePath), xattr.GetMimeType)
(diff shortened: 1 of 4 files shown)


filesystems/watcher.go
```diff
@@ -561,10 +561,10 @@ func parsecurrentOffsetBytes(name string, currentOffsetBytes []byte) int64 {
 	return currentOffsetInt
 }
 
-func (w *Watcher) concurrentUploadLogFile(logFileList []string, ctx context.Context) error {
+func (w *Watcher) concurrentUploadLogFile(ctx context.Context, logFileList []string) error {
 	// Iterate over each file and setup the work
 	var wg sync.WaitGroup
-	const CONUCRRENT_UPLOADERS = 8
```
sargun: Please define this at the top of the file.

filesystems/watcher.go
```diff
@@ -561,6 +561,55 @@ func parsecurrentOffsetBytes(name string, currentOffsetBytes []byte) int64 {
...
+			err := w.uploadLogfile(ctx, logFile, true)
+			errs = multierror.Append(errs, err)
+		}
+	}
+	// Start the workers
+	for i := 0; i < uploadWorkers; i++ {
+		go uploadWork(uploadFileC)
+	}
+
+	wg.Wait()
```
sargun, line 604: So, a couple things:
1. One of the things you can use here is an errgroup: https://pkg.go.dev/golang.org/x/sync/errgroup - It will encapsulate the logic of tying together context cancellation and returning errors. There is also go-multierror's "Group": https://pkg.go.dev/github.com/hashicorp/go-multierror#Group
2. You cannot perform the line: 
```go
				errs = multierror.Append(errs, ctx.Err())
```

Safely, because it'll result in a concurency violation.

filesystems/watcher.go
```diff
@@ -701,10 +738,28 @@ func buildFileListInDir2(dirName string, fileList []string, checkModifiedTimeThr
 				continue
 			}
 
+			elem, ok := dedupUpload[fileInfo.Name()]
```
sargun, line 741: I think that it'd be simpler to use a cache for the dedupe upload storage. We use github.com/karlseguin/ccache/v2 elsewhere in the code.

You can have the key be something like:
{Filename, Ctime}, and the value be {Size, Mtime}
And if size, mtime changes, or the lookup in cache is a miss, then upload.

sargun: Also, IMHO, the dedupe logic should live in watcher.uploadLogfile.

uploader/uploader.go
```diff
@@ -79,41 +86,59 @@ func uploadDir(ctx context.Context, uploader Backend, local string, remote strin
 		if err != nil {
 			errs = multierror.Append(errs, err)
 		} else {
-			uploadErrs := make([]error, len(finfos))
-
-			// Iterate over each file and upload it
+			log.WithField("file", local).Info("is a directory - to read files")
```
sargun, line 89: Do we need to make normal uploads concurrent, given that we only care about slow container shutdown, can this logic only happen in uploadAllLogFiles?

uploader/uploader.go
```diff
@@ -125,10 +150,11 @@ func (e *Uploader) Upload(ctx context.Context, local, remote string, ctypeFunc C
 	}
 
 	if fi.IsDir() {
+		log.WithField("Going to upload ", local).WithField("to remote", remote).Info(" - directory")
```
sargun, line 153: Your with field here shoudl just have the ("local", local). It's not a "message", but a structured log entry

sargun (changes requested):

uploader/uploader.go
```diff
@@ -65,9 +65,16 @@ func NewUploaderWithBackend(backend Backend) *Uploader {
 	return &Uploader{backend}
 }
 
+type uploadFileInfo struct {
+	src  string
+	dest string
+}
+
 // Upload of all of files in a directory but not its subdirectories. Performs the uploads in parallel.
```
solarkennedy, line 73: This `uploadDir`, it looked like it was parallized already? 
Is this change just to cap the worker count of the uploaders?

filesystems/watcher.go
MyUmmaGumma: Thanks for surfacing this! Using errgroup now.

filesystems/watcher.go
MyUmmaGumma: I'm using ccache now but chose to omit ctime because it requires separate system calls to get the ctime and didn't see a benefit over using size and mtime (could be convinced otherwise).

MyUmmaGumma: Also moved the dedupe logic to uploadLogFile.

uploader/uploader.go
MyUmmaGumma: Since the upload/uploader.go change is not used currently I would not worry about this being exercised right now. As it is, it is already concurrent but the concurrency is not limited so I am planning on leaving the change there.

uploader/uploader.go
MyUmmaGumma: Removed.

uploader/uploader.go
MyUmmaGumma: Correct. As I noted on the PR - this code is not used by the uploadLogFiles on container.Stop() - I've left these changes in to cap the concurrency when this code does get used. But for now watcher.go is where the uploads happen.

sargun (changes requested):
So, I think if you revert the changes in uploadDir, that'd be good, because right now I think we're reviewing two separate paths, but right now, IMHO, this is a little too confusing to me.

filesystems/watcher.go
```diff
@@ -721,13 +721,13 @@ func (w *Watcher) uploadLogfile(ctx context.Context, fileToUpload string, finali
 		//Check if we uploaded this file before
 		elem := dedupCache.Get(fileToUpload)
 		if elem == nil {
-			dedupCache.Set(fileToUpload, uploaded{stat.ModTime(), stat.Size()}, dedupCacheExpirationSec)
+			dedupCache.Set(fileToUpload, uploaded{stat.ModTime(), stat.Size()}, dedupCacheExpirationTime)
```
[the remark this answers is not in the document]
MyUmmaGumma: Added code to remove duplicates from the slice of log files to be uploaded. Removed the dedup logic.

filesystems/watcher.go
```diff
@@ -40,7 +40,7 @@ const (
 	StdioAttr                    = "user.stdio"
 	waitForDieAfterContextCancel = time.Minute
 	concurrentUploaders          = 3
-	dedupCacheExpirationSec      = time.Second * 10
+	dedupCacheExpirationTime     = time.Second * 10
```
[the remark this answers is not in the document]
MyUmmaGumma: Removed.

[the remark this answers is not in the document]
MyUmmaGumma: Also reverted changes to uploader/upload.go

filesystems/watcher.go
```diff
@@ -561,6 +563,43 @@ func parsecurrentOffsetBytes(name string, currentOffsetBytes []byte) int64 {
...
+		fileToUploadCopy := fileToUpload
+		err := sem.Acquire(ctx, 1)
+		if err != nil {
+			continue
+		}
+
+		g.Go(func() error {
+			defer sem.Release(1)
+			if !CheckFileForStdio(fileToUploadCopy) {
+				if err := w.uploadLogfile(ctx, fileToUploadCopy, true); err != nil {
```
sargun, line 586: I think it's better to use multierror.Group. The reason being is that if one of the uploaders fails, you don't want to cancel all of them:
https://pkg.go.dev/github.com/hashicorp/go-multierror#Group

filesystems/watcher.go
```diff
@@ -561,6 +563,43 @@ func parsecurrentOffsetBytes(name string, currentOffsetBytes []byte) int64 {
...
+					return err
+				}
+				log.Error("Uploaded concurrently ", fileToUploadCopy)
```
sargun, line 589: why is this a log.Error?

filesystems/watcher.go
```diff
@@ -561,6 +563,43 @@ func parsecurrentOffsetBytes(name string, currentOffsetBytes []byte) int64 {
...
+
+			}
+			return nil
+		})
+	}
+
+	if err := g.Wait(); err != nil {
+		log.Error("Error during concurrent upload of files: g.Wait() err ", err)
```
sargun, line 597: Use log.WithError(err)...

filesystems/watcher.go
```diff
@@ -578,27 +617,19 @@ func (w *Watcher) uploadAllLogFiles(ctx context.Context) error {
...
-	for _, logFile := range logFileList {
-		if ctx.Err() != nil {
-			errs = multierror.Append(errs, ctx.Err())
-			break
-		}
-		if CheckFileForStdio(logFile) {
-			continue
+	//Remove duplicate entries in the list of files to upload
+	filenames := make(map[string]bool)
+	uniqueFiles := []string{}
```
sargun, line 622: Do we actually get dupes here? Also, any reason not to use sets:
https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets

MyUmmaGumma: Since this is the entire list of files we upload, IMO if we catch it here we should not have any duplicates. Catching duplicates now with sets.

filesystems/watcher.go
MyUmmaGumma: done

filesystems/watcher.go
MyUmmaGumma: removed - debugging artifact

filesystems/watcher.go
MyUmmaGumma: Done.

filesystems/watcher.go
```diff
@@ -561,6 +565,63 @@ func parsecurrentOffsetBytes(name string, currentOffsetBytes []byte) int64 {
...
+
+func (w *Watcher) concurrentUploadLogFile(ctx context.Context, logFileList []string) error {
+	safeError := SafeMultiError{}
+	// How many workers do we need
+	uploadWorkers := concurrentUploaders
+	if len(logFileList) < uploadWorkers {
+		uploadWorkers = len(logFileList)
+	}
+
+	g, ctx := errgroup.WithContext(ctx)
```
sargun, line 591: Drop this line.

sargun (changes requested):

filesystems/watcher.go
```diff
@@ -561,6 +565,63 @@ func parsecurrentOffsetBytes(name string, currentOffsetBytes []byte) int64 {
...
+				if err := w.uploadLogfile(ctx, fileToUploadCopy, true); err != nil {
+					log.WithError(err).Errorf("Error during concurrent upload of files")
+					safeError.AppendErr(err)
+				}
+			}
+			return nil
+		})
+	}
+
+	if err := g.Wait(); err != nil {
```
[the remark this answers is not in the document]
MyUmmaGumma: Done

filesystems/watcher.go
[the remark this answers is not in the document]
MyUmmaGumma: Done.

filesystems/watcher.go
```diff
@@ -561,6 +565,63 @@ func parsecurrentOffsetBytes(name string, currentOffsetBytes []byte) int64 {
...
+	// How many workers do we need
+	uploadWorkers := concurrentUploaders
+	if len(logFileList) < uploadWorkers {
+		uploadWorkers = len(logFileList)
+	}
+
+	g, ctx := errgroup.WithContext(ctx)
+	// Limit concurrency
+	sem := semaphore.NewWeighted(int64(uploadWorkers))
+	for _, fileToUpload := range logFileList {
```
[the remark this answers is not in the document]
MyUmmaGumma: Done.

sargun (approved):

PR #710, opened 2021-08-17
Add assignNextIPv6Address

sargun:
This makes it so that we allocate addresses from the "bottom" /80 of the IPv6 space in a given subnet.

### DIFF
--- a/vpc/service/assign_addresses_v3.go
+++ b/vpc/service/assign_addresses_v3.go
@@ -1623,3 +1623,53 @@ func assignSpecificIPv4AddressV3(ctx context.Context, tx *sql.Tx, branchENI *ec2
 		PrefixLength: uint32(prefixlength),
 	}, nil
 }
+
+func assignNextIPv6Address(ctx context.Context, tx *sql.Tx, subnetid string) (net.IP, error) {
+	sequenceName := fmt.Sprintf("v6seq_%s", strings.ReplaceAll(subnetid, "-", "_"))
+	_, err := tx.ExecContext(ctx, fmt.Sprintf("CREATE SEQUENCE IF NOT EXISTS %s MINVALUE 0 MAXVALUE 65534 CACHE 10 CYCLE", sequenceName))
+	if err != nil {
+		return nil, fmt.Errorf("Cannot create sequence %s: %w", sequenceName, err)
+	}
+
+	/*
+			The address layout looks like:
+
+		 ┌───────────────────────────┬──────────────┬───────────────────────────┬────────────────────┐
+		 │ AWS Subnet CIDR (64-bits) │ 0s (16 bits) │ Time in Seconds (32-bits) │ Sequence (16-bits) │
+		 └───────────────────────────┴──────────────┴───────────────────────────┴────────────────────┘
+
+	*/
+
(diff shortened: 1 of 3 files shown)


MyUmmaGumma (approved):

vpc/service/assign_addresses_v3.go
```diff
@@ -1623,3 +1623,53 @@ func assignSpecificIPv4AddressV3(ctx context.Context, tx *sql.Tx, branchENI *ec2
 		PrefixLength: uint32(prefixlength),
 	}, nil
 }
+
+func assignNextIPv6Address(ctx context.Context, tx *sql.Tx, subnetid string) (net.IP, error) {
+	sequenceName := fmt.Sprintf("v6seq_%s", strings.ReplaceAll(subnetid, "-", "_"))
+	_, err := tx.ExecContext(ctx, fmt.Sprintf("CREATE SEQUENCE IF NOT EXISTS %s MINVALUE 0 MAXVALUE 65534 CACHE 10 CYCLE", sequenceName))
```
solarkennedy, line 1629: At first glance this looks wasteful to do on every new IP. 
Is there not some existing procedure for adding new subnets to the vpc service? 
Or is that something we are trying to get away from?


If there was, then this CREATE SEQ could be part of the "add subnet to titus procedure", then we wouldn't have to call it every time we make add an IP?
Then again, I don't know how expensive this is, it is just "one" db call and it does nothing 99% of the time.

vpc/service/assign_addresses_v3.go
```diff
@@ -1623,3 +1623,53 @@ func assignSpecificIPv4AddressV3(ctx context.Context, tx *sql.Tx, branchENI *ec2
...
+	/*
+			The address layout looks like:
+
+		 ┌───────────────────────────┬──────────────┬───────────────────────────┬────────────────────┐
+		 │ AWS Subnet CIDR (64-bits) │ 0s (16 bits) │ Time in Seconds (32-bits) │ Sequence (16-bits) │
+		 └───────────────────────────┴──────────────┴───────────────────────────┴────────────────────┘
+
+	*/
+
+	row := tx.QueryRowContext(ctx, "SELECT nextval($1)", sequenceName)
```
solarkennedy, line 1643: Why do the unix time at all if we already are making the DB do the work of accounting for the sequence?

solarkennedy (approved):

vpc/service/assign_addresses_v3.go
sargun: Because you're not guaranteed to have this number always increment, and you generally want to do what you can to spread it out. For example:

```sql
START TRANSACTION;
SELECT nextval('foo');
ROLLBACK;
START TRANSACTION;
SELECT nextval('foo'); -- It will return the same value
```

vpc/service/assign_addresses_v3.go
sargun: This is a temporary thing. It's relatively cheap, and takes a shared read lock out most of the time. Honestly, though, this is "low hanging fruit" optimization, and we can do all sorts of tricks, like swapping around the sequence number and putting it first, and using micros for the LSB.

PR #714, opened 2021-08-18
Send the effective networkMode back to the control-plane

solarkennedy:
We only get 1 ipv4 and 1 ipv6 to send back to the control-plane
in the form of the pod status.

In the IPv6Only+transition mode, that ipv4 represents the transition ip,
and should not be considered the "normal" ip for the pod.

For this reason, it is helpful to send the effective network-mode
back to the control-plane with the rest of the network configuration
updates, so that it can make better decisions about what data (ip addresses)
it is getting.

Pairs with
https://github.com/Netflix/titus-control-plane/pull/1086

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -1447,6 +1447,7 @@ func (r *DockerRuntime) Start(parentCtx context.Context, pod *v1.Pod) (string, *
 			IsRoutableIP: true,
 			IPAddress:    ipv4addr.Address.Address,
 			EniIPAddress: ipv4addr.Address.Address,
+			NetworkMode:  r.c.EffectiveNetworkMode(),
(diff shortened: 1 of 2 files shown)


sargun (approved):

sargun:
I know it's too late to do this, but it'd be nice to use protobuf -- at some point, on the messages sent back to the control plane

solarkennedy:
> I know it's too late to do this, but it'd be nice to use protobuf -- at some point, on the messages sent back to the control plane

Are you sure for this type of communication you want to use protobuf? Compared to the current way which is by annotating the pod with strings and letting the shared informer get the update.

solarkennedy:
If you want, I can also slowly change these annotations to have more modern fully qualified names on the pod?

PR #715, opened 2021-08-18
Remove extra TASK_STARTING updates to the control-plane

solarkennedy:
I went a little overboard before sending all these updates
back to the control-plane. In practice I don't think they are needed,
they take up extra processing on the controllers and space in cassandra.

On failure though, they are helpful, and we still get them.
We should still log the things as normal.

On success, I think we only need One (1) TASK_STARTING update.

### DIFF
--- a/executor/runner/runner.go
+++ b/executor/runner/runner.go
@@ -147,7 +147,7 @@ type update struct {
 }
 
 func (r *Runner) prepareContainer(ctx context.Context) update {
-	logger.G(ctx).Debug("Running prepare")
+	logger.G(ctx).Debug("Running prepare on main container")
 	prepareCtx, prepareCancel := context.WithCancel(ctx)
 	defer prepareCancel()
 	go func() {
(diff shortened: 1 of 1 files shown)


executor/runner/runner.go
```diff
@@ -162,7 +162,7 @@ func (r *Runner) prepareContainer(ctx context.Context) update {
 	// These steps may or may not have completed depending on if/where a failure occurred.
 	if err := r.runtime.Prepare(prepareCtx); err != nil {
 		r.metrics.Counter("titus.executor.launchTaskFailed", 1, nil)
-		logger.G(ctx).Error("Task failed to create container: ", err)
+		logger.G(ctx).Error("Task failed to create main container: ", err)
```
sargun, line 165: Use WithError...

sargun (approved):
lgtm

corindwyer (approved):

tbak (approved):

PR #717, opened 2021-08-20
Fix cleanup of IPv6 BPF map

sargun:
There was a bug on how the key was derived for the IPv6 HTB classid
mapping BPF map.

### DIFF
--- a/vpc/tool/container2/setup_container_linux.go
+++ b/vpc/tool/container2/setup_container_linux.go
@@ -486,29 +486,13 @@ func updateBPFMap(ctx context.Context, mapName string, key []byte, value uint16)
 
 func updateBPFMaps(ctx context.Context, assignment *vpcapi.AssignIPResponseV3) error {
 	if assignment.Ipv4Address != nil {
-		buf := make([]byte, 8)
-
-		binary.LittleEndian.PutUint16(buf, uint16(assignment.VlanId))
-		ip := net.ParseIP(assignment.Ipv4Address.Address.Address).To4()
-		if len(ip) != 4 {
-			panic("Length of IP is not 4")
-		}
-		copy(buf[4:], ip)
-		if err := updateBPFMap(ctx, "ipv4_map", buf, uint16(assignment.ClassId)); err != nil {
+		if err := updateBPFMap(ctx, "ipv4_map", ipv4key(assignment), uint16(assignment.ClassId)); err != nil {
 			return err
 		}
 	}
 
 	if assignment.Ipv6Address != nil {
-		buf := make([]byte, 20)
-		binary.LittleEndian.PutUint16(buf, uint16(assignment.VlanId))
-		ip := net.ParseIP(assignment.Ipv6Address.Address.Address).To16()
-		if len(ip) != 16 {
-			panic("Length of IP is not 16")
-		}
-
-		copy(buf[4:], ip)
-		if err := updateBPFMap(ctx, "ipv6_map", buf, uint16(assignment.ClassId)); err != nil {
+		if err := updateBPFMap(ctx, "ipv6_map", ipv6key(assignment), uint16(assignment.ClassId)); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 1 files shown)


vpc/tool/container2/setup_container_linux.go
```diff
@@ -745,25 +727,59 @@ func deleteFromIPv6BPFMap(ctx context.Context, assignment *vpcapi.AssignIPRespon
 	defer unix.Close(int(fd))
 	runtime.KeepAlive(path)
 
+	key := ipv6key(assignment)
```
solarkennedy, line 730: is the `buf` even used anymore in this function?

vpc/tool/container2/setup_container_linux.go
```diff
@@ -745,25 +727,59 @@ func deleteFromIPv6BPFMap(ctx context.Context, assignment *vpcapi.AssignIPRespon
...
 		logger.G(ctx).WithError(errors.Wrapf(err, "Unable to delete element for map with file descriptor %d", fd)).Error()
-		err2 := errors.Wrap(err, "Unable to delete element for ipv4 map")
+		err2 := errors.Wrap(err, "Unable to delete element for ipv6 map")
 		return err2
 	}
 	runtime.KeepAlive(uba)
 	return nil
 }
 
+func ipv6key(assignment *vpcapi.AssignIPResponseV3) []byte {
```
solarkennedy, line 754: good stuff, making these functions

vpc/tool/container2/setup_container_linux.go
MyUmmaGumma, line 730: Should we be checking if key is nil - we seem to be accessing key[0] next.
(Same with the v6key below)

solarkennedy:
Since the ipv6 key is changing (but I guess wasn't working in the first place), are you worried about leftover ipv6 keys on existing agents, after this new code is deployed mutably? (or are you thinking they will get auto-remediated like they are today when the map fills up anyway)

vpc/tool/container2/setup_container_linux.go
```diff
@@ -745,25 +727,59 @@ func deleteFromIPv6BPFMap(ctx context.Context, assignment *vpcapi.AssignIPRespon
...
 	)
-	if err != unix.ENOENT && err != 0 {
+	if err == unix.ENOENT {
+		err := errors.Wrap(err, "Unable to delete element for ipv6 map (notfound)")
+		logger.G(ctx).WithError(err).WithField("ip", assignment.Ipv6Address.Address.Address).WithField("classid", assignment.ClassId).WithField("vlanid", assignment.VlanId).Error("Element not found")
+		return err
+	} else if err != 0 {
 		logger.G(ctx).WithError(errors.Wrapf(err, "Unable to delete element for map with file descriptor %d", fd)).Error()
-		err2 := errors.Wrap(err, "Unable to delete element for ipv4 map")
+		err2 := errors.Wrap(err, "Unable to delete element for ipv6 map")
```
solarkennedy, line 747: Do we need the err2 pattern? Can you just return `errors.wrap` ?

vpc/tool/container2/setup_container_linux.go
```diff
@@ -745,25 +727,59 @@ func deleteFromIPv6BPFMap(ctx context.Context, assignment *vpcapi.AssignIPRespon
...
+		return nil
+	}
+	buf := make([]byte, 8)
+
+	binary.LittleEndian.PutUint16(buf, uint16(assignment.VlanId))
+	ip := net.ParseIP(assignment.Ipv4Address.Address.Address).To4()
+	if len(ip) != 4 {
+		panic("Length of IP is not 4")
+	}
+	copy(buf[4:], ip)
```
MyUmmaGumma, line 779: Out of curiosity - What do we do with the first 4 bytes of this buffer eventually?

MyUmmaGumma (approved):

vpc/tool/container2/setup_container_linux.go
```diff
@@ -694,17 +671,22 @@ func deleteFromIPv4BPFMap(ctx context.Context, assignment *vpcapi.AssignIPRespon
 	defer unix.Close(int(fd))
 	runtime.KeepAlive(path)
```
solarkennedy, line 672: On a second pass, if you wanted to, I think you could DRY out the raw bpf map stuff with a function like:

    deleteKeyFromBPFMap(mapPath string, key []byte)

that would do the fd BPF_OBJ_GET, uba, BPF_MAP_DELETE_ELEM, and return just the error.

and make `"/sys/fs/bpf//tc/globals/ipv4_map\000"` consts 

Maybe. We are talking 2 copies of the code here, and my goto for refactoring is 3 copies. Up to you.

solarkennedy (approved):

vpc/tool/container2/setup_container_linux.go
sargun: VLAN ID / sg

vpc/tool/container2/setup_container_linux.go
sargun: because err2 is an "error" whereas err is an syscall.Errno.

PR #719, opened 2021-08-23
Create GC / transition NS deletion loop

sargun:


vpc/service/gc_transition_ns.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+	defer func() {
+		_ = tx.Rollback()
+	}()
+
+	n, err := tx.ExecContext(ctx, `
+DELETE FROM assignments
+WHERE is_transition_assignment
+AND transition_last_used < now() - $1 * INTERVAL '1 second'
+AND created_at < now() - INTERVAL '1 minute'
```
solarkennedy, line 60: Hehe, I don't know how you parse this in your head, but are you sure it isn't easier to put it as:

    now() - created_at >  INTERVAL '1 minute'
    (age) > 1min

solarkennedy (approved):

vpc/service/gc_transition_ns.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	err = tx.Commit()
+	if err != nil {
+		return err
+	}
+
+	logger.G(ctx).WithField("n", n).Info("GC'd transition assignments")
+	return nil
+}
```
solarkennedy, line 76: This task just deletes from the database. 
Is this the normal happy path for this? 

No right? Because normally `SELECT transition_assignment FROM assignments` is going to have everything that is up and normal, and anything that is leftover from there is going to be somehow lost through the cracks and somehow manually deleted or something?

To put it another way, what process cleans up transition NS's on the agent?

solarkennedy (approved):

vpc/service/gc_transition_ns.go
sargun: Yeah. Working on the part for the agents still. Probably should have opened this as a draft PR.

vpc/service/gc_transition_ns.go
```diff
@@ -0,0 +1,76 @@
...
+	"fmt"
+	"time"
+
+	"github.com/Netflix/titus-executor/logger"
+	"github.com/Netflix/titus-executor/vpc/tracehelpers"
+	"go.opencensus.io/trace"
+)
+
+const (
+	timeBetweenGCTransitionNS = 2 * time.Minute
```
MyUmmaGumma, line 14: How do you determine the time intervals for GC?

solarkennedy:
@MyUmmaGumma  / @aloktiagi  please take a look at this PR and begin thinking about adding it in?

solarkennedy:
Or @hechaoli can you take over this PR?
I know you have been working on the transition ns stuff.

hechaoli:
Transition name space GC which includes both the namespace on the instance and the record in the VPC service DB have been implemented in #944. This can be closed.

PR #725, opened 2021-08-31
Refactor the statusMessageChan update stuff

solarkennedy:
The statusMessageChan is a channel that transports updates from
a the docker runtime's health to the runner.

As we move towards multiple containers, we are going to need
to get updates for different containers, any of which can oom, die,
etc.

This refactor prepares the statusMessageChan machinery for multiple
containers, particularly away the closing of the channel as a method
to trigger titusdriver.Lost (should never happen).

----

Factoring out this change, related to https://github.com/Netflix/titus-executor/pull/692, to make it easier to have a discussion about, or hopefully make it less controversial.

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -1515,12 +1515,18 @@ func (r *DockerRuntime) statusMonitor(cancel context.CancelFunc, containerID str
 			if errors.Is(err, context.Canceled) {
 				return
 			}
-			log.Fatalf("Got unexpected error while listening for docker events for %s, bailing: %s", containerID, err)
+			log.WithError(err).Errorf("Got unexpected error while listening for docker events for %s, bailing: %s", containerID, err)
+			return
 		case event := <-eventChan:
 			log.Debug("Got docker event: ", event)
 			isTerminalEvent := r.handleDockerEvent(event, statusMessageChan)
 			if isTerminalEvent {
-				log.Infof("Closing docker status monitor for %s because terminal docker event received", r.getContainerNameFromID(containerID))
+				cName, err := r.getContainerNameFromID(containerID)
+				if err != nil {
+					log.Infof("Closing docker status monitor for %s because terminal docker event received", cName)
+				} else {
+					log.WithError(err).Error("Closing docker status monitor, encountered and error looking up the container name")
+				}
 				return
 			}
 		}
@@ -1875,7 +1881,6 @@ func (r *DockerRuntime) getPlaformContainerNames() []string {
 // This function returns true if the handling of docker events
 // should stop.
 func (r *DockerRuntime) handleDockerEvent(message events.Message, statusMessageChan chan runtimeTypes.StatusMessage) bool {
-	validateMessage(message)
 	action := strings.Split(message.Action, " ")[0]
 	action = strings.TrimRight(action, ":")
 	l := log.WithFields(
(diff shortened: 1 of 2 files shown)


executor/runtime/docker/docker.go
```diff
@@ -1887,58 +1891,115 @@ func handleDockerEvent(c runtimeTypes.Container, message events.Message, statusM
...
+func (r *DockerRuntime) getContainerNameFromDockerEvent(m events.Message) string {
+	return r.getContainerNameFromID(m.ID)
+}
+
+func (r *DockerRuntime) getContainerNameFromID(id string) string {
+	// Most common case, just the main container, we return the string "main"
+	if id == r.c.ID() {
+		return mainContainerName
+	}
+	r.c.ExtraPlatformContainers()
```
alekar, line 1987: Is this a typo?

executor/runtime/docker/docker.go
```diff
@@ -1887,58 +1891,115 @@ func handleDockerEvent(c runtimeTypes.Container, message events.Message, statusM
...
+		}
 	}
-
-	return true
+	for _, c := range r.c.ExtraUserContainers() {
+		if id == c.Status.ContainerID {
+			return c.Name
+		}
+	}
+	return "Unknown container"
```
alekar, line 1998: This seems to be an error condition, perhaps return error here and log the entire message before marking it "unknown container"?

executor/runtime/docker/docker.go
```diff
@@ -1503,23 +1506,21 @@ func getMainContainerRoot(inspectOutput types.ContainerJSON) string {
...
 	for {
 		// 3. If the current state of the container is terminal, send it, and bail
 		// 4. Else, keep sending messages until we bail
 		select {
 		case err := <-errChan:
-			log.Fatal("Got error while listening for docker events, bailing: ", err)
+			if errors.Is(err, context.Canceled) {
+				return
+			}
+			log.Fatalf("Got unexpected error while listening for docker events for %s, bailing: %s", containerID, err)
```
alekar, line 1518: Maybe we should log error here and return. Why kill the executor with fatal?

executor/runtime/docker/docker.go
solarkennedy: yes. I'm surprised that didn't trigger any lint

executor/runtime/docker/docker.go
```diff
@@ -1887,58 +1891,115 @@ func handleDockerEvent(c runtimeTypes.Container, message events.Message, statusM
...
 }
 
 // The only purpose of this is to test the sanity of our filters, and Docker
-func validateMessage(c runtimeTypes.Container, message events.Message) {
-	if c.ID() != message.ID {
-		panic(fmt.Sprint("c.ID() != message.ID: ", message))
-	}
+func validateMessage(message events.Message) {
 	if message.Type != "container" {
 		panic(fmt.Sprint("message.Type != container: ", message))
```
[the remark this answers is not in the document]
solarkennedy: I'm not sure. It was there before so I didn't change it. Honestly I don't think it adds any value, so I'll remove it.

executor/runtime/docker/docker.go
solarkennedy: I agree.

alekar (approved):

solarkennedy:
Whoa, that last commit started to fail the `TestShutdown`.
That is pretty strange. I'm investigating...

solarkennedy:
Oh disregard, I introduced a real bug in the last commit. Yay tests.

PR #726, opened 2021-08-31
TitusMainContainerVolume mount support

solarkennedy:
I expect a common pattern in Titus will be to
"Share this volume between the main container".

This would power things like Fuse mounts from a fuse sidecar,
or just sharing /mydata to things like nginx or whatever.

Our experience when working with this kind of pattern in
stock kubelet is kinda painful, becuase it means making
and EmptyDir, making an init container to copy files over
from the main one to that dir, and then sharing that EmptyDir
volume with another container.

But we control the runtime, we can make the common thing easy,
and I bet 90% of the kinds of volumes we will want will be like this.

----

To use this feature, one (the titus api) would make a pod spec like this:

````
spec {
  "volumes": [{
    "name": "fuseVolume",
    "flexVolume":
      "driver": "TitusMainContainerVolume"
      "options": {
        "source:" "/fuse",
      },
  }],
  containers: [
    {
      "name": "main",
      "image": "customerImage",
      .....
    },
    {
      "name": "fuseSidecar",
      "image": "myFuseImage",
      "volumeMounts": [{
         "name": "fuseVolume",
         "mountPath": "/fuse",
         "readOnly": false,
       }]
    }      
  ]
}
````

And that would in turn make `/fuse` show up as `/fuse` in the sidecar, and it can do whatever it needs to, and they both see it! No emptydirs, external mounts, or tmpfs or anything like that. Ephemeral storage quota will still be respected.

### CODE REVIEW

**Comment by solarkennedy** [comment]
Pairs with: https://github.com/Netflix/titus-api-definitions/pull/145

File: executor/runtime/docker/docker.go [hunk 1]
```diff
@@ -1588,7 +1588,7 @@ func (r *DockerRuntime) createAllExtraContainers(ctx context.Context, pod *v1.Po
 	for idx := range r.c.ExtraUserContainers() {
 		c := r.c.ExtraUserContainers()[idx]
 		group.Go(func(ctx context.Context) error {
-			cid, err := r.createExtraContainerInDocker(ctx, c.V1Container, mainContainerID, mainContainerRoot)
+			cid, err := r.createExtraContainerInDocker(ctx, c.V1Container, mainContainerID, mainContainerRoot, *pod)
```
sargun, line 1591: Why not just pass by pointer here?

executor/runtime/docker/docker.go
```diff
@@ -1869,6 +1870,37 @@ func (r *DockerRuntime) getPlaformContainerNames() []string {
...
+	volumes := pod.Spec.Volumes
+	for _, volumeMount := range c.VolumeMounts {
+		v := getVolumeByName(volumes, volumeMount.Name)
+		if v.Name == "" {
+			continue
+		}
+		if v.FlexVolume.Driver == "TitusMainContainerVolume" {
+			m := mount.Mount{
+				Type:        "bind",
+				Source:      mainContainerRoot + v.FlexVolume.Options["source"],
```
sargun, line 1884: use filepath.join.

executor/runtime/docker/docker.go
```diff
@@ -1869,6 +1870,37 @@ func (r *DockerRuntime) getPlaformContainerNames() []string {
...
+				BindOptions: &mount.BindOptions{Propagation: mount.PropagationRShared},
+			}
+			mounts = append(mounts, m)
+		}
+	}
+	return mounts
+}
+
+func getVolumeByName(volumes []v1.Volume, name string) v1.Volume {
+	for _, v := range volumes {
```
sargun, line 1896: It might be easier to return *v1.Volume here.

sargun (approved):

executor/runtime/docker/docker.go
solarkennedy: I've fixed that in all these related function so it is always the pointer now.
I'm a tad paranoid about mutating the pod when I don't mean to.

executor/runtime/docker/docker.go
solarkennedy: Why though?

alekar (approved):

PR #727, opened 2021-09-02
Allow tini to start titus-metratrond

MyUmmaGumma:
In order for metatron client on Titus ipv6only containers in prod to connect to metatron servers and because of the way[ NLBs NAT IPv6 to IPv4 for NLBs,](https://docs.google.com/document/d/1A-ix_JXPDCYXYOfwJ_kzLC43OVaKTUbGX-3TFXgclkk/edit#), one of the needs is to get TSA to intercept connect requests for metatron client like it does for the container. 

In this diff we allow Tini to start metatron client so that the client can get its call intercepted by TSA.

Verified by changing the .service file manually on the agent.

This has the unintended side effect of metatron logs getting redirected to stdout and stderr. Working on that in the interim.

### DIFF
--- a/root/lib/systemd/system/titus-sidecar-metatron-sync@.service
+++ b/root/lib/systemd/system/titus-sidecar-metatron-sync@.service
@@ -9,8 +9,9 @@ PartOf=titus-container@%i.target
 
 [Service]
 EnvironmentFile=/var/lib/titus-environments/%i.env
+Environment="TINI_HANDOFF=true"
 # Run as root (UID 0, GID 0) and with CAP_DAC_OVERRIDE so that containers with a `USER` instruction work
-ExecStart=/usr/bin/runc --root /var/run/docker/runtime-${TITUS_OCI_RUNTIME}/moby exec --user 0:0 --cap CAP_DAC_OVERRIDE ${TITUS_CONTAINER_ID} /sbin/docker-init -s -- /titus/metatron/bin/titus-metatrond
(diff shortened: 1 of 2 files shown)


root/lib/systemd/system/titus-sidecar-metatron-sync@.service
```diff
@@ -10,7 +10,7 @@ PartOf=titus-container@%i.target
 [Service]
 EnvironmentFile=/var/lib/titus-environments/%i.env
 # Run as root (UID 0, GID 0) and with CAP_DAC_OVERRIDE so that containers with a `USER` instruction work
-ExecStart=/usr/bin/runc --root /var/run/docker/runtime-${TITUS_OCI_RUNTIME}/moby exec --user 0:0 --cap CAP_DAC_OVERRIDE ${TITUS_CONTAINER_ID} /titus/metatron/bin/titus-metatrond
+ExecStart=/usr/bin/runc --root /var/run/docker/runtime-${TITUS_OCI_RUNTIME}/moby exec --user 0:0 --cap CAP_DAC_OVERRIDE ${TITUS_CONTAINER_ID} /sbin/docker-init -s -- /titus/metatron/bin/titus-metatrond
```
solarkennedy, line 13: Can you add env variable overrides for stderr/stdout redirect off (and make tini to respect `""`) so that the output goes to the journal before?

Also can you add a HANDOFF=true always here, because there is no reason to keep tini in the process tree for this case

solarkennedy: (Oh you said that in the discription working on interim, carry on)

MyUmmaGumma: Changed

solarkennedy (approved):
Let's not forget to remove this when we don't need it!

PR #728, opened 2021-09-02
Added ?download=true option for downloading logs from the logviewer

solarkennedy:
This adds a new url option to set the content-disposition option for
logs. This header forces your browser to do a "save as" with the right
filename and such.

I haven't done any testing really.

cc @parshap

### DIFF
--- a/logviewer/api/handlers.go
+++ b/logviewer/api/handlers.go
@@ -30,11 +30,12 @@ func LogHandler(w http.ResponseWriter, r *http.Request) {
 	if fileName == "" {
 		fileName = "stdout"
 	}
+	addDownloadHeader := r.URL.Query().Get("download") == "true"
 
-	logHandler(w, r, containerID, fileName)
+	logHandler(w, r, containerID, fileName, addDownloadHeader)
 }
 
-func logHandler(w http.ResponseWriter, r *http.Request, containerID, fileName string) {
+func logHandler(w http.ResponseWriter, r *http.Request, containerID, fileName string, addDownloadHeader bool) {
 	containerLogsRoot := buildLogLocationBase(containerID)
 
 	filePath, err := securejoin.SecureJoin(containerLogsRoot, fileName)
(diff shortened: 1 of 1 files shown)


logviewer/api/handlers.go
```diff
@@ -30,11 +30,16 @@ func LogHandler(w http.ResponseWriter, r *http.Request) {
 	if fileName == "" {
 		fileName = "stdout"
 	}
+	addDownloadHeader := false
+	downloadOption := r.URL.Query().Get("download")
```
sargun, line 34: Couldn't this be:

```go
addDownloadHeader = r.URL.Query().Get("download") == "true"
```

logviewer/api/handlers.go
```diff
@@ -57,6 +62,12 @@ func logHandler(w http.ResponseWriter, r *http.Request, containerID, fileName st
 		return
 	}
 	defer shouldClose(fout)
+
+	if addDownloadHeader {
+		basename := filepath.Base(fileName)
+		w.Header().Set("Content-Disposition", "attachment; filename="+basename)
```
sargun, line 68: Is there any escaping necccessary? For example, if the filename has a ; in it?

sargun (approved):

logviewer/api/handlers.go
solarkennedy: Ah totally

logviewer/api/handlers.go
solarkennedy: Uhhh maybe?
https://stackoverflow.com/questions/93551/how-to-encode-the-filename-parameter-of-content-disposition-header-in-http

But modern docs:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition

Sounds like you just need quotes, but whoops, I didn't end the quote here!

But yea, sounds like it is a bit of a mess here, but it doesn't sound like I need to urlencode the name.

But it does sound like I need to escape quotemarks

solarkennedy: Fixed trailing quote and added a comment linking to the docs on this header

PR #729, opened 2021-09-03
Transition ns 5

sargun:
TODO: 
- [ ] GC Code - right now if a transition NS is added to an ENI, it will never be reused for different security groups nor detached
- [x] Code to move / propagate mounts from the "root" namespace
- [x] TSA Integration


vpc/tool/container2/setup_container_linux.go
```diff
@@ -121,6 +140,241 @@ func DoSetupContainer(ctx context.Context, netnsfd int, assignment *vpcapi.Assig
...
+	// We do this above because the previous fd gets closed via the defer.
+	fd, err := unix.Dup(int(ns))
+	if err != nil {
+		return 0, fmt.Errorf("Could not dup netns: %w", err)
+	}
+
+	return fd, nil
+}
+
+func crossMount(ctx context.Context, src *os.File, pid1dirfd int) error {
```
solarkennedy, line 336: Far out.

I'm trying to come up with a potentially better way to think about this.

Because fundamentally this thing gets an fd, and creates a new namespace in a ... dir.

Maybe there is a better set of inputs and outputs for this thing to do.

Till we come up with something better though, because we control the produce and consumer, I say ship.

vpc/tool/container2/setup_container_linux.go
```diff
@@ -121,6 +140,241 @@ func DoSetupContainer(ctx context.Context, netnsfd int, assignment *vpcapi.Assig
...
+		return 0, fmt.Errorf("Could not restore original network namespace: %w", err)
+	}
+
+	if newHandleErr != nil {
+		return 0, fmt.Errorf("Could not get netns handle: %w", err)
+	}
+	defer handle.Delete()
+
+	r := rand.New(rand.NewSource(time.Now().UnixNano())) // nolint: gosec
+	transitionInterfaceName := fmt.Sprintf("transition-%d", r.Intn(10000))
```
solarkennedy, line 285: neat!

solarkennedy: Kinda an optimization, but if we know the transition NS is always a unique set of SGs/subnet(?), which I guess that logic is in the vpc service itself, but it would be cool if we could give it a significant name.

Or at least you could do `transition-$assignmentID` ?

cmd/titus-vpc-tool/setup_container.go
```diff
@@ -15,17 +15,19 @@ func setupContainercommand(ctx context.Context, v *pkgviper.Viper, iipGetter ins
...
 			default:
 				return fmt.Errorf("Version %q not recognized", v.GetString(generationFlagName))
 			}
 		},
 	}
 
-	cmd.Flags().Int("netns", 3, "The File Descriptor # of the network namespace to setup")
+	cmd.Flags().Int("pid-1-dir-fd", 3, "The File Descriptor # of the pid 1 directory to setup")
 	cmd.Flags().String(generationFlagName, generationDefaultValue, "Generation of VPC Tool to use, specify v1, or v2")
+	cmd.Flags().String("transition-namespace-dir", "/run/transition", "Directory to mount transition namespaces into")
```
solarkennedy, line 31: `Directory in the container to mount transition namespaces into` ?
Actually, is it in the container? 
Or is it not, just the fd gets bind-mounted?

executor/runner/runner.go
```diff
@@ -23,6 +24,10 @@ import (
 	v1 "k8s.io/api/core/v1"
 )
 
+var (
+	errTaskKilledBeforeCreation = errors.New("Task was killed before Task was created")
```
solarkennedy, line 28: Maybe we could formalize the steps or phases that the titus-executor does, to make it more obvious what "create" means in this context.

vpc/tool/container2/setup_container_linux.go
sargun: That gets too long, interface names are limited to 14 or 15 characters.

sargun:
@solarkennedy Heads up, this makes the behaviour so that if we have a transition NS, we use that's IPv4 address to sendto the control plane.

MyUmmaGumma (approved):

vpc/tool/assign3/assign_network.go
```diff
@@ -75,6 +75,7 @@ func Assign(ctx context.Context, instanceIdentityProvider identity.InstanceIdent
 			err = errors.Wrap(writeError, err.Error())
 			return err
 		}
+		return err
```
solarkennedy, line 78: whoa whoops? I guess we were swallowing this error before?

executor/runtime/docker/docker.go
```diff
@@ -460,7 +460,8 @@ func (r *DockerRuntime) mainContainerDockerConfig(c runtimeTypes.Container, bind
 
 	// Always setup tmpfs: it's needed to ensure Metatron credentials don't persist across reboots and for SystemD to work
 	hostCfg.Tmpfs = map[string]string{
-		"/run": "rw,exec,size=" + defaultRunTmpFsSize,
+		"/run":       "rw,exec,size=" + defaultRunTmpFsSize,
+		"/run/netns": "rw,size=" + defaultRunTmpFsSize,
```
solarkennedy, line 464: Do we really need this to be *in* the container? Why can't it live in /var/run/titus-executor/ something?

solarkennedy: Just adds to the surface area of what is inside the container and changes the behavior on all containers (even non-transition ones).

inject/titus-mnt.c
```diff
@@ -0,0 +1,118 @@
+#define _GNU_SOURCE 1
```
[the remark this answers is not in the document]
sargun: titus-bind-mnt-ns?

executor/runtime/docker/docker.go
sargun: I mean, I'm trying to make it so that standard tools (like `ip netns`) will work with this. They expect /run/netns to be a tmpfs. I can put this behind a flag so it's only done on containers with transition mode?

solarkennedy: Oh. Huh. 
No this is fine I guess. As long as users don't depend on this.

I still think it is a little weird that we end up mounting this ns ... directory (?) *in* the container, only to titus-mnt it *back* out for titus-vpc-tool to run? (am I understanding it correctly?)

sargun: it's a file. So, the transition namespace is /run/netns/transition. This is also the way the TSA figures out which transition namespace to use for a given container.

PR #735, opened 2021-09-20
Rollback pd

sargun:
Undo Prefix Delegation
    
    This commit does two things:
     * Removes the logic to always assign a prefix to each branch ENI
     * Uses an alternative method to assign IPv6 addresses
    
    The way the logic works for assigning IPv6 addresses:
    * We introduce a new structure in the database, "worker_id". At
      the startup of each instance of the titusvpcservice, it gets
      a worker ID. There are only 4095 worker IDs available
    * Every time we assign an IPv6 address to a container, we use a
      a per-subnet incrementing counter
    * For a given subnet, we start the counter (when the first IP
      is assigned), at a random number.
    
    The risk here is if there is a random number collision, and the
    subnet is densely populated. If this happens, we may be able to
    alleviate the situation by restarting the service. One other
    choice is to hoist this to the database, but that involves /
    requires that we do at least one more DB write per assign.


solarkennedy:
Why introduce something new here (workerID) instead of doing a "true" rollback to the old way of doing things?

vpc/service/assign_addresses_v3.go
```diff
@@ -1115,34 +1116,138 @@ func getBorderGroupForENI(ctx context.Context, tx *sql.Tx, eni *ec2.NetworkInter
...
-	prefixInt := cidr.IPv6tod(net.IP)
-	newIPInt := prefixInt.Add(prefixInt, big.NewInt(lastAssigned))
-	newIP := cidr.DtoIPv6(newIPInt)
+	var counter uint32
+	r := rand.New(rand.NewSource(time.Now().UnixNano())) // nolint: gosec
+	atomic.StoreUint32(&counter, r.Uint32())
+	actual, _ := vpcService.counters.LoadOrStore(aws.StringValue(branchENI.SubnetId), &counter)
+	counterPointer := actual.(*uint32)
+
+	base := atomic.AddUint32(counterPointer, uint32(ipsToAssign)) - uint32(ipsToAssign)
```
MyUmmaGumma, line 1220: I think this was mentioned earlier but whats the reason to abandon the earlier method?

vpc/service/assign_addresses_v3.go
```diff
@@ -1115,34 +1116,138 @@ func getBorderGroupForENI(ctx context.Context, tx *sql.Tx, eni *ec2.NetworkInter
...
 	}
 
-	prefixInt := cidr.IPv6tod(net.IP)
-	newIPInt := prefixInt.Add(prefixInt, big.NewInt(lastAssigned))
-	newIP := cidr.DtoIPv6(newIPInt)
+	var counter uint32
+	r := rand.New(rand.NewSource(time.Now().UnixNano())) // nolint: gosec
+	atomic.StoreUint32(&counter, r.Uint32())
+	actual, _ := vpcService.counters.LoadOrStore(aws.StringValue(branchENI.SubnetId), &counter)
+	counterPointer := actual.(*uint32)
```
MyUmmaGumma, line 1218: Assuming this is the counter that can collide in a dense subnet - should we check if there is a collision at all - or is it too early here or it does not matter and the rest of the steps that actually come up with the address matters? Also are the chances of collisions small even in the most densely populated subnets?

MyUmmaGumma (approved):
If there are only 4095 worker IDs and we assign a worker ID every time we restart the service, does it matter at all if we overrun 4096 and an instance gets assigned an older ID. Maybe what I am asking is what is the role of the worker ID? An identity? Or something to introduce more randomness? EDIT: @DonavanFritz gave an explanation on the use of the worker ID to avoid contention on a single counter.

vpc/service/assign_addresses_v3.go
sargun: Which earlier method? We cannot rely on using random assign IP, due to our scale and limitations with AWS.

vpc/service/assign_addresses_v3.go
sargun: There are two things that would have to go wrong:
1. The worker ID would have to wrap. In the past ~180 days, only ~85 instances of the service have been started. I'm not worried about it wrapping unless there's a crash loop (very possible)
2. The per-subnet randomization of where each worker starts would have to collide.

PR #736, opened 2021-09-22
Disable CGO in titus executor build

sargun:


solarkennedy:
hehe, what could possibly go wrong!

solarkennedy (approved):

PR #741, opened 2021-09-30
Add an API to Reset Security Groups on vpcservice

MyUmmaGumma:
This change adds a GRPC API to be called by Spinnaker whenever a user wants to delete a security group on Spinnaker. The intent is that Spinnaker will in turn call this new API before issuing the call to AWS. 

If the Security Group (SG) is already associated with a container Titus will return an error to Spinnaker. If not, Titus will update its internal state and return to Spinnaker that the SG is safe for deletion.


vpc/service/internal_apis.go
```diff
@@ -407,3 +410,234 @@ func (vpcService *vpcService) doDetachBranchNetworkInterface(ctx context.Context
...
+	}(resetSgTx)
+
+	sgToDelete := in.GetSgId()
+
+	logger.G(ctx).Debug("1. Check if the SG to be reset is associated with a container ", sgToDelete)
+	// 1. Check if the SG to be reset is associated with a container
+	rows, err := resetSgTx.QueryContext(ctx, `
+SELECT COUNT(*) FROM branch_enis
+WHERE ARRAY[$1] <@ security_groups AND branch_eni IN
+(SELECT branch_eni FROM branch_eni_attachments WHERE branch_eni_attachments.association_id IN
```
sargun, line 441: I think this query should actually be:

```sql

SELECT branch_eni,
       account_id,
       az,
  (SELECT count(*)
   FROM assignments
   WHERE branch_eni_association =
       (SELECT association_id
        FROM branch_eni_attachments bea
        WHERE bea.branch_eni = branch_enis.branch_eni
          AND state = 'attached'))
FROM branch_enis
WHERE security_groups @> $1::text[]
```
This gives you the AZ, and branch_eni IDs. You then check the count column to make sure it's 0.
You can then give the user feedback to which specific branch_enis have a count of > 0 as well.

sargun: you could join the availability_zones table in to get the region as well.

vpc/service/internal_apis.go
```diff
@@ -407,3 +410,234 @@ func (vpcService *vpcService) doDetachBranchNetworkInterface(ctx context.Context
...
+		return &empty.Empty{}, err
+	}
+	defer func(tx *sql.Tx) {
+		_ = tx.Rollback()
+	}(resetSgTx)
+
+	//4. Reset the SG Id to default
+	logger.G(ctx).Debug("4. Reset the SG Id to default %s", sgToDelete)
+	rows, err = resetSgTx.QueryContext(ctx,
+		"UPDATE branch_enis SET security_groups = ARRAY[$1],dirty_security_groups=true WHERE ARRAY[$2] <@ security_groups returning branch_eni",
```
sargun, line 534: This update should be in the above transaction.

vpc/service/internal_apis.go
```diff
@@ -407,3 +410,234 @@ func (vpcService *vpcService) doDetachBranchNetworkInterface(ctx context.Context
...
+		defaultSecurityGroupID, sgToDelete)
+	if err != nil {
+		err = fmt.Errorf("Cannot mark security groups as dirty for %s : %w", sgToDelete, err)
+		tracehelpers.SetStatus(err, span)
+		return &empty.Empty{}, err
+	}
+
+	var enisWithSgToUpdate []string
```
sargun, line 542: On a per ENI basis you should:

1. Open a traditional TXN
2. SELECT FOR NO KEY UPDATE on the branch eni
3. Alter it in AWS
4. Update the dirty flag to false.

vpc/service/internal_apis.go
sargun: Why do you start a new txn here?

vpc/service/internal_apis.go
sargun: On step 2, make sure the dirty flag is still true, and only update it to the values in the DB.

sargun (changes requested):

vpc/proto/vpc.proto
```diff
@@ -314,6 +318,7 @@ service TitusAgentVPCService {
     rpc DisassociateTrunkNetworkInterface (DisassociateTrunkNetworkInterfaceRequest) returns (DisassociateTrunkNetworkInterfaceResponse);
     rpc DescribeTrunkNetworkInterface (DescribeTrunkNetworkInterfaceRequest) returns (DescribeTrunkNetworkInterfaceResponse);
     rpc DetachBranchNetworkInterface (DetachBranchNetworkInterfaceRequest) returns (DetachBranchNetworkInterfaceResponse);
+    rpc ResetSecurityGroup(SecurityGroupRequest) returns (google.protobuf.Empty);
```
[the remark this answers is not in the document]
MyUmmaGumma: Done.

vpc/service/internal_apis.go
```diff
@@ -407,3 +410,234 @@ func (vpcService *vpcService) doDetachBranchNetworkInterface(ctx context.Context
 		AssociationID: associationID,
 	}, nil
 }
+
+func (vpcService *vpcService) ResetSecurityGroup(ctx context.Context, in *vpcapi.SecurityGroupRequest) (*empty.Empty, error) {
```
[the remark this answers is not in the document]
MyUmmaGumma: Done.

vpc/proto/vpc.proto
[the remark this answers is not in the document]
MyUmmaGumma: done.

vpc/service/internal_apis.go
```diff
@@ -407,3 +409,231 @@ func (vpcService *vpcService) doDetachBranchNetworkInterface(ctx context.Context
...
+	defer func(tx *sql.Tx) {
+		_ = tx.Rollback()
+	}(resetSgTx)
+
+	sgToDelete := request.GetSgId()
+
+	// First get the default security group that we will replace the Reset SG with
+	logger.G(ctx).WithField("resetSg", sgToDelete).Debug("Get Default security group ID of the VPC to use in place of the SG")
+	row := resetSgTx.QueryRowContext(ctx, `
+SELECT region, account, group_id FROM security_groups WHERE  group_name='default' and account IN (SELECT account FROM security_groups WHERE group_id = $1) limit 1
```
sargun, line 439: Again, I don't understand why this can't be:

```sql
SELECT region, account, group_id FROM security_groups WHERE group_name='default' and account = (SELECT account FROM security_groups WHERE group_id = $1 LIMIT 1)
```

vpc/service/internal_apis.go
```diff
@@ -407,3 +409,231 @@ func (vpcService *vpcService) doDetachBranchNetworkInterface(ctx context.Context
...
+			"Security group does not exist")
+	}
+	if err != nil {
+		err = fmt.Errorf("Could not query database for VPC ID of %s: %w ", sgToDelete, err)
+		tracehelpers.SetStatus(err, span)
+		return &vpcapi.ResetSecurityGroupResponse{}, err
+	}
+	err = row.Scan(&region, &account, &defaultSecurityGroupID)
+	if err != nil {
+		err = fmt.Errorf("Could not get VPC ID of %s: %w ", sgToDelete, err)
```
sargun, line 454: errant hanging space.

vpc/service/internal_apis.go
```diff
@@ -407,3 +409,231 @@ func (vpcService *vpcService) doDetachBranchNetworkInterface(ctx context.Context
...
+		tracehelpers.SetStatus(err, span)
+		return &vpcapi.ResetSecurityGroupResponse{}, err
+	}
+
+	if err != nil {
```
sargun, line 459: Why are you checking error again here?

vpc/service/internal_apis.go
```diff
@@ -407,3 +409,231 @@ func (vpcService *vpcService) doDetachBranchNetworkInterface(ctx context.Context
...
+		err = fmt.Errorf("Could not get EC2 session for branch ENI: %w", err)
+		tracehelpers.SetStatus(err, span)
+		return &vpcapi.ResetSecurityGroupResponse{}, err
+	}
+
+	if len(defaultSecurityGroupID) == 0 {
```
sargun, line 465: AFAIK, this is impossible :P.

I'm not sure how we would have a row with the name = default, but the ID is empty.

vpc/service/internal_apis.go
```diff
@@ -407,3 +409,231 @@ func (vpcService *vpcService) doDetachBranchNetworkInterface(ctx context.Context
...
+		err = fmt.Errorf("Security group %s does not exist: %w ", sgToDelete, err)
+		tracehelpers.SetStatus(err, span)
+		return &vpcapi.ResetSecurityGroupResponse{}, status.Errorf(codes.InvalidArgument,
+			"Security group does not exist")
+	}
+
+	err = resetSgTx.Commit()
```
sargun, line 472: This transaction doesn't need to be serializable.  Also, it can be read only.

vpc/service/internal_apis.go
```diff
@@ -407,3 +409,231 @@ func (vpcService *vpcService) doDetachBranchNetworkInterface(ctx context.Context
...
+		tracehelpers.SetStatus(err, span)
+		return &vpcapi.ResetSecurityGroupResponse{}, err
+	}
+
+	var enisWithSgToUpdate []string
+	for rows.Next() {
+		var updatedEni string
+		err := rows.Scan(&updatedEni)
+		if err != nil {
+			err = fmt.Errorf("Could not get VPC ID of %s: %w ", sgToDelete, err)
```
sargun, line 540: I don't think this error string is correct.

vpc/service/internal_apis.go
```diff
@@ -407,3 +409,231 @@ func (vpcService *vpcService) doDetachBranchNetworkInterface(ctx context.Context
...
+	var dirtySecurityGroups bool
+	err = row.Scan(pq.Array(&securityGroups), &dirtySecurityGroups)
+	if err != nil {
+		err = errors.Wrap(err, "Unable to scan branch_enis for dirty_security_groups")
+		tracehelpers.SetStatus(err, span)
+		return err
+	}
+
+	//Check if anything changed on the SG <-> ENI association
+	if !dirtySecurityGroups && len(securityGroups) == 1 && securityGroups[0] == defSgS[0] {
```
sargun, line 605: If dirtysecuritygroups is false, we can bail here entirely.

vpc/service/internal_apis.go
```diff
@@ -407,3 +409,231 @@ func (vpcService *vpcService) doDetachBranchNetworkInterface(ctx context.Context
...
+		}
+		return nil
+	}
+
+	_, err = session.ModifyNetworkInterfaceAttribute(ctx, ec2.ModifyNetworkInterfaceAttributeInput{
+		NetworkInterfaceId: aws.String(eni),
+		Groups:             aws.StringSlice(defSgS),
+	})
+	if err != nil {
+		_ = tx.Rollback()
```
sargun, line 620: You don't need to explicitly rollback here.

vpc/service/internal_apis.go
sargun: I think you can set  modified_at = transaction_timestamp() here.

vpc/service/internal_apis.go
MyUmmaGumma: Changed.

vpc/service/internal_apis.go
MyUmmaGumma: True, removed.

vpc/service/internal_apis.go
MyUmmaGumma: Removed.

vpc/service/internal_apis.go
MyUmmaGumma: Removed.

vpc/service/internal_apis.go
MyUmmaGumma: Changed.

vpc/service/internal_apis.go
MyUmmaGumma: Done.

vpc/service/internal_apis.go
MyUmmaGumma: Done.

vpc/service/integration_test.go
```diff
@@ -799,6 +807,97 @@ func testGenerateAssignmentIDNewSG(ctx context.Context, t *testing.T, md integra
 	assert.NilError(t, err)
 }
 
+func testResetSecurityGroup(ctx context.Context, t *testing.T, md integrationTestMetadata, service *vpcService, session *ec2wrapper.EC2Session) {
+	//This test requires an SG that is not currently associated with any container
+	specialResetSgTestSg := "sg-01d281a4cc5f620c9"
```
sargun, line 812: Please parameterize this.

sargun (approved):

PR #742, opened 2021-10-01
Added /run tmpfs for other containers in a pod

solarkennedy:
This matches the /run configuration behavior on the main
container. This allows sidecars to also use /run
in a tmpfs manner, and have faith the files will not
be on disk, etc.

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -1830,6 +1830,9 @@ func (r *DockerRuntime) k8sContainerToDockerConfigs(v1Container v1.Container, ma
 		VolumesFrom: []string{mainContainerID},
 		Mounts:      mounts,
 		Init:        &b,
+		Tmpfs: map[string]string{
+			"/run": "rw,exec,size=" + defaultRunTmpFsSize,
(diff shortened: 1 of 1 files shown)


executor/runtime/docker/docker.go
```diff
@@ -1830,6 +1830,9 @@ func (r *DockerRuntime) k8sContainerToDockerConfigs(v1Container v1.Container, ma
 		VolumesFrom: []string{mainContainerID},
 		Mounts:      mounts,
 		Init:        &b,
+		Tmpfs: map[string]string{
+			"/run": "rw,exec,size=" + defaultRunTmpFsSize,
```
kzwang, line 1834: should this be `/run` or `/tmp`?

solarkennedy: `/run`, to match existing behavior with the main container.

Titus doesn't seem to do anything special with `/tmp` and assumes images come with it.

sargun: Would this be a /run (of that size) per container? Shouldn't this be shared?

PR #743, opened 2021-10-01
Added a new /run-shared along with /logs to be shared with all containers in a pod

solarkennedy:
This new shared location is to help facilitate IPC between containers,
without burdening every container to make their own shared volume/volumeMount

I couldn't use anything in /run because the /run tmpfs gets mounted on top,
and is harder to control.


executor/runtime/docker/docker.go
```diff
@@ -1740,12 +1748,20 @@ func (r *DockerRuntime) k8sContainerToDockerConfigs(v1Container v1.Container, ma
...
-		mounts = append(mounts, mount.Mount{
-			Type:     "bind",
-			Source:   mainContainerRoot + "/logs",
-			Target:   "/logs",
-			ReadOnly: false,
-		})
+		mounts = append(
+			mounts, mount.Mount{
+				Type:     "bind",
+				Source:   mainContainerRoot + "/logs",
```
sargun, line 1754: filepath.join.

executor/runtime/docker/docker.go
```diff
@@ -1519,10 +1527,10 @@ func (r *DockerRuntime) statusMonitor(cancel context.CancelFunc, containerID str
 			isTerminalEvent := r.handleDockerEvent(event, statusMessageChan)
 			if isTerminalEvent {
 				cName, err := r.getContainerNameFromID(containerID)
-				if err != nil {
+				if err == nil {
 					log.Infof("Closing docker status monitor for %s because terminal docker event received", cName)
```
sargun, line 1531: use withfield please.

executor/runtime/docker/docker.go
```diff
@@ -1273,6 +1273,14 @@ func (r *DockerRuntime) pushEnvironment(ctx context.Context, c runtimeTypes.Cont
 		log.Fatal(err)
 	}
 
+	if err := tw.WriteHeader(&tar.Header{
+		Name:     "/run-shared",
+		Mode:     0777,
+		Typeflag: tar.TypeDir,
+	}); err != nil {
+		log.Fatal(err)
```
sargun, line 1281: log.witherror.fatal?

sargun (approved):

alekar (approved):

PR #746, opened 2021-10-08
Fixed support for shared volume mount propagation

solarkennedy:
Previously we were passing in literally the v1 k8s mount
propgation string, which docker didn't like.

This adds the appropriate conversion and a test.

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -1881,6 +1881,20 @@ func v1ContainerHealthcheckToDockerHealthcheck(probe *v1.Probe) *container.Healt
 	return &hc
 }
 
+// v1MountPropToDockerProp converts the incoming pod mount propagation configuration
+// into something that docker can understand, using the documentation for what each means:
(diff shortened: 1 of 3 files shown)


solarkennedy:
Hmm. While I think this code is correct, I can't get it to pass on CI due to 

```time=2021-10-13T22:09:35Z level=info msg=Returning TASK_LOST for Task: Failed to start a user/platform container: Failed to start touch-sentinel user container: Error response from daemon: linux mounts: path /var/lib/docker/overlay2/4cf0a1daea22d3e4607909f9efe012d0328512f95dd49c3d99478737cb828b7f/merged/etc is mounted on /var/lib/docker/overlay2/4cf0a1daea22d3e4607909f9efe012d0328512f95dd49c3d99478737cb828b7f/merged but it is not a shared or slave mount jobID=Titus-75429 taskID=Titus-86629-TestBasicMultiContainerCustomSharedVolumes-0-2
```

This passes on my laptop though.

PR #747, opened 2021-10-08
Added extra container stdout/err to be archived by default

solarkennedy:
Now that we have multi-container, the naming convention is
just like before (stdout/stderr) except we tack on the container
name, so stdout-nging/stderr-nging).

This addes support to archive these logs by default, so you
can troubleshoot a failed container, even if it isn't the main one.

----

Is there some other place to add *rotation* too?


aaronlehmann (approved):
Thanks!

alekar:
You need to change how `stdioRotate` works with `PotentialStdioNames` as we discussed in chat.

alekar (approved):

PR #751, opened 2021-10-21
Create file caches for instance identity and pkcs7

MyUmmaGumma:
Reduce calls to IMDS by caching the identity document

### DIFF
--- a/vpc/tool/identity/ec2.go
+++ b/vpc/tool/identity/ec2.go
@@ -44,8 +44,48 @@ const (
 
 type ec2Provider struct{}
 
-func (e *ec2Provider) GetIdentity(ctx context.Context) (*vpcapi.InstanceIdentity, error) {
+func (e *ec2Provider) GetIdentityFromFile(ctx context.Context) (*ec2metadata.EC2InstanceIdentityDocument, []byte, []byte, error) {
 	var pkcsFile *os.File
+	_, span := trace.StartSpan(ctx, "GetIdentityFromFile")
+	defer span.End()
+	// Attempt to read the data off of cache
+	instanceIdentityFile, err := os.OpenFile(InstanceIdentityFile, os.O_RDONLY, 0644)
+	if err != nil {
+		tracehelpers.SetStatus(ErrEC2MetadataFileNotFound, span)
+		return nil, nil, nil, err
+	}
+
+	pkcsFile, err = os.OpenFile(PkcsFile, os.O_RDONLY, 0644)
+	if err != nil {
+		tracehelpers.SetStatus(ErrPkcsFileNotFound, span)
+		return nil, nil, nil, err
+	}
+	defer instanceIdentityFile.Close()
+	defer pkcsFile.Close()
+
+	// Read the instance Identity file
+	instanceID, err := ioutil.ReadAll(instanceIdentityFile)
+	if err != nil {
+		tracehelpers.SetStatus(err, span)
+		return nil, nil, nil, err
+	}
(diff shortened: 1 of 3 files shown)


vpc/tool/identity/ec2.go
```diff
@@ -31,15 +35,63 @@ var (
...
+	// Attempt to read the data off of /run
+	instanceIdentityFile, err := os.OpenFile(InstanceIdentityFile, os.O_RDONLY, 0644)
+	if errors.Is(err, os.ErrNotExist) || err != nil { //should we separate out the does not exist with the regular error?
+		tracehelpers.SetStatus(ErrEC2MetadataServiceUnavailable, span)
+		goto fetchFromImds
+	}
+	pkcsFile, err = os.OpenFile(PkcsFile, os.O_RDONLY, 0644)
+	if errors.Is(err, os.ErrNotExist) || err != nil { //should we separate out the does not exist with the regular error?
+		// handle the case where the file doesn't exist
+		tracehelpers.SetStatus(ErrPkcsFileNotFound, span)
```
sargun, line 63: If the file doesn't exist, why not try to perform the below action, and retry? Or launch a goroutine to write the files out?

vpc/tool/identity/ec2.go
```diff
@@ -74,6 +126,22 @@ func (e *ec2Provider) GetIdentity(ctx context.Context) (*vpcapi.InstanceIdentity
 
 	stats.Record(ctx, getIdentityLatency.M(float64(time.Since(start).Nanoseconds())), getIdentitySuccess.M(1))
 
+	//Write the files to /run
+	iidfile, err := os.OpenFile(InstanceIdentityFile, os.O_WRONLY, 0644)
```
sargun, line 130: Don't ever write files directly that must be written atomically. Use O_TMPFILE + rename. https://github.com/google/renameio

vpc/tool/identity/ec2.go
```diff
@@ -74,6 +132,8 @@ func (e *ec2Provider) GetIdentity(ctx context.Context) (*vpcapi.InstanceIdentity
 
 	stats.Record(ctx, getIdentityLatency.M(float64(time.Since(start).Nanoseconds())), getIdentitySuccess.M(1))
 
+	go cacheIdentityFiles(ctx, resp, pkcs7)
```
sargun, line 135: I think this should be synchronous, because many titus vpc tool invocations are short lived.

So, I think the flow should go:
1. Call identity cache
2. Cache checks if files are on disk, if files are on disk, return identity from them
3. Fetch identity from IMDS
4. Write out cache
5. Return identity from identity fetched in (3)

sargun (changes requested):

vpc/tool/identity/ec2.go
MyUmmaGumma: done.

vpc/tool/identity/ec2.go
MyUmmaGumma: done.

vpc/tool/identity/ec2.go
```diff
@@ -84,6 +153,24 @@ func (e *ec2Provider) GetIdentity(ctx context.Context) (*vpcapi.InstanceIdentity
 	}, err
 }
 
+func cacheIdentityFiles(ctx context.Context, resp string, pkcs string) {
+	_, span := trace.StartSpan(ctx, "GetIdentity")
```
sargun, line 157: span name should be cacheIdentityFiles

sargun (approved):

PR #753, opened 2021-10-21
Added sshdmulti for multiple containers

solarkennedy:
This adds a new "sshdMulti" Titus System Service to run
multiple sshds for each container in a pod.

In the C&W way, the main container is untouched, and the sshd runs
as normal as before.

The Mulit one is special and spawns one per extra container, incrementing
the port number by one each.

No additional supervision is added though. They are second-class in that way.

Some concerns:

1. Gotta get 7523+ ports open, seems tractable
2. I'm a bit worried about all the extra ram that will go towards the host (not task!)

### DIFF
--- a/root/apps/titus-executor/bin/titus-sshd-multi
+++ b/root/apps/titus-executor/bin/titus-sshd-multi
@@ -1,0 +1,28 @@
+#!/usr/bin/env bash
+# Spawns and extra sshd for ever additional container past
+# the main container in a Pod.
+
+function listExtraContainersFor {
+  # It is important to get the names of the extra containers in the same order that they
+  # are reported in the pod, because that is what the control-plane knows.
+  # The order here ensures that we pick a port number that matches with what
+  # other systems will assume.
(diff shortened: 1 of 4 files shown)


root/apps/titus-executor/bin/titus-sshd-multi
```diff
@@ -0,0 +1,28 @@
...
+    jq -r '.spec.containers[1:] | .[].name' "/run/titus-executor/default__$1/pod.json"
+  fi
+}
+
+function spawnSshdForContainer {
+  local cName=$1
+  local port=$2
+  echo "Spawning titus-sshd for docker container $cName on port $port"
+  PID1=$(docker inspect "${TASK_ID}-$cName" --format '{{.State.Pid}}')
+  TITUS_PID_1_DIR="/proc/$PID1/root/proc/1" /apps/titus-executor/bin/titus-nsenter /titus/sshd/run-titus-sshd -D -e -p "$port"
```
JackOfMostTrades, line 20: Is this going to pick up the same sshd config as the main container? In particular, wanting to check if it has the same authorized_keys / authorized_principals.

PR #758, opened 2021-10-27
RFC: Added host entries for all containers in a pod

solarkennedy:
I want to make it easier for containers to live in a world
where there could be port conflicts.

This isnt a 100% solution, but it at least allows in theory
containers in a pod to play nice and bind only to their own
"localhost" address.

With this in place, multiple containers in a pod could all
listen on $TITUS_CONTAINER_NAME:80 and it would work!


solarkennedy:
I'm going to close this for now.
Just a POC of how "Easy" it would be to add this to the multi-container contract.

PR #760, opened 2021-10-27
Implement Liveness Probes in C&W Containers

solarkennedy:
This PR implements basic liveness probes (exec only).
Specifically, we already have the plubming here, we just were not
actually failing containers that fail their probes like k8s does.
(docker does not do this natively).

It is too bad docker events doesn't have more data about the health
status more than healthy/unhealthy, but at least these do work for
the most basic healthchecks.

Future improvements could be to incorporate lifecycle policies to
auto-restart platform sidecar and things like that, keeping track
of restart counts and other fancy things.

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -1917,7 +1917,6 @@ func (r *DockerRuntime) statusMonitor(cancel context.CancelFunc, containerID str
 			log.WithError(err).Errorf("Got unexpected error while listening for docker events for %s, bailing: %s", containerID, err)
 			return
 		case event := <-eventChan:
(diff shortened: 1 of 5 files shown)


solarkennedy:
Still having a super hard getting the standalone test to pass on CI, but passes locally.
Somehow the `health_status` docker event slips by (even though I see it when I run `docker events`).
All other events are fine. `health_status` shows up in other tests, just not this one. WTF.

solarkennedy:
I'm giving up on trying to make the standalone test pass. Something about docker-in-docker makes it lose the event, but I was unable to figure out why. It does pass locally on my machine.

PR #762, opened 2021-10-28
Added an imageinspect on all extra containers and correctly set entrypoint for them

solarkennedy:
Normally, we would be able to pass through our container spec's
command/entrypoint as is, and let docker let the image
defaults take over if unset.
    
However, on Titus sidecars, we want tini, which means we must manipulate the entrypoint.
    
On the main container, we don't need to do this to get tini because
we run with `init:true`, and this injects tini for us into the
final entrypoint calculation. This does *not* happen on sidecars
with share pid namespaces, so we must inject it ourselves.

But we *still* want to default to use the image's entrypoint if available, just like docker/k8s would.

This code inspects the image so that we can do that.

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -1623,10 +1623,20 @@ func (r *DockerRuntime) pullAllExtraContainers(ctx context.Context, pod *v1.Pod)
 	otherUserContainers := append(r.c.ExtraUserContainers(), r.c.ExtraPlatformContainers()...)
 	group := groupWithContext(ctx)
 	for _, c := range otherUserContainers {
-		image := c.V1Container.Image
+		c2 := c
 		group.Go(func(ctx context.Context) error {
+			image := c2.V1Container.Image
 			l.Debugf("pulling other container image %s", image)
-			return pullWithRetries(ctx, r.cfg, r.metrics, r.client, image, doDockerPull)
+			err := pullWithRetries(ctx, r.cfg, r.metrics, r.client, image, doDockerPull)
+			if err != nil {
+				return err
+			}
+			imageInspect, err2 := imageExists(ctx, r.client, image)
+			if err2 != nil {
+				return fmt.Errorf("Failed to inspect %s after pull: %s", image, err2)
+			}
+			c2.ImageInspect = imageInspect
+			return nil
 		})
 	}
 	return group.Wait()
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -1633,7 +1633,7 @@ func (r *DockerRuntime) pullAllExtraContainers(ctx context.Context, pod *v1.Pod)
 			}
 			imageInspect, err2 := imageExists(ctx, r.client, image)
 			if err2 != nil {
-				return fmt.Errorf("Failed to inspect %s after pull: %s", image, err2)
+				return fmt.Errorf("Failed to inspect %s after pull: %w", image, err2)
 			}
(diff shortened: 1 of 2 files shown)


executor/runtime/docker/docker.go
```diff
@@ -1873,6 +1873,34 @@ func (r *DockerRuntime) k8sContainerToDockerConfigs(v1Container v1.Container, ma
...
+	originalEntrypoint := getExtraContainerEntrypoint(c)
+	return append([]string{"/sbin/docker-init", "-s", "--"}, originalEntrypoint...)
+}
+
+// getExtraContainerEntrypoint computes the original entrypoint that the user
+// wants to run, given two sources:
+// 1. the input k8s pod "command" (takes precidence)
+// 2. the input ENTRYPOINT on the image (should be used if no command is specified on the container spec)
+func getExtraContainerEntrypoint(c *runtimeTypes.ExtraContainer) []string {
+	if c.V1Container.Command != nil && len(c.V1Container.Command) > 0 {
```
aaronlehmann, line 1897: Nit: The nil check is redundant.

executor/runtime/docker/docker.go
```diff
@@ -1873,6 +1873,34 @@ func (r *DockerRuntime) k8sContainerToDockerConfigs(v1Container v1.Container, ma
...
+	// The reason we do this is because, even with init=true, docker will only inject tini
+	// on containers running in a private pid namespace.
+	// On titus, we want tini on *every* container, because it gives us features like stdout/err
+	originalEntrypoint := getExtraContainerEntrypoint(c)
+	return append([]string{"/sbin/docker-init", "-s", "--"}, originalEntrypoint...)
+}
+
+// getExtraContainerEntrypoint computes the original entrypoint that the user
+// wants to run, given two sources:
+// 1. the input k8s pod "command" (takes precidence)
```
aaronlehmann, line 1894:
```suggestion
// 1. the input k8s pod "command" (takes precedence)
```

executor/runtime/docker/docker.go
```diff
@@ -1873,6 +1873,34 @@ func (r *DockerRuntime) k8sContainerToDockerConfigs(v1Container v1.Container, ma
...
+// 2. the input ENTRYPOINT on the image (should be used if no command is specified on the container spec)
```
aaronlehmann, line 1895: Would it make sense to fall back to the COMMAND on the image, or is that redundant with the value from V1Container?

executor/runtime/docker/docker.go
```diff
@@ -1619,10 +1619,20 @@ func (r *DockerRuntime) pullAllExtraContainers(ctx context.Context, pod *v1.Pod)
...
+			image := c2.V1Container.Image
 			l.Debugf("pulling other container image %s", image)
-			return pullWithRetries(ctx, r.cfg, r.metrics, r.client, image, doDockerPull)
+			err := pullWithRetries(ctx, r.cfg, r.metrics, r.client, image, doDockerPull)
+			if err != nil {
+				return err
+			}
+			imageInspect, err2 := imageExists(ctx, r.client, image)
+			if err2 != nil {
+				return fmt.Errorf("Failed to inspect %s after pull: %s", image, err2)
```
sargun, line 1632: use %w for the second formatter.

sargun (approved):

executor/runtime/docker/docker.go
solarkennedy: I don't think so.
I'm pretty sure docker will do that for us.

aaronlehmann (approved):

PR #763, opened 2021-11-03
enable traffic steering syscall interception on Tini for TSA

aloktiagi:
### Description of the Change

### DIFF
--- a/tini/src/seccomp_fd_notify.c
+++ b/tini/src/seccomp_fd_notify.c
@@ -183,6 +183,25 @@ static int install_notify_filter(void) {
 		BPF_STMT(BPF_RET | BPF_K, SECCOMP_RET_ALLOW),
 	};
 
+	struct sock_filter traffic_steering_filter[] = {
+		/* X86_64_CHECK_ARCH_AND_LOAD_SYSCALL_NR */
+		BPF_STMT(BPF_LD | BPF_W | BPF_ABS, (offsetof(struct seccomp_data, arch))),
+		BPF_JUMP(BPF_JMP | BPF_JEQ | BPF_K, AUDIT_ARCH_X86_64, 0, 2),
+		BPF_STMT(BPF_LD | BPF_W | BPF_ABS, (offsetof(struct seccomp_data, nr))),
+		BPF_JUMP(BPF_JMP | BPF_JGE | BPF_K, X32_SYSCALL_BIT, 0, 1),
+		BPF_STMT(BPF_RET | BPF_K, SECCOMP_RET_ALLOW),
+
+		/* Trap setsocktopt */
+		BPF_JUMP(BPF_JMP | BPF_JEQ | BPF_K, __NR_setsockopt, 0, 1),
+		BPF_STMT(BPF_RET + BPF_K, SECCOMP_RET_USER_NOTIF),
+		/* Trap listen */
+		BPF_JUMP(BPF_JMP | BPF_JEQ | BPF_K, __NR_listen, 0, 1),
+		BPF_STMT(BPF_RET + BPF_K, SECCOMP_RET_USER_NOTIF),
+
+		/* Every other system call is allowed */
+		BPF_STMT(BPF_RET | BPF_K, SECCOMP_RET_ALLOW),
+	};
+
 	/* If we don't have CAP_SYSADMIN, we MUST set NO_NEW_PRIVS.
 	Why? Read https://unix.stackexchange.com/a/562899/411719
 	Or, from the seccomp man page:
@@ -198,6 +217,7 @@ static int install_notify_filter(void) {
 
 	char *is_handle_net = getenv("TITUS_SECCOMP_AGENT_HANDLE_NET_SYSCALLS");
 	char *is_handle_perf = getenv("TITUS_SECCOMP_AGENT_HANDLE_PERF_SYSCALLS");
+	char *is_traffic_steering = getenv("TITUS_SECCOMP_AGENT_HANDLE_TRAFFIC_STEERING");
 	if (is_handle_net != NULL) {
 		prog.filter = net_filter;
 		prog.len = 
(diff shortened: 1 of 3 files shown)


tini/src/seccomp_fd_notify.c
```diff
@@ -179,23 +181,60 @@ struct sock_filter net_filter[] = {
...
+	     i < (unsigned short)(sizeof(check_filter) / sizeof(check_filter[0])) &&
+		 filter_idx < BPF_MAXINSNS;
+		 i++) {
+		filter[filter_idx] = check_filter[i];
+		filter_idx++;
+	}
+
 	char *is_handle_net = getenv("TITUS_SECCOMP_AGENT_HANDLE_NET_SYSCALLS");
 	char *is_handle_perf = getenv("TITUS_SECCOMP_AGENT_HANDLE_PERF_SYSCALLS");
+	char *is_traffic_steering = getenv("TITUS_TRAFFIC_STEERING");
```
solarkennedy, line 194: What do you think about `TITUS_SECCOMP_AGENT_HANDLE_TRAFFIC_STEERING` ?

MyUmmaGumma (approved):

tini/src/seccomp_fd_notify.c
```diff
@@ -179,23 +184,60 @@ struct sock_filter net_filter[] = {
...
+	for (int i = 0;
+	     i < (unsigned short)(sizeof(allow_filter) / sizeof(allow_filter[0])) &&
+	     filter_idx < BPF_MAXINSNS;
+		 i++) {
+		filter[filter_idx] = allow_filter[i];
+		filter_idx++;
+	}
+
+	prog.filter = filter;
+	prog.len = filter_idx;
```
solarkennedy, line 239: Whoa, is this right? Are we nesting the filters now instead of concatenating them?

aloktiagi: Not nesting, we are still concatenating based on what config is enabled.

solarkennedy:
In the future can you write commit messages when you `git commit`, and that way the explanation for what you are trying to do will automatically fill in the github PR description?

tini/src/seccomp_fd_notify.c
```diff
@@ -179,23 +184,60 @@ struct sock_filter net_filter[] = {
 		}
 	}
 
+	for (int i = 0;
+	     i < (unsigned short)(sizeof(check_filter) / sizeof(check_filter[0])) &&
```
solarkennedy, line 188: Can you add a comment to explain what this code does? 
I don't understand it.

I think I don't understand it because you end up using the `check_filter` variable in 2 ways.
1. It stores the kind of first bpf filter  we need
2. It grows to include the final computed filter with all the individual filters combined.

Maybe have two different variables?

tini/src/seccomp_fd_notify.c
```diff
@@ -179,23 +184,60 @@ struct sock_filter net_filter[] = {
...
-		prog.filter = net_filter;
-		prog.len = 
-			(unsigned short)(sizeof(net_filter) / sizeof(net_filter[0]));
+		for (int i = 0;
+		     i < (unsigned short)(sizeof(net_filter) / sizeof(net_filter[0])) &&
+			 filter_idx < BPF_MAXINSNS;
+		     i++) {
+			filter[filter_idx] = net_filter[i];
+			filter_idx++;
+		}
```
solarkennedy, line 205: Why is this just "not" simply appending to an array?

tini/src/seccomp_fd_notify.c
```diff
@@ -179,23 +184,60 @@ struct sock_filter net_filter[] = {
...
 		PRINT_INFO("Networking system calls will be intercepted");
 	} else if (is_handle_perf != NULL) {
-		prog.filter = perf_filter;
-		prog.len = 
-			(unsigned short)(sizeof(perf_filter) / sizeof(perf_filter[0]));
+		for (int i = 0;
+		     i < (unsigned short)(sizeof(perf_filter) / sizeof(perf_filter[0])) &&
```
sargun, line 209: 1. The size of thing could be an aRRAY_SIZE macro.
2. This concatenation bit looks like it could be reused.

tini/src/seccomp_fd_notify.c
```diff
@@ -161,7 +155,18 @@ struct sock_filter net_filter[] = {
 		/* Trap connect */
 		BPF_JUMP(BPF_JMP | BPF_JEQ | BPF_K, __NR_connect, 0, 1),
 		BPF_STMT(BPF_RET + BPF_K, SECCOMP_RET_USER_NOTIF),
+	};
 
+	struct sock_filter traffic_steering_filter[] = {
```
sargun, line 160: Does it make sense to borrow the cBPF code from the kernel which rewrites jumps?

tini/src/seccomp_fd_notify.c
aloktiagi: We have a variable `filter` that is an array of size `BPF_MAXINSNS + 1`.

We are populating this array upto a max size of `BPF_MAXINSNS` with individual arrays of filters i.e. `check_filter`, `net_filter`, `perf_filter`, `traffic_steering_filter` and `allow_filter` depending on what feature flag is enabled.

The `filter` variable is finally installed using the call to seccomp()

tini/src/seccomp_fd_notify.c
aloktiagi: Fixed.

tini/src/seccomp_fd_notify.c
aloktiagi: Looked at https://github.com/torvalds/linux/blob/master/samples/seccomp/bpf-fancy.c#L44. We can borrow that. Need to spend some time on crafting that. Will open another PR.

solarkennedy (approved):
I think @sargun should be your primary reviewer here

solarkennedy:
I'm good to go if @sargun  is.

PR #766, opened 2021-11-05
TITUS-5775: Basic titus-mount-ceph binary

solarkennedy:
This is way simpler than the NFS one, mostly because
of the reduced constraint of not needing to do DNS
resolution inside the container.

If possible, I would like the titus-storage binary do the
heavy lifting of composing the `source=` MOUNT_OPTIONS
for this binary to receive.

---- 

cc @ xiaolong302

Now that we have 3 binaries that use common functions, if you want I'll make another
pr after this is merge that starts putting things into a common file, if you think
that is actually worth it.

### DIFF
--- a/mount/titus-mount-ceph.c
+++ b/mount/titus-mount-ceph.c
@@ -93,7 +93,8 @@ static inline int move_mount(int from_dfd, const char *from_pathname,
 			mount_error(fd, key ?: "create");                      \
 	} while (0)
 
-static void configure_source(int fsfd, const char *source) {
+static void configure_source(int fsfd, const char *source)
+{
 	E_fsconfig(fsfd, FSCONFIG_SET_STRING, "source", source, 0);
 }
 
@@ -103,12 +104,15 @@ static void process_option(char *option, int fsfd)
            We supply all the inputs here, so it is safe, but parsing things
            like this is a little dangerous */
 	char *key, *value, *saveptr;
+	char value2[256] = "";
+
 	key = strtok_r(option, "=", &saveptr);
 	option = NULL;
 	value = strtok_r(option, "=", &saveptr);
 
-	char value2[256] = "";
+	assert(strlen(value) < sizeof(value2) - 1);
 	strcpy(value2, value);
+
 	/* We have to tack on an extra '==' because the secret is base64 encoded,
 	 * and this very basic string parsing will treat those == as tokens and skip
 	 * past them */
@@ -124,10 +128,11 @@ static void process_option(char *option, int fsfd)
 
 static void do_fsconfigs(int fsfd, char *options)
 {
-	char *str1, *token, *saveptr;
+	char *str1 = options;
+	char *token, *saveptr;
 	/* Mount options come in in the classic comma-separated key=value pairs
 	   we need to split them up and pass them in for fsconfig to handle one at a time */
-	for (str1 = options;; str1 = NULL) {
+	for (;; str1 = NULL) {
 		token = strtok_r(str1, ",", &saveptr);
 		if (token == NULL)
 			break;
@@ -243,10 +248,6 @@ int main(int argc, char *argv[])
 	errno = 0;
 	container_pid = strtol(argv[1], NULL, 10);
 	assert(errno == 0);
-
-	int buf_size = sysconf(_SC_PAGESIZE);
-	char final_options[buf_size];
-	strcpy(final_options, options);
 
 	if (!(target && flags && options && source)) {
 		fprintf(stderr,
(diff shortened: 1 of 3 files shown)


mount/titus-mount-ceph.c
```diff
@@ -0,0 +1,257 @@
...
+{
+	return syscall(__NR_move_mount, from_dfd, from_pathname, to_dfd,
+		       to_pathname, flags);
+}
+
+#define E_fsconfig(fd, cmd, key, val, aux)                                     \
+	do {                                                                   \
+		if (fsconfig(fd, cmd, key, val, aux) == -1)                    \
+			mount_error(fd, key ?: "create");                      \
+	} while (0)
```
alekar, line 94: This is a weird construct? Did you add this for debugging?

MyUmmaGumma (approved):

solarkennedy:
I was wrong about composing "source=". It is tricky because ceph sources have commas in them, which confuses my lame C code string splitting.

Instead I've gone with a dedicated MOUNT_SOURCE var so it doesn't need to be manipulated in any way.

mount/titus-mount-ceph.c
```diff
@@ -0,0 +1,284 @@
...
+	E_fsconfig(fsfd, FSCONFIG_SET_STRING, "source", source, 0);
+}
+
+static void process_option(char *option, int fsfd)
+{
+	/* Splits up a k=v string and runs fsconfig on it.
+           We supply all the inputs here, so it is safe, but parsing things
+           like this is a little dangerous */
+	char *key, *value, *saveptr;
+	key = strtok_r(option, "=", &saveptr);
```
sargun, line 106: newline after you state all these variables, also, move the declaration of value2 up here.

mount/titus-mount-ceph.c
```diff
@@ -0,0 +1,284 @@
...
+	option = NULL;
+	value = strtok_r(option, "=", &saveptr);
+
+	char value2[256] = "";
```
sargun, line 110: probably assert that strlen(value) < sizeof(value) - 1.

mount/titus-mount-ceph.c
```diff
@@ -0,0 +1,284 @@
...
+		key, value2);
+	E_fsconfig(fsfd, FSCONFIG_SET_STRING, key, value2, 0);
+}
+
+static void do_fsconfigs(int fsfd, char *options)
+{
+	char *str1, *token, *saveptr;
+	/* Mount options come in in the classic comma-separated key=value pairs
+	   we need to split them up and pass them in for fsconfig to handle one at a time */
+	for (str1 = options;; str1 = NULL) {
```
sargun, line 130: why not initialize str1 above, and make this a for (;;) loop?

mount/titus-mount-ceph.c
```diff
@@ -0,0 +1,284 @@
...
+	if (ret == -1) {
+		perror("setns net / mount");
+		exit(1);
+	}
+}
+
+static int setup_fsfd_in_namespaces(int sk, int pidfd)
+{
+	int fsfd, ret;
+	ret = setns(pidfd, CLONE_NEWUSER);
```
sargun, line 153: Dumb question, can you do setns(pidfd, CLONE_NEWNET | CLONE_NEWNS | CLONE_NEWUSER)?

mount/titus-mount-ceph.c
```diff
@@ -0,0 +1,284 @@
...
+	if (ret == -1) {
+		perror("setns mnt");
+		return 1;
+	}
+	fsfd = fsopen("ceph", 0);
+	if (fsfd == -1) {
+		perror("fsopen");
+		return 1;
+	}
+	assert(send_fd(sk, fsfd) == 0);
```
sargun, line 173: I find that the clone + cloned fd table trick is a little easier than the sendfd trick.

mount/titus-mount-ceph.c
solarkennedy: I'm not 100% sure, but I thought, no, that doesn't work, because NEWUSER requires forking, and doesn't take effect in the current process?

solarkennedy: Or are you asking, can I just set them all at once?

Yes, I can. Previously though I split it out, one for each namespace to make it easier to understand *why* it couldn't setns.

mount/titus-mount-ceph.c
solarkennedy: Yea but the sendfd trick works.

Has something changed between all the work we did to get this sendfd trick working with fscreate? (something changed that we no longer need this trick?)

PR #775, opened 2021-11-10
Ignore multi-container exits of 137, letting the "real" exit come thr…

solarkennedy:
…ough

Sometimes (race) one docker die event of a sidecar will come *before* the
"real" die event for the main container comes in.

This results in a returned exit code of 137 (docker kill). Which doesn't
meet what we expect (0, exit of the main container).

This PR adds an exception for non-main 137s, so that we don't stop
listening for more docker events, so that the "real" exit code
can shine through.

This fixes the flakey tests and hopefully results in a general better
experience.

Hopefully this doesn't introduce a new edge case where something really
does have a 137 exit code "oom", but not a docker "oom" event, or something
like that.
If we do see this, at least well have a log to see what happened in the wild.

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -2047,20 +2047,27 @@ func (r *DockerRuntime) handleDockerEvent(message events.Message, statusMessageC
 		l.Debugf("Skipping docker start event for %s, no need to update the pod", cName)
 		return nonTerminalDockerEvent
 	case "die":
-		// TODO: Handle the difference between platform/user sidecar, not all
-		// "die"s should result in a Finish
-		if exitCode := message.Actor.Attributes["exitCode"]; exitCode == "0" {
+		exitCode := message.Actor.Attributes["exitCode"]
+		if exitCode == "0" {
 			statusMessageChan <- runtimeTypes.StatusMessage{
 				Status: runtimeTypes.StatusFinished,
 				Msg:    cName + " container successfully exited with 0",
 			}
+			return isTerminalDockerEvent
+		} else if exitCode == "137" && cName != mainContainerName {
+			// An exit code of 137 means it was killed.
+			// If we are not on the 'main' container, and docker doesn't have us in the 'oom' case,
+			// then it *probably* means that a sidecar container got killed while we were tearing down the pod
+			// In this case, we should not update the task status, and let the "real" docker event do that for us.
+			l.Infof("Ignoring %s container dying with exit code 137, probably meaning that it was killed while the main container was exiting", cName)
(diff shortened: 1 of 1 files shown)


executor/runtime/docker/docker.go
```diff
@@ -2047,13 +2047,18 @@ func (r *DockerRuntime) handleDockerEvent(message events.Message, statusMessageC
...
 		return nonTerminalDockerEvent
 	case "die":
-		// TODO: Handle the difference between platform/user sidecar, not all
-		// "die"s should result in a Finish
 		if exitCode := message.Actor.Attributes["exitCode"]; exitCode == "0" {
 			statusMessageChan <- runtimeTypes.StatusMessage{
 				Status: runtimeTypes.StatusFinished,
 				Msg:    cName + " container successfully exited with 0",
 			}
+		} else if exitCode := message.Actor.Attributes["exitCode"]; exitCode == "137" && cName != mainContainerName {
```
aaronlehmann, line 2055: Nit: why not `if message.Actor.Attributes["exitCode"] == "137"`?

aaronlehmann (approved):

executor/runtime/docker/docker.go
solarkennedy: Huh. Yea I guess since we are not using the `exitCode` var later, we don't need it.

executor/runtime/docker/docker.go
```diff
@@ -2047,20 +2047,27 @@ func (r *DockerRuntime) handleDockerEvent(message events.Message, statusMessageC
...
-		// "die"s should result in a Finish
-		if exitCode := message.Actor.Attributes["exitCode"]; exitCode == "0" {
+		exitCode := message.Actor.Attributes["exitCode"]
+		if exitCode == "0" {
 			statusMessageChan <- runtimeTypes.StatusMessage{
 				Status: runtimeTypes.StatusFinished,
 				Msg:    cName + " container successfully exited with 0",
 			}
+			return isTerminalDockerEvent
+		} else if exitCode == "137" && cName != mainContainerName {
```
alekar, line 2057: A lot of us are trained to think `137 = OOMKill`. I am not sure this is the best idea.

alekar (approved):

executor/runtime/docker/docker.go
solarkennedy: I agree. :(
Hopefully in those cases the `oom` docker event will come through.

solarkennedy: Yea we need to retrain to think "137" == killed with signal 9 (128 + 9), which is *usually* killed by oom killer

PR #776, opened 2021-11-12
Allow true ipv6 containers in Titus

solarkennedy:
On true Ipv6 mode, this code crashes becase the ipv4 assignment is nil


solarkennedy:
This removes all assumptions from our code that assumes that there will be an ipv6.

I got here by continuously launching an IPv6Only container till it worked.
And now it works!

I have one cosmetic change to make on control-plane, but otherwise I think we've made it!

solarkennedy:
Oh yea IAM doesn't work because sts.amazonaws.com doesn't have a dual stack endpoint :(

MyUmmaGumma (approved):

PR #779, opened 2021-11-16
Don't set EC2_LOCAL_IPV4 in ipv6+transition mode

solarkennedy:
I think eventually the transition ip will be contained in a different
part of the vpcservice response?
When that happens, the v4 allocation in transition mode will truly be nil.

Till then, we need this thing to prevent EC2_LOCAL_IPV4 from "leaking"
the transition ip.

### DIFF
--- a/executor/runtime/types/container.go
+++ b/executor/runtime/types/container.go
@@ -1143,6 +1143,14 @@ func (c *TitusInfoContainer) parseLogStdioCheckInterval(logStdioCheckIntervalStr
 	return duration, nil
 }
 
+func isNetworkModeIPv6Only(c Container) bool {
+	if c.EffectiveNetworkMode() == titus.NetworkConfiguration_Ipv6Only.String() ||
+		c.EffectiveNetworkMode() == titus.NetworkConfiguration_Ipv6AndIpv4Fallback.String() {
+		return true
+	}
+	return false
+}
+
 func populateContainerEnv(c Container, config config.Config, userEnv map[string]string) map[string]string {
 	// Order goes (least priority, to highest priority:
 	// -Hard coded environment variables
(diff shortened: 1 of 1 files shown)


executor/runtime/types/container.go
```diff
@@ -1224,7 +1224,7 @@ func populateContainerEnv(c Container, config config.Config, userEnv map[string]
 		env["NETFLIX_NETWORK_MODE"] = netMode
 	}
 	vpcAllocation := c.VPCAllocation()
-	if a := vpcAllocation.IPV4Address(); a != nil {
+	if a := vpcAllocation.IPV4Address(); a != nil && c.EffectiveNetworkMode() != titus.NetworkConfiguration_Ipv6AndIpv4Fallback.String() {
```
troshko111, line 1227: If we're in IPv6-only mode, I'd set the `EC2IPv4EnvVarName` to `127.0.0.1` so that software which requires a valid value there does not break in an undesired way.

executor/runtime/types/container.go
```diff
@@ -1225,7 +1225,14 @@ func populateContainerEnv(c Container, config config.Config, userEnv map[string]
 	}
 	vpcAllocation := c.VPCAllocation()
 	if a := vpcAllocation.IPV4Address(); a != nil {
-		env[metadataserverTypes.EC2IPv4EnvVarName] = a.Address.Address
+		if c.EffectiveNetworkMode() == titus.NetworkConfiguration_Ipv6AndIpv4Fallback.String() {
```
troshko111, line 1228: Does this cover both IPv6-only and IPv6WithTransition?

troshko111 (approved):

executor/runtime/types/container.go
solarkennedy: No, only transition.

Are you sure with *true* v6only containers we *still* want this variable to be set with a dummy value?

I assumed that in true v6 only mode we wouldn't need any workarounds... idk because it doesn't have any fallback mechanism at all for v4?

troshko111: > Are you sure with true v6only containers we still want this variable to be set with a dummy value?

Leaning yes at this point, it would be easier to add a fake value and test different types of services' behavior when it's unset than the other way around.

sargun (approved):

executor/runtime/types/container.go
```diff
@@ -1143,6 +1143,14 @@ func (c *TitusInfoContainer) parseLogStdioCheckInterval(logStdioCheckIntervalStr
 	return duration, nil
 }
 
+func isNetworkModeIPv6Only(c Container) bool {
+	if c.EffectiveNetworkMode() == titus.NetworkConfiguration_Ipv6Only.String() ||
+		c.EffectiveNetworkMode() == titus.NetworkConfiguration_Ipv6AndIpv4Fallback.String() {
+		return true
```
solarkennedy, line 1149: Just

`return c.EffectiveNetworkMode() == titus.NetworkConfiguration_Ipv6Only.String() ||
		c.EffectiveNetworkMode() == titus.NetworkConfiguration_Ipv6AndIpv4Fallback.String()`

PR #784, opened 2021-12-02
Enbled X11 forwarding for Titus ssh

solarkennedy:
Currently users have to jump through a lot
of hoops to get X11 working (basically running a second sshd).

Enabling this allows users to x11 forward without
going through hoops.

### DIFF
--- a/executor/runtime/docker/docker_ssh.go
+++ b/executor/runtime/docker/docker_ssh.go
@@ -76,7 +76,7 @@ PasswordAuthentication no
 #GSSAPIAuthentication no
 #GSSAPICleanupCredentials yes
 
-X11Forwarding no
+X11Forwarding yes
 X11DisplayOffset 10
 PrintMotd no
 PrintLastLog no


executor/runtime/docker/docker_ssh.go
```diff
@@ -76,7 +76,7 @@ PasswordAuthentication no
 #GSSAPIAuthentication no
 #GSSAPICleanupCredentials yes
 
-X11Forwarding no
+X11Forwarding yes
```
JackOfMostTrades, line 79: Hmm, I think this config was copy/pasted to be consistent with the [EC2 base OS config](https://stash.corp.netflix.com/projects/BASEAMI/repos/nflx-sshd-config/browse/root/etc/ssh/sshd_config#73).

Should we synchronize this change with base OS?

solarkennedy: Yea sure, I can do that.
As long as security signs off on it :)

JackOfMostTrades: cc @russell-lewis in case there's any context about why we've had this off historically?

russell-lewis: We weren't running X11 anywhere at the time.

russell-lewis (approved):

JackOfMostTrades (approved):

PR #789, opened 2022-01-06
Make it easier to identify incomplete assignments during GC

sargun:
Return specific error code on incomplete assignment as opposed to generic Unknown error

### DIFF
--- a/vpc/service/assign_addresses_v3.go
+++ b/vpc/service/assign_addresses_v3.go
@@ -298,6 +298,10 @@ WHERE assignment_id = $1
 				tracehelpers.SetStatus(err, span)
 				return nil, err
 			}
+		} else {
+			err = status.Errorf(codes.FailedPrecondition, "Assignment %s is not completed", assignmentID)
+			tracehelpers.SetStatus(err, span)
+			return nil, err
 		}
 		return nil, nil
 	}


vpc/service/assign_addresses_v3.go
```diff
@@ -298,6 +298,10 @@ WHERE assignment_id = $1
 				tracehelpers.SetStatus(err, span)
 				return nil, err
 			}
+		} else {
+			err = status.Errorf(codes.FailedPrecondition, "Assignment %s is not completed", assignmentID)
+			tracehelpers.SetStatus(err, span)
+			return nil, err
 		}
 		return nil, nil
```
solarkennedy, line 306: Do we still need this line here?

vpc/service/assign_addresses_v3.go
```diff
@@ -298,6 +298,10 @@ WHERE assignment_id = $1
 				tracehelpers.SetStatus(err, span)
 				return nil, err
 			}
+		} else {
+			err = status.Errorf(codes.FailedPrecondition, "Assignment %s is not completed", assignmentID)
```
solarkennedy, line 302: This error indicates that we tried to assign an ip, but there was already some unfinished (incomplete) assignment in progress or something?

solarkennedy (approved):

vpc/service/assign_addresses_v3.go
sargun: Right, and that the assignment is either in the process of being completed, or has failed to complete.

vpc/service/assign_addresses_v3.go
sargun: yep, for the first case.

PR #792, opened 2022-01-11
Make it so detach unused branch ENI scans concurrently

sargun:
Right now ENI detaches are done on a "global" level. This means the queries
are serialized, and the detaches are serialized as well. Given that
these can both be expensive operations, it can stall out the draining
of ENIs in a given subnet.

This change makes it so that each subnet operates independently.


solarkennedy:
As long as you don't think that having N subnets => N simultaneous detaches across our fleet isn't going to backfire :)

solarkennedy (approved):

PR #794, opened 2022-01-12
GC Last IPv6 address, as RA bug no longer exists

sargun:


solarkennedy (approved):
I don't remember this bug.

sargun:
> I don't remember this bug.

Lucky you.

PR #796, opened 2022-01-13
Always make TSA handle perf syscalls

solarkennedy:
This changes makes it so *whenever* TSA is enabled for net, it also
grants perf syscalls too "for free". This is an incremental step towards
enabling perf syscalls on all titus containers.

This is only part1. Part2 will come in a differnet PR unifying the
seccomp profiles.

### DIFF
--- a/tini/src/seccomp_fd_notify.c
+++ b/tini/src/seccomp_fd_notify.c
@@ -142,13 +142,29 @@ static int install_notify_filter(void) {
 		BPF_STMT(BPF_RET | BPF_K, SECCOMP_RET_ALLOW),
 	};
 
-	struct sock_filter net_filter[] = {
+struct sock_filter net_and_perf_filter[] = {
 		/* X86_64_CHECK_ARCH_AND_LOAD_SYSCALL_NR */
 		BPF_STMT(BPF_LD | BPF_W | BPF_ABS, (offsetof(struct seccomp_data, arch))),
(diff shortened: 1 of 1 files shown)


solarkennedy:
@aloktiagi do you have opinions on this? Your new traffic steering will also adjust how this goes

MyUmmaGumma (commented):
I think we should retain different identities for perf and net syscall interception
1. These are still in flux.
2. We might want to separate out individual perf and net calls also for different system behavior.

aloktiagi:
> 

@MyUmmaGumma when enabling ioctl + network syscall interception, doesn't that create a deadlock or was it only sendmsg?

solarkennedy:
> > 
> 
> @MyUmmaGumma when enabling ioctl + network syscall interception, doesn't that create a deadlock or was it only sendmsg?

Oh crap, do we both try to intercept ioctl?
On the perf one we only intercept 2 particular ioctls (if that makes a difference)

sargun (approved):

solarkennedy:
@MyUmmaGumma / @aloktiagi can you confirm: is it ok for me to add the perf (includes ioctl intercept) filter for network interception too?

solarkennedy:
I still want this in place. My ultimate goal here is to enable `perf` on all tasks, without opting in.

To do this, it means TSA needs to use the perf filter all the time.

solarkennedy:
Closing this as we are archiving this repository.

PR #800, opened 2022-01-19
Fixed nil pointer error on null mount propagation

solarkennedy:
This allows us to not have mount propagation set at all
(nil) and get a default None propagation.

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -2002,12 +2002,22 @@ func v1ContainerHealthcheckToDockerHealthcheck(probe *v1.Probe) *container.Healt
 // v1MountPropToDockerProp converts the incoming pod mount propagation configuration
 // into something that docker can understand, using the documentation for what each means:
 // https://kubernetes.io/docs/concepts/storage/volumes/#mount-propagation
-func v1MountPropToDockerProp(fo v1.MountPropagationMode) mount.Propagation {
-	switch fo {
+func v1MountPropToDockerProp(vm v1.VolumeMount) mount.Propagation {
+	// Currently only supporting RShared and RSlave because we are mostly mounting
+	// paths in the docker daemon root.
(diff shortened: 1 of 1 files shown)


executor/runtime/docker/docker.go
```diff
@@ -2002,12 +2002,18 @@ func v1ContainerHealthcheckToDockerHealthcheck(probe *v1.Probe) *container.Healt
 // v1MountPropToDockerProp converts the incoming pod mount propagation configuration
 // into something that docker can understand, using the documentation for what each means:
 // https://kubernetes.io/docs/concepts/storage/volumes/#mount-propagation
-func v1MountPropToDockerProp(fo v1.MountPropagationMode) mount.Propagation {
-	switch fo {
+func v1MountPropToDockerProp(vm v1.VolumeMount) mount.Propagation {
+	if vm.MountPropagation == nil {
+		return ""
```
aaronlehmann, line 2007: Is the empty string a valid value for `MountPropagation`? Docker's API schema defines the field like this:

```
          Propagation:
            description: "A propagation mode with the value `[r]private`, `[r]shared`, or `[r]slave`."
            type: "string"
            enum:
              - "private"
              - "rprivate"
              - "shared"
              - "rshared"
              - "slave"
              - "rslave"
```

aaronlehmann: Looking at the code, it does seem to specifically handle an empty string. So I think this is fine...

aaronlehmann (approved):

executor/runtime/docker/docker.go
solarkennedy: Yea I was tempted to set something explicit, but seems like if a user didn't set anything, then we shouldn't set anything, and let docker handle it? idk. We'll see if the tests pass.

solarkennedy: No, I was just going to leave the docker setting unset if the user has it unset. We don't *need* to set it.

aaronlehmann: I'm confused what made the tests fail, and what you changed to fix it.

PR #801, opened 2022-01-20
Allow standalone test suite to run (degraded) as non-root

solarkennedy:
I would like to be able to have the option to run the standalone tests
in degraded mode as non-root on my linux desktop, just like on macos.

This change extends the build flags and standalone test runner to behave
the same, whether one is on darwin or linux+nonroot.


alekar (approved):
TIL about noroot in the go:build section. Fantastic.

solarkennedy:
> TIL about noroot in the go:build section. Fantastic.

I mean, I just made up that string. One must force that with go build/test -tags noroot to make it work.

PR #804, opened 2022-01-25
Make dirs on share source volume mounts if we need to TITUS-5850

solarkennedy:
For users of sharedContainerVolumeSource, we can't always depend on the
source dir existing. Ideally it would, but sometimes it is just a
scratch container, or who knows. Often it is simply outside of their
control.

For those cases, we create the dir for them as best we can.

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -2055,9 +2055,14 @@ func (r *DockerRuntime) getContainerSharedVolumeSourceMounts(mainContainerRoot s
 			if v.FlexVolume.Options["sourceContainer"] != "main" && v.FlexVolume.Options["sourceContainer"] != "" {
 				return nil, fmt.Errorf("only 'main' SharedContainerVolume volumes are supported. Volume: %+v", v)
 			}
+			source := filepath.Join(mainContainerRoot, v.FlexVolume.Options["sourcePath"])
+			err := mkdirAllInContainer(source)
+			if err != nil {
+				return mounts, fmt.Errorf("Unable to use shared volume source %s with a sourcePath of %s: %w", volumeMount.Name, v.FlexVolume.Options["sourcePath"], err)
+			}
 			m := mount.Mount{
 				Type:        "bind",
-				Source:      filepath.Join(mainContainerRoot, v.FlexVolume.Options["sourcePath"]),
+				Source:      source,
 				Target:      volumeMount.MountPath,
 				ReadOnly:    volumeMount.ReadOnly,
 				BindOptions: &mount.BindOptions{Propagation: v1MountPropToDockerProp(volumeMount)},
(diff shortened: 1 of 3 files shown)


executor/runtime/docker/docker_unsupported.go
```diff
@@ -57,3 +57,11 @@ func mountTmpfs(path string, size string) error {
 func unmountTmpfs(path string) error {
 	return nil
 }
+
+func makeBackingFsDev(home, tempdir string) error {
```
aaronlehmann, line 61: What's this for?

executor/runtime/docker/docker.go
```diff
@@ -2055,9 +2055,14 @@ func (r *DockerRuntime) getContainerSharedVolumeSourceMounts(mainContainerRoot s
 			if v.FlexVolume.Options["sourceContainer"] != "main" && v.FlexVolume.Options["sourceContainer"] != "" {
 				return nil, fmt.Errorf("only 'main' SharedContainerVolume volumes are supported. Volume: %+v", v)
 			}
+			source := filepath.Join(mainContainerRoot, v.FlexVolume.Options["sourcePath"])
+			err := mkdirAllInContainer(source)
```
aaronlehmann, line 2059: Do we also need an `mkdirAllInContainer` to create `volumeMount.MountPath` in the container that will have the mount?

sargun: Shouldn't we be able to mkdir in the tarball process that we use to create /titus and such?

solarkennedy: I assumed that the mountPath in the container itself would already exist, as that is a path controlled by the person doing the thing? But you are right maybe that is not the case.

@sargun Yes in theory that is possible, I didn't want to overload that tar function too much (seems weird to me), because here we can get more direct about what we are doing, and print better errors. In the tar function there is no such thing as "mkdir -p if a folder doesn't exist"

executor/runtime/docker/docker_unsupported.go
solarkennedy: Not related, sorry, but I had to add it so I could run the tests as non-root on my laptop.

executor/runtime/docker/docker.go
sargun: If you mkdir in that code, and the path already exists (as long as there's no conflict with directories vs. files), it should "just work" (in terms of mkdir -p style).

aaronlehmann: No, we can't assume the mountPath exists in the image. For example, the newtagen sidecar needs to be able to mount arbitrary paths from the main container to enable file sync.

solarkennedy: docker will create the destination automatically for you if it doesn't exist.
```docker run -it --mount type=bind,source="$(pwd)",target=/doesntexist busybox ls /doesntexist```

aaronlehmann:
LGTM

solarkennedy:
I'm going to close this in favor of a different approach that uses tar

PR #807, opened 2022-02-04
Updated mockvk backend.go to be pod v1 compat

solarkennedy:
I've just kinda copy/pasted in a modern backend.go and did any other
adjustment needed to make it work

### DIFF
--- a/vk/backend/backend.go
+++ b/vk/backend/backend.go
@@ -7,29 +7,32 @@ import (
 	"fmt"
 	"os"
 	"os/signal"
+	"path/filepath"
 	"sync"
 	"time"
 
 	"github.com/Netflix/metrics-client-go/metrics"
-	"github.com/Netflix/titus-executor/api/netflix/titus"
 	"github.com/Netflix/titus-executor/config"
 	titusdriver "github.com/Netflix/titus-executor/executor/drivers"
 	"github.com/Netflix/titus-executor/executor/runner"
 	runtimeTypes "github.com/Netflix/titus-executor/executor/runtime/types"
+	"github.com/Netflix/titus-executor/logger"
 	podCommon "github.com/Netflix/titus-kube-common/pod"
 	resourceCommon "github.com/Netflix/titus-kube-common/resource"
 	units "github.com/docker/go-units"
 	"github.com/golang/protobuf/proto" // nolint: staticcheck
+	"github.com/google/renameio"
 	"github.com/pkg/errors"
-	"github.com/virtual-kubelet/virtual-kubelet/log"
 	"golang.org/x/sys/unix"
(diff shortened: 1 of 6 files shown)


vk/backend/backend.go
```diff
@@ -7,29 +7,32 @@ import (
...
 	units "github.com/docker/go-units"
 	"github.com/golang/protobuf/proto" // nolint: staticcheck
+	"github.com/google/renameio"
 	"github.com/pkg/errors"
-	"github.com/virtual-kubelet/virtual-kubelet/log"
 	"golang.org/x/sys/unix"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
```
sargun, line 30: I think you might have added an extra newline here.

solarkennedy: I didn't. This newline is in backend.go in master.

vk/backend/backend.go
```diff
@@ -113,169 +116,302 @@ func state2containerState(prevState *v1.ContainerState, currState titusdriver.Ti
 	}
 }
 
-func RunWithBackend(ctx context.Context, rp runtimeTypes.ContainerRuntimeProvider, m metrics.Reporter, statuses *os.File, pod *v1.Pod, cfg config.Config) error { // nolint: gocyclo
+type Backend struct {
```
sargun, line 119: In general, it's better to not expose a struct like this because it could be accidentally instantiated by an external package incorrectly. It's better to expose an interface.

vk/backend/backend.go
```diff
@@ -113,169 +116,302 @@ func state2containerState(prevState *v1.ContainerState, currState titusdriver.Ti
...
 
-	// TODO: pick one, agreed upon resource name after migration to k8s scheduler is complete.
-	disk, _ := resource.ParseQuantity("2G")
-	for _, k := range []v1.ResourceName{v1.ResourceEphemeralStorage, v1.ResourceStorage, "titus/disk"} {
-		if v, ok := limits[k]; ok {
-			disk = v
-			break
+		data, err := base64.StdEncoding.DecodeString(containerInfoStr)
+		if err != nil {
+			return nil, errors.Wrap(err, "Could not decode containerInfo from base64")
```
sargun, line 161: Use fmt.Errorf now.

sargun (approved):

vk/backend/backend.go
solarkennedy: I don't want to drift from what we have "upstream" in master.

Keep in mind I didn't write this code, I just copy/pasted this file from what we already had.

PR #809, opened 2022-02-08
Do not skip locked subnet_usable_prefix

sargun:
Right now, if the subnet_usable_prefix table has a lock on it we skip
    the row. Unfortunately, when many queries are hitting the table, we
    can get into contention, and this can then cause branch ENI creation
    to fail.


solarkennedy:
Can you put a description (commit message) to say more about this change?

sargun:
@solarkennedy Fixed.

PR #813, opened 2022-02-09
[WIP] static allocation

aloktiagi:
### Description of the Change


sargun:
Can you split up the IPv6 address allocation code, and the subnet CIDR reservation code? (It's fine if you can't, but it'd be easier).

aloktiagi:
> Can you split up the IPv6 address allocation code, and the subnet CIDR reservation code? (It's fine if you can't, but it'd be easier).

Yeah will do that and add an integration test for this logic. For now just built up on your work.

sargun:
@aloktiagi github has draft PRs that might be easier to use here.

PR #815, opened 2022-02-10
Seccomp rebase

solarkennedy:
This new approach to our seccomp files separates the titus-specific
changes into a jsonnet-based manipulator.

Previously we were hand-editing the json files based on the upstream
docker json, but the "what did we change" information was never encoded.

If this works, this should be a more sustainable way to make changes and
"rebase" on top of modern versions of the docker seccomp json and still
keep our changes.

Also includes tests to ensure that the output polices are at least
understandable by docker.


Makefile
```diff
@@ -74,6 +74,7 @@ test-standalone: titus-agent | $(clean) $(builder)
 .PHONY: test-misc
 test-misc:
 	shellcheck --shell=sh --exclude=SC1008 ./hack/images/titus-sshd/run-titus-sshd
+	make -C executor/runtime/docker/seccomp/ test
```
sargun, line 77: `$(MAKE) -C ...`

executor/runtime/docker/seccomp/Makefile
```diff
@@ -1,10 +1,33 @@
...
 
-all: seccomp.go
+all: seccomp.go test fmt
 
+.PHONY: clean
+clean:
+	rm *.json
+
+.PHONY: fmt
+fmt:
```
sargun, line 10: This should be called by the upper fmt target.

executor/runtime/docker/seccomp/Makefile
```diff
@@ -1,10 +1,33 @@
...
+	jsonnetfmt --in-place *.jsonnet *.libsonnet
+
+test: seccomp.go
+	go test -v github.com/Netflix/titus-executor/executor/runtime/docker/seccomp
+
+stock.docker.json:
+	wget https://raw.githubusercontent.com/moby/moby/master/profiles/seccomp/default.json -O $@
```
sargun, line 17: curl?

executor/runtime/docker/seccomp/seccomp_test.go
```diff
@@ -0,0 +1,40 @@
+//go:build cgo
+// +build cgo
+
+// These tests require CGO (unfortunately) because libseccomp does
```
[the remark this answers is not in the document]
solarkennedy: They do get exercised (why it took so long to get a green build).
I'll update the comment

PR #819, opened 2022-02-14
Added new network annotations to replace the old ones

solarkennedy:
We currently set "unadorned" annotations, which served us as we
transitioned from `NetworkConfiguratoinDetails` passed out as json to
mesos.

Now, since we are on v1 pods, I would like to start using more "proper"
annotations, `network.netflix.net/address-ipv4`, etc from kube-common.

I'll leave the old annotations (eg `EniID`) until the control-plane
codebase understands the new ones.

### DIFF
--- a/executor/runtime/types/types.go
+++ b/executor/runtime/types/types.go
@@ -446,6 +446,30 @@ func (n *NetworkConfigurationDetails) ToMap() map[string]string {
 	return m
 }
 
+func (n *NetworkConfigurationDetails) ToAnnotationMap() map[string]string {
+	a := map[string]string{}
+	if n.NetworkMode != "" {
+		a[podCommon.AnnotationKeyEffectiveNetworkMode] = n.NetworkMode
+	}
+	if n.IPAddress != "" {
+		a[podCommon.AnnotationKeyIPAddress] = n.IPAddress
+	}
+	if n.EniIPAddress != "" {
+		a[podCommon.AnnotationKeyIPv4Address] = n.EniIPAddress
+	}
+	if n.EniIPv6Address != "" {
+		a[podCommon.AnnotationKeyIPv6Address] = n.EniIPv6Address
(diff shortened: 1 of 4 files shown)


tbak (approved):

executor/runtime/types/types.go
```diff
@@ -446,6 +446,30 @@ func (n *NetworkConfigurationDetails) ToMap() map[string]string {
...
+func (n *NetworkConfigurationDetails) ToAnnotationMap() map[string]string {
+	a := map[string]string{}
+	if n.NetworkMode != "" {
+		a[podCommon.AnnotationKeyEffectiveNetworkMode] = n.NetworkMode
+	}
+	if n.IPAddress != "" {
+		a[podCommon.AnnotationKeyIPAddress] = n.IPAddress
+	}
+	if n.EniIPAddress != "" {
+		a[podCommon.AnnotationKeyIPv4Address] = n.EniIPAddress
```
sargun, line 458: Can we rename this annotation? I think it's confusing.

sargun (approved):

executor/runtime/types/types.go
solarkennedy: Absolutely! 

Did you have a suggestion though? Or do you want me to mimic exactly what comes from the vpc response object?

sargun: I'm not sure. I think EIP? Or ElasticIPv4Addresss?

solarkennedy: But why do we have to have the "EIP" part? Sure in the vpcservice we use those words because it is dealing with those api calls, but for pod annotations can we use more generic terms?

solarkennedy: (that way we could reuse them for other places (ccas, localhost, etc) where they are not EIPs and it won't be weird)

PR #825, opened 2022-03-02
Change trunk eni table id column to big integer.

aloktiagi:
Change trunk eni table id column to big integer.


vpc/service/db/migrations/38_trunk_eni_id.down.sql
```diff
@@ -0,0 +1,7 @@
+START TRANSACTION;
+LOCK TABLE htb_classid IN ACCESS EXCLUSIVE MODE;
+LOCK TABLE trunk_enis IN ACCESS EXCLUSIVE MODE;
+ALTER SEQUENCE trunk_enis_id_seq AS INTEGER NO CYCLE;
```
MyUmmaGumma, line 4: Does it matter that this is back to `int` even though during the incident resolution it was marked as `bigint`?

sargun: it'll be a problem if we've exceeeded 2**31.

aloktiagi: @MyUmmaGumma thats the `down` migration file to migrate down to a previous version.

MyUmmaGumma: Understood, I was wondering if the previous version is one on the schema or the one on the box(which is bigint now)

MyUmmaGumma (approved):

PR #826, opened 2022-03-02
Enabled MOTD on titus-ssh

solarkennedy:
These changes allow the /etc/motd to be printed when one logs into
a Titus container, just like EC2.


bmoyles:
:shipit: 👍🏻

bmoyles (approved):

PR #830, opened 2022-03-09
Only do TINI_HANDOFF if we have a systemd image *and* command

solarkennedy:
Historically we've always set TINI_HANDOFF=true for *any* workload that
uses a systemd-ready image, *even if it doesn't actually run systemd!*.

The problem is that when we use a non-systemd command (`sleep`), we
still allowed sleep to be pid1, which is bad.

This change makes it so we only do TINI_HANDOFF if we are using a
systemd-ready image, *and* we are actually running a systemd command.

This should cover the 95%. If there are other paths that lead to
systemd, we can add them.

This is a departure from our existing behavior, but hopefully a good
one! It means that users of systemd-ready (baseos) images who run
something other than systemd (often with run-to-completion workloads)
will get good PID1 behavior.

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -108,6 +108,10 @@ var NoEntrypointError = &runtimeTypes.BadEntryPointError{Reason: errors.New("Ima
 // applications shall tolerate the presence of such names.
 var environmentVariableKeyRegexp = regexp.MustCompile("^[A-Za-z_][A-Za-z0-9_]*$")
 
+// possibleSystemdPaths is a list of paths that represent commands that end up running systemd.
+// We need this so that we can know if we should `exec` into them or not (TINI_HANDOFF)
+var possibleSystemdPaths = []string{"/sbin/init", "/nflx/init", "/lib/systemd/systemd"}
+
 // Poor man's OS compat
 type ucred struct {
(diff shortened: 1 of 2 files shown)


executor/runtime/docker/docker.go
```diff
@@ -108,6 +108,10 @@ var NoEntrypointError = &runtimeTypes.BadEntryPointError{Reason: errors.New("Ima
 // applications shall tolerate the presence of such names.
 var environmentVariableKeyRegexp = regexp.MustCompile("^[A-Za-z_][A-Za-z0-9_]*$")
 
+// possibleSystemdPaths is a list of paths that represent commands that end up running systemd.
+// We need this so that we can know if we should `exec` into them or not (TINI_HANDOFF)
+var possibleSystemdPaths = []string{"/sbin/init", "/nflx/init", "/lib/systemd/systemd"}
```
sargun, line 113: Should these be a configurable?

sargun (approved):

executor/runtime/docker/docker.go
solarkennedy: Eh. I don't think so. @bmoyles assures me that those are the only 3.
If we get more in the wild, I'll roll back and add it in.

PR #834, opened 2022-03-10
Fix transition assignment

sargun:


MyUmmaGumma (approved):

executor/runtime/types/types.go
```diff
@@ -491,14 +491,16 @@ func (n *NetworkConfigurationDetails) ToAnnotationMap() map[string]string {
...
+	}
 	// If we have an Elastic IP, we consider that to be the main one
 	if n.ElasticIPAddress != "" {
 		return n.ElasticIPAddress
 	}
 
-	// In transition mode, we don't want to "leak" the v4 as the primary ip
-	// But otherwise, for compatibility reasons, we consider this to be the "main" IP
-	if n.IPAddress != "" && n.NetworkMode != titus.NetworkConfiguration_Ipv6AndIpv4Fallback.String() {
+	if n.IPAddress != "" {
```
solarkennedy, line 503: Do you want this below the EIP block on line 507?

I *think* we want to treat the EIP as "the" primary ip we report for the pod if we have it

solarkennedy (approved):

executor/runtime/types/types.go
sargun: If we're in transition mode, EIPs wont work.

solarkennedy (approved):

PR #840, opened 2022-03-22
Add titus.containers.status metric

solarkennedy:
Added new `titus.containers.status` that reports once a minute based on the pod.status.containerstatus section.
It uses the specatord socket that is in the container itself, so it shows up on customer dashboards.

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -17,6 +17,7 @@ import (
 	"path/filepath"
 	"regexp"
 	"runtime"
+	"sort"
 	"strconv"
 	"strings"
 	"sync"
@@ -33,6 +34,7 @@ import (
 	vpcapi "github.com/Netflix/titus-executor/vpc/api"
 	"github.com/Netflix/titus-executor/vpc/tracehelpers"
 	vpcTypes "github.com/Netflix/titus-executor/vpc/types"
+	"github.com/docker/distribution/reference"
 	"github.com/docker/docker/api/types"
 	dockerTypes "github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/container"
@@ -108,6 +110,11 @@ var NoEntrypointError = &runtimeTypes.BadEntryPointError{Reason: errors.New("Ima
 // Portable Character Set and do not begin with a digit. Other characters may be permitted by an implementation;
 // applications shall tolerate the presence of such names.
 var environmentVariableKeyRegexp = regexp.MustCompile("^[A-Za-z_][A-Za-z0-9_]*$")
+
+// Spectatord tag keys and values are only allowed to use characters in the set -._A-Za-z0-9.
+// Others will be converted to an _.
+// See https://github.com/Netflix-Skunkworks/spectatord#allowed-characters.
+var spectatordUnexpectedTagCharRegexp = regexp.MustCompile("[^-._A-Za-z0-9]")
 
 // possibleSystemdPaths is a list of paths that represent commands that end up running systemd.
 // We need this so that we can know if we should `exec` into them or not (TINI_HANDOFF)
@@ -1627,6 +1634,8 @@ func (r *DockerRuntime) Start(parentCtx context.Context, pod *v1.Pod) (string, *
 		}
 	}
 
+	go r.reportContainerStatusMetrics(parentCtx)
+
 	// Last, we actually start the containers. And by start we mean:
 	// platform sidecars first (via docker start)
 	// user extra containers second (via docker start)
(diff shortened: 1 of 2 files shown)


executor/runtime/docker/docker.go
```diff
@@ -109,6 +111,11 @@ var NoEntrypointError = &runtimeTypes.BadEntryPointError{Reason: errors.New("Ima
 // applications shall tolerate the presence of such names.
 var environmentVariableKeyRegexp = regexp.MustCompile("^[A-Za-z_][A-Za-z0-9_]*$")
 
+// Spectatord tag keys and values are only allowed to use characters in the set -._A-Za-z0-9.
+// Others will be converted to an _.
+// See https://github.com/Netflix-Skunkworks/spectatord#allowed-characters.
+var spectatordUnexpectedTagCharRegexp = regexp.MustCompile("[^-._A-Za-z0-9]")
```
yun-wang, line 117: The current implementation replaces every illegal char to a `_`. Consecutive illegal chars become consecutive `_`s. If we want to replace consecutive illegal chars to a single `_`, I can tweak the expression to do so.

solarkennedy: I think that is probably fine.
It is not like a user can put arbitrary strings here in the first place.

executor/runtime/docker/docker.go
```diff
@@ -1640,6 +1649,137 @@ func (r *DockerRuntime) Start(parentCtx context.Context, pod *v1.Pod) (string, *
...
+
+func (r *DockerRuntime) containerStatusMetrics() []string {
+	l := log.WithField("taskID", r.c.TaskID())
+	var metricLines []string
+	// 1. Loop through all containers.
+	for _, c := range r.c.Pod().Status.ContainerStatuses {
+		// 2. Get container metadata and state.
+		metricTags := make(map[string]string)
+		metricTags["nf.container"] = c.Name
+		metricTags["nf.process"] = processNameForContainer(c.Name, r.c.Pod())
```
solarkennedy, line 1703: @copperlight  should this simply not be set if we don't have a PROCESS_NAME?
As is this code would set it to the empty string.

executor/runtime/docker/docker.go
```diff
@@ -1640,6 +1649,137 @@ func (r *DockerRuntime) Start(parentCtx context.Context, pod *v1.Pod) (string, *
...
+	if named, isNamed := ref.(reference.Named); isNamed {
+		name = reference.Path(named)
+	}
+	if digested, isDigested := ref.(reference.Digested); isDigested {
+		digest = digested.Digest().String()
+	}
+	return name, digest, nil
+}
+
+func imageTagForContainer(containerName string, c runtimeTypes.Container) string {
```
solarkennedy, line 1756: Your function here is  titus-kube-common `GetImageTagForContainer`

executor/runtime/docker/docker.go
yun-wang: Empty tag values are filtered out during sanitization: https://github.com/Netflix/titus-executor/blob/TITUS-5663/executor/runtime/docker/docker.go#L1771.

executor/runtime/docker/docker.go
```diff
@@ -1640,6 +1649,137 @@ func (r *DockerRuntime) Start(parentCtx context.Context, pod *v1.Pod) (string, *
...
+	var metricLines []string
+	// 1. Loop through all containers.
+	for _, c := range r.c.Pod().Status.ContainerStatuses {
+		// 2. Get container metadata and state.
+		metricTags := make(map[string]string)
+		metricTags["nf.container"] = c.Name
+		metricTags["nf.process"] = processNameForContainer(c.Name, r.c.Pod())
+		imageName, imageVersion, err := imageNameAndDigest(c.Image, c.Name)
+		if err != nil {
+			l.WithError(err).Warnf("Error getting image metric tags for container status metrics")
```
solarkennedy, line 1706: We should `continue` here. If we got any error here, our line will be bogus and not worth sending

executor/runtime/docker/docker.go
yun-wang: Oh nice, I can remove this function then, as long as the plan is to follow the same annotation pattern (`AnnotationKeyImageTagPrefix + cName`) for other containers. The todo for us is adding the expected annotations to the pod.

executor/runtime/docker/docker_test.go
```diff
@@ -249,3 +254,67 @@ func TestFlags(t *testing.T) {
 	_, flags := NewConfig()
 	properties.ConvertFlagsForAltSrc(flags)
 }
+
+func TestContainerStatusMetrics(t *testing.T) {
```
solarkennedy, line 258: awesome

executor/runtime/docker/docker.go
```diff
@@ -1640,6 +1649,137 @@ func (r *DockerRuntime) Start(parentCtx context.Context, pod *v1.Pod) (string, *
 	return logDir, details, statusMessageChan, err
 }
 
+func (r *DockerRuntime) reportContainerStatusMetrics(ctx context.Context) {
```
solarkennedy, line 1652: how about `periodicallyReportContainerStatusMetrics`

executor/runtime/docker/docker.go
copperlight: Multiple replacements is the correct implementation - this is how other Atlas clients handle it.

executor/runtime/docker/docker.go
copperlight: Correct, do not add the tag if it has no value. If that's removed during a sanitization step, then that should be fine.

executor/runtime/docker/docker.go
```diff
@@ -1640,6 +1649,137 @@ func (r *DockerRuntime) Start(parentCtx context.Context, pod *v1.Pod) (string, *
...
+		}
+	}
+	return ""
+}
+
+func spectatordTags(tags map[string]string) string {
+	var sTags []string
+	maxLen := 120
+	for k, v := range tags {
+		if v == "" {
```
copperlight, line 1770: Very good, removing empty tags here.

copperlight (approved):
Answered the commentary in the PR - this looks good, nice work!

executor/runtime/docker/docker.go
yun-wang: If the image string is bad, does it mean there's a bigger problem? Is it worth reporting an unhealthy status in this case?

executor/runtime/docker/docker.go
```diff
@@ -1640,6 +1649,137 @@ func (r *DockerRuntime) Start(parentCtx context.Context, pod *v1.Pod) (string, *
...
+	if err != nil {
+		if retriesLeft--; retriesLeft > 0 {
+			time.Sleep(1 * time.Second)
+			return connectToSpectatord(spectatordSocket, retriesLeft)
+		}
+	}
+	return conn, err
+}
+
+func (r *DockerRuntime) containerStatusMetrics() []string {
```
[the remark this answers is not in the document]
yun-wang: Renamed it to `generateContainerStatusSpectatordMetrics`, let me know if this works.

executor/runtime/docker/docker.go
solarkennedy: The string couldn't be so bad that we got this far (that the container launched)
Other errors will crop up.

yun-wang: Makes sense. PR is updated with `continue` here, so it's ready to go :)

executor/runtime/docker/docker.go
```diff
@@ -1640,6 +1650,128 @@ func (r *DockerRuntime) Start(parentCtx context.Context, pod *v1.Pod) (string, *
 	return logDir, details, statusMessageChan, err
 }
 
+func (r *DockerRuntime) periodicallyReportContainerStatusMetrics(ctx context.Context) {
+	l := log.WithField("taskID", r.c.TaskID())
+	defer func() {
+		if r := recover(); r != nil {
```
solarkennedy, line 1656: @sargun do you have any opinion on this technique here?
I don't want this go routine to ever crash titus-executor, but do you think this is overkill? (or has other unintended effects?)

PR #849, opened 2022-03-25
Centralized on the 'main' container name string

solarkennedy:


solarkennedy:
The real reason I needed this was in our standalone tests we were using the taskid, which is old and not reflective of how we do it in prod.

PR #852, opened 2022-03-25
  Refactor more tini code in prep to go multi-tini

solarkennedy:
This is another tiny change in preparation to move to multi-tini.

I'm just very paranoid of breaking this stuff, so I figure small PRs are
easier to read.

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -177,9 +177,7 @@ func NewDockerRuntime(ctx context.Context, m metrics.Reporter, dockerCfg Config,
 		return nil, err
 	}
 
-	// We bind-mount tini in as /sbin/docker-init to ensure we can always
-	// depend on it being there, regardless of the host docker configuration.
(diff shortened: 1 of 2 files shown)


executor/runtime/docker/docker.go
```diff
@@ -2450,40 +2452,53 @@ func (r *DockerRuntime) setupEFSMounts(parentCtx context.Context, c runtimeTypes
...
+
+	listeners := make(map[string]*net.UnixListener)
+
+	for _, c := range pod.Spec.Containers {
+		listener, err := r.setupTiniListener(c.Name)
+		if err != nil {
+			return nil, fmt.Errorf("Error while setting up tini listener for %s: %w", c.Name, err)
+		}
+		listeners[c.Name] = listener
+		go func() {
```
aaronlehmann, line 2489: Instead of having one goroutine per listener, why not have a single goroutine that loops over `listeners` once the context is cancelled?

solarkennedy: No particular reason?

PR #853, opened 2022-03-26
Move extraContainer creation to be in the Prepare phase, and use tini everywhere

solarkennedy:
Previously, we would create and start any sidecar containers in the `Start` phase.
There was no tini launching, they would just go.

With this PR:

- All containers in a pod are now created at the Prepare() phase
- All containers start (but not launched) in the Start() phase
- All containers launch at the end of the start phase via tini

No actual behavior should change, but this refactor sets us up to have titus-storage take over all these storage things (because it requires having started, but not launched containers).

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -2039,8 +2039,14 @@ func (r *DockerRuntime) k8sContainerToDockerConfigs(c *runtimeTypes.ExtraContain
 		},
 	}
 	if mainContainerRoot != "" {
-		_ = os.MkdirAll(path.Join(mainContainerRoot, "/logs"), 0700)
-		_ = os.MkdirAll(path.Join(mainContainerRoot, "/run-shared"), 0700)
+		err := os.MkdirAll(path.Join(mainContainerRoot, "/logs"), 0700)
+		if err != nil {
+			return nil, nil, nil, fmt.Errorf("Error making dir on /logs in the container: %w", err)
+		}
+		err = os.MkdirAll(path.Join(mainContainerRoot, "/run-shared"), 0700)
+		if err != nil {
+			return nil, nil, nil, fmt.Errorf("Error making dir on /run-shared in the container: %w", err)
+		}
 		mounts = append(
 			mounts, mount.Mount{
 				Type:     "bind",
@@ -2076,7 +2082,10 @@ func (r *DockerRuntime) k8sContainerToDockerConfigs(c *runtimeTypes.ExtraContain
 			// These are static certs that the metatron service will create, but we also share them
 			// between all containers in the pod.
 			if mainContainerRoot != "" {
-				_ = os.MkdirAll(path.Join(mainContainerRoot, "/metatron"), 0700)
+				err := os.MkdirAll(path.Join(mainContainerRoot, "/metatron"), 0700)
+				if err != nil {
+					return nil, nil, nil, fmt.Errorf("Error making dir on /metatron in the container: %w", err)
+				}
 				mounts = append(mounts, mount.Mount{
 					Type:     "bind",
 					Source:   path.Join(mainContainerRoot, "/metatron"),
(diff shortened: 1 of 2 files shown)


executor/runtime/docker/docker.go
```diff
@@ -2041,6 +2039,8 @@ func (r *DockerRuntime) k8sContainerToDockerConfigs(c *runtimeTypes.ExtraContain
 		},
 	}
 	if mainContainerRoot != "" {
+		_ = os.MkdirAll(path.Join(mainContainerRoot, "/logs"), 0700)
```
aaronlehmann, line 2042: I think we should for an error and at least log it. Same with the other `MkdirAll` calls below.

executor/runtime/docker/docker.go
```diff
@@ -1175,16 +1176,36 @@ func (r *DockerRuntime) Prepare(ctx context.Context, pod *v1.Pod) error { // nol
...
 		goto error
 	}
 	ctx = logger.WithField(ctx, "containerID", r.c.ID())
-	logger.G(ctx).Info("Container successfully created")
+	logger.G(ctx).Info("Main Container successfully created")
+
+	if len(pod.Spec.Containers) > 1 {
+		mainContainerRoot, err = r.inspectAndGetMainContainerRoot(ctx)
+		if err != nil {
+			goto error
```
aaronlehmann, line 1191: A more idiomatic pattern for this:

```go
func (r *DockerRuntime) Prepare(ctx context.Context, pod *v1.Pod) (retErr error) {

        ...

        defer func() {
                if retErr != nil && !docker.IsErrNotFound(retErr) { // if the difference in behavior for the "not found" error is intentional
                        tracehelpers.SetStatus(retErr, span)
                        log.Error("Unable to create main container: ", retErr)
                        r.metrics.Counter("titus.executor.dockerCreateContainerError", 1, nil)
                }
        }()
```

Then just return errors as needed.

executor/runtime/docker/docker.go
```diff
@@ -1593,12 +1623,13 @@ func (r *DockerRuntime) Start(parentCtx context.Context, pod *v1.Pod) (string, *
...
 	}
 
-	mainTiniConn, err := r.waitForTiniConnection(ctx, listeners[runtimeTypes.MainContainerName])
+	tiniConns, err := r.waitForTiniConnections(ctx, listeners)
 	if err != nil {
 		eventCancel()
 		err = fmt.Errorf("container prestart error: %w", err)
 		return "", nil, statusMessageChan, err
 	}
+	mainTiniConn := tiniConns[runtimeTypes.MainContainerName]
```
aaronlehmann, line 1632: This will be nil in the Docker for Mac case. Is that okay?

executor/runtime/docker/docker.go
```diff
@@ -2524,6 +2528,25 @@ func (r *DockerRuntime) waitForTiniConnection(ctx context.Context, l *net.UnixLi
 	}
 }
 
+func (r *DockerRuntime) waitForTiniConnections(ctx context.Context, listeners map[string]*net.UnixListener) (map[string]*net.UnixConn, error) {
+	if listeners == nil {
```
aaronlehmann, line 2532: I'm not sure why this check makes any difference. Without it, this function would just return an empty map, which is equivalent to a nil map as far as reads are concerned.

executor/runtime/docker/docker.go
solarkennedy: Yea I guess. 
These mkdirs are kinda just "best effort", but I suppose it would make for a clearer error code if we couldn't mkdir.
These are temporary. My next PR to make titus-storage handle this will be doing these things (granted it should also handle errors!)

executor/runtime/docker/docker.go
solarkennedy: I don't know about the history behind the `goto` thing.
I agree I find it a bit surprising.

Let me propose this on a different PR.

executor/runtime/docker/docker.go
solarkennedy: Not that I use Linux, eh? 
I think was probably the only one running the standalone tests locally on my laptop (I originally made the docker-for-mac compatibility things).

But yes, if nil the other functions should just fail early. I think I've already sprinkled those in.

executor/runtime/docker/docker.go
solarkennedy: I agree

PR #854, opened 2022-03-28
Only run systemd stuff if the effective cmd/entry is systemd

solarkennedy:
We are still treating tasks that are not running systemd as if they
were, which means we are TINI_HANDOFF=true when we should not!

This change makes us have stricter requirements before we consider an
image to be running systemd, so strict that the test (running /bin/bash)
now fails. I removed it because I think we might be better off with the
actual integration test, or some other approach.


bmoyles (approved):
lgtm

solarkennedy:
There is still an edge case here where the entrypoint could be bash and their cmd might be a string that execs into systemd somehow. I'm going to just leave that as unsupported for now.

PR #855, opened 2022-03-28
Removed use of goto in Prepare()

solarkennedy:
The only thing I don't understand is why we don't do cancel stuff when we get a `docker.IsErrNotFound(`.

Honestly I think this was probably a bug, so I removed that special case.

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -965,7 +965,7 @@ func (r *DockerRuntime) createVolumeContainer(ctx context.Context, containerName
 }
 
 // Prepare host state (pull images, create fs, create container, etc...)
-func (r *DockerRuntime) Prepare(ctx context.Context, pod *v1.Pod) error { // nolint: gocyclo
+func (r *DockerRuntime) Prepare(ctx context.Context, pod *v1.Pod) (err error) { // nolint: gocyclo
 	var volumeContainers []string
 	var sharedMountDirs []string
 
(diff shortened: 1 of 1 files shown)


executor/runtime/docker/docker.go
```diff
@@ -978,35 +978,39 @@ func (r *DockerRuntime) Prepare(ctx context.Context, pod *v1.Pod) error { // nol
...
-		myImageInfo         *types.ImageInspect
-		dockerCfg           *container.Config
-		hostCfg             *container.HostConfig
-		systemServices      []*runtimeTypes.ServiceOpts
-		size                int64
-		runTmpfs            string
-		err                 error
+		myImageInfo *types.ImageInspect
+		imageSize   int64
+		err         error
```
aaronlehmann, line 983: It's better to use a named return value, so the code that returns from the function doesn't have to return this exact `err` variable for the defer to work.

executor/runtime/docker/docker.go
```diff
@@ -1144,14 +1148,13 @@ func (r *DockerRuntime) Prepare(ctx context.Context, pod *v1.Pod) error { // nol
...
 			// or a bad deploy, or heck just a typo in the name or something
 			if sidecarConfig.Required {
-				err = fmt.Errorf("Unable to get volume container of required sidecar %s", sidecarConfig.ServiceName)
-				goto error
-			} else {
-				// If the ContainerName is still uninitialized, but this service is not required, then it is
-				// ok to just log about it, but it should not be included on the list of volume containers to use
-				// (because it doesn't exist)
-				logger.G(ctx).Warnf("Skipping volume container of optional sidecar %s", sidecarConfig.ServiceName)
+				return fmt.Errorf("Unable to get volume container of required sidecar %s", sidecarConfig.ServiceName)
```
aaronlehmann, line 1151: This is an example of something that won't work without the named return value.

solarkennedy: Oh right. Yuck.

aaronlehmann (approved):

executor/runtime/docker/docker.go
```diff
@@ -978,35 +978,38 @@ func (r *DockerRuntime) Prepare(ctx context.Context, pod *v1.Pod) error { // nol
...
+		myImageInfo *types.ImageInspect
+		imageSize   int64
 	)
 	dockerCreateStartTime := time.Now()
 	group := groupWithContext(ctx)
+
+	defer func() {
+		if err != nil {
+			tracehelpers.SetStatus(err, span)
+			log.Error("Unable to create main container: ", err)
```
sargun, line 990: Might as well do WithError here.

executor/runtime/docker/docker.go
```diff
@@ -1163,55 +1165,49 @@ func (r *DockerRuntime) Prepare(ctx context.Context, pod *v1.Pod) error { // nol
...
+	containerCreateBody, err := r.client.ContainerCreate(ctx, dockerCfg, hostCfg, nil, r.c.TaskID())
 	r.c.SetID(containerCreateBody.ID)
 
 	r.metrics.Timer("titus.executor.dockerCreateTime", time.Since(dockerCreateStartTime), r.c.ImageTagForMetrics())
 	if docker.IsErrNotFound(err) {
 		return &runtimeTypes.RegistryImageNotFoundError{Reason: err}
 	}
 	if err != nil {
-		goto error
+		return err
```
sargun, line 1186: can we use tracehelpers.setstatus on these error messages?

sargun (approved):
I'm not sure what you mean by: 

> The only thing I don't understand is why we don't do cancel stuff when we get a docker.IsErrNotFound(.

executor/runtime/docker/docker.go
solarkennedy: I don't know much about what that does, but we already have that on the `defer`, isn't that enough?

solarkennedy:
Rebased

executor/runtime/docker/docker.go
solarkennedy: For example, any `return err` that is non nil in this function should call

PR #856, opened 2022-03-30
Titus 5894

solarkennedy:
### Description of the Change


solarkennedy:
While this code works (barely), I'm closing it in favor of #857, which has @sargun 's suggested "titus-inits" approach of getting pid1dirs.

PR #857, opened 2022-03-30
Enable titus-storage to mount /mnt-shared on multi-container workloads TITUS-5894

solarkennedy:
Now that we have tini on all containers, I would like to slowly offload all storage duties of titus-executor into the `titus-storage` titus-system-service.

This PR does a lot of prerequisite work (pod.json, $task/inits/$cname, etc) to make it so titus-storage has all the data it needs to actually mount stuff.

Then I added code for adding /mnt-shared, which will only be activated on multi-container workloads.

In my testing I found it was a bit racy on shutdown, depending on how the container was killed, so I added a big cleanup function to ensure we never leave behind any mounts.

Lastly, I'm I had to add a `ConditionPathExists` to suppress the running of `titus-storage` during our tests (which only run titus-standalone, and there is no pod.json at all). If you think this is now good, I can spend more time making our standalone_tests actually write out pod files like VK does. Or if you think that our integration tests at another layer are enough, that could be fine too.

### DIFF
--- a/.buildkite/build.sh
+++ b/.buildkite/build.sh
@@ -30,6 +30,8 @@ redact_secrets() {
 # For CI/CD, we touch some files to make them new, so that make doesn't think they need
 # to be re-generated
 touch vpc/service/db/migrations/bindata.go
+touch executor/runtime/docker/seccomp/*.json
+touch executor/runtime/docker/seccomp/seccomp.go
 
 log "Building executor"
 make clean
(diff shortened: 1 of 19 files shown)


solarkennedy:
Alright, I did it. I added extra commits to add a functional test and to ensure the titus-storage system service gets run under CI-CD, and it passes!

I'm happy where this ended up, so I'm going to merge before I break anything else, but I will sleep at night knowing that if I do break this feature somehow in the future, buildkite will tell us (and I won't have to wait to deploy to staging).

PR #859, opened 2022-03-31
Add platform.sidecars.instance metric

yun-wang:
Add a new `platform.sidecars.instance` metric. The metric is only emitted for tasks started with platform sidecars. It's reported once a minute, same as the `titus.containers.status` metric (#840). This metric can be used to see:
- How many applications are using a platform sidecar.
- For each platform sidecar, what are the channels and the channel definitions used.

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -1664,7 +1664,7 @@ func (r *DockerRuntime) Start(parentCtx context.Context, pod *v1.Pod) (string, *
 		}
 	}
 
-	go r.periodicallyReportContainerStatusMetrics(parentCtx)
+	go r.periodicallyReportTaskMetrics(parentCtx)
 
 	// Last, we actually launch the containers. And by launch we mean we tell tini `L` and it makes things go
 	err = r.launchAllContainers(ctx, tiniConns)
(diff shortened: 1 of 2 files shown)


solarkennedy (approved):

executor/runtime/docker/docker.go
```diff
@@ -1781,6 +1783,32 @@ func imageNameAndDigest(image, containerName string) (name, digest string, err e
...
+		releaseSuffix := fmt.Sprintf(".%s/%s", pod.AnnotationKeySuffixSidecars, pod.AnnotationKeySuffixSidecarsRelease)
+		if !strings.HasSuffix(k, releaseSuffix) {
+			continue
+		}
+		sidecar := strings.TrimSuffix(k, releaseSuffix)
+		release := strings.Split(v, "/") // Expected format: $channel/$channelDefID
+		if len(release) != 2 || release[0] == "" || release[1] == "" {
+			l.Warnf("Unexpected release value for platform sidecar %q: %q", sidecar, v)
+			continue
+		}
```
aaronlehmann, line 1799: I'm wondering if this annotation parsing code should live in `titus-kube-common`.

yun-wang: I'm not aware of other Go code that needs to parse the release annotation. I think it's okay to leave the parsing logic here for now. We can definitely move it to `titus-kube-common` if code sharing is beneficial.

aaronlehmann: I think in general we've tried to centralize our annotation handling code there, to abstract usage of annotations instead of spreading parsing/formatting logic over many repos. @solarkennedy can jump in and correct me if I'm wrong, though.

solarkennedy: Yea I guess that would be a nice to have. I'm ok either way. Certainly as soon as there is a second place that needs this parsing, it should be shared.
This is certainly non-trivial parsing of the constants.

PR #863, opened 2022-04-05
Netflix runc version to support hybrid cgroup hierarchy

aloktiagi:
Netflix runc version to support hybrid cgroup hierarchy

### DIFF
--- a/executor/runtime/types/container.go
+++ b/executor/runtime/types/container.go
@@ -31,7 +31,8 @@ const (
 	jobTypeLabelKey      = "com.netflix.titus.job.type"
 
 	// DefaultOciRuntime is the default oci-compliant runtime used to run system services
-	DefaultOciRuntime = "runc"
+	// runc-alt is netflix runc version to support hybrid cgroup hierarchy
+	DefaultOciRuntime = "runc-alt"
 	MainContainerName = "main"
 )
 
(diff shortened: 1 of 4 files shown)


solarkennedy:
With this in place, GPU tasks won't get cgroup v2, will that be a problem?

solarkennedy:
I don't know what it will take to get this to work in integration tests.
I think you will have to install the alt runtime in the Dockerfile for the tests?

sargun:
Yeah, I think we should install the runc-alt for the integration tests to make sure out mix and match works.

aloktiagi:
Would this require an update to the titusoss/titus-agent image?

aloktiagi:
Ok. The last commit made it work.

aloktiagi:
@sargun @solarkennedy how do you feel about this? ok to go ahead?

solarkennedy:
Can you rebase onto a single commit?

I'm going to be OOO for the next week, so if you want to push it needs to be in coordination with @sargun . 
Also the upgrade of titus-executor needs to be with the same commit as a change to titus-agent to include the equivalent change to the docker daemon.json and nflx-runc package.

sargun (approved):

aloktiagi:
> Can you rebase onto a single commit?
> 
> I'm going to be OOO for the next week, so if you want to push it needs to be in coordination with @sargun . Also the upgrade of titus-executor needs to be with the same commit as a change to titus-agent to include the equivalent change to the docker daemon.json and nflx-runc package.

Sounds good. Will update the titus-agent PR with the required titus-executor version.

sargun:
I don't think we will need this anymore.

aloktiagi:
> I don't think we will need this anymore.

Correct! closing.

PR #867, opened 2022-04-08
Removed legacy network annotations

solarkennedy:
This cleans up our pods a little bit, removing "legacy" old style
annotations like `EniId` in favor of the new style that comes from
kube-common.

This can be merged and shipped after
https://github.com/Netflix/titus-control-plane/pull/1257
is fully deployed.

### DIFF
--- a/executor/runtime/types/types.go
+++ b/executor/runtime/types/types.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"regexp"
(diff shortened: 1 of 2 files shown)


MyUmmaGumma (approved):

solarkennedy:
I think we've made it, and I don't believe these legacy annotations are in use by the CP anymore.

PR #868, opened 2022-04-12
Use timeout context to work around docker hangs

solarkennedy:
Sometimes docker calls will hang. If that happens we don't want to leave
hanging titus-executors around.

We don't want disconnected containers either though?

I don't exactly know if I'm using context() correctly.

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -2959,6 +2959,8 @@ func (r *DockerRuntime) Kill(ctx context.Context, wasKilled bool) error { // nol
 		containerStopTimeout = time.Second * time.Duration(*killWait)
 	}
 	cStopPtr := &containerStopTimeout
+	ctxWithTimeout, cancel := context.WithTimeout(context.Background(), containerStopTimeout)
+	defer cancel()
 
 	if wasKilled {
 		// We're being told by the API to stop, so use the configured stop timeout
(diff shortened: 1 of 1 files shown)


executor/runtime/docker/docker.go
```diff
@@ -2959,6 +2959,8 @@ func (r *DockerRuntime) Kill(ctx context.Context, wasKilled bool) error { // nol
 		containerStopTimeout = time.Second * time.Duration(*killWait)
 	}
 	cStopPtr := &containerStopTimeout
+	ctxWithTimeout, cancel := context.WithTimeout(ctx, containerStopTimeout)
```
sargun, line 2962: I don't think we want to descend from that context because it. can be cancelled. I think we would want to do `context.WithTimeout(context.Background(), ...)`. This creates a second-order problem that what if it fails?

executor/runtime/docker/docker.go
```diff
@@ -2984,7 +2986,7 @@ func (r *DockerRuntime) Kill(ctx context.Context, wasKilled bool) error { // nol
 	}
 
 	logger.G(ctx).Debug("Stopping main container")
-	if err := r.client.ContainerStop(context.TODO(), r.c.ID(), cStopPtr); err != nil {
+	if err := r.client.ContainerStop(ctxWithTimeout, r.c.ID(), cStopPtr); err != nil {
 		r.metrics.Counter("titus.executor.dockerStopContainerError", 1, nil)
 		log.Errorf("container %s : stop %v", r.c.TaskID(), err)
```
sargun, line 2991: While we're at it, Add WithError?

executor/runtime/docker/docker.go
solarkennedy: Yea...
I'm not super sure. Seems like we gotta fail *eventually*, but yea, we will leave disconnected containers in their wake?

executor/runtime/docker/docker.go
```diff
@@ -2969,33 +2971,33 @@ func (r *DockerRuntime) Kill(ctx context.Context, wasKilled bool) error { // nol
 		logger.G(ctx).Info("Stopping+Cleaning up containers because they finished or died")
 	}
 
-	if containerJSON, err := r.client.ContainerInspect(context.TODO(), r.c.ID()); docker.IsErrNotFound(err) {
+	if containerJSON, err := r.client.ContainerInspect(ctxWithTimeout, r.c.ID()); docker.IsErrNotFound(err) {
```
gabrielhartmann, line 2974: Timing out here is good, but we should also retry.  I have observed at least one case in the wild where the HTTP request hung forever, but subsequent manual calls succeeded.  So transient failures are a real possibility I think.

gabrielhartmann (approved):

solarkennedy:
> I'm torn here. If we timeout we'll probably mask a real issue that hopefully some other trigger will catch. But it is probably less annoying to users to actually see some task status, vs. whatever we show when this call just hangs?

Yea I agree, I've been hesitant to push the green button because I'm afraid this would cause more issues than it solves, even if we had retries.

solarkennedy:
Closing this as we are archiving this repository.

PR #871, opened 2022-04-22
Ignore extra containers dying with 143

solarkennedy:
An exit code of 143 on an extra container means it was killed (99% of
the time).

This usually happens when systemd shuts down, not because the extra
container did anything wrong.

So for this reason, this PR ignores that exit code so that the "real"
return code for the pod (presumably from the main container) will be
revealed.


bmoyles (approved):
LGTM!

aaronlehmann (approved):

PR #873, opened 2022-04-25
Fix segfault on mkdir in containers

solarkennedy:
Sometimes we need to create+open folders that happen to be on a tmpfs
(/run) on systemd containers.

But then we get this:
https://bugzilla.kernel.org/show_bug.cgi?id=183461

The solution seems to be to just setuid before mkdir.

### DIFF
--- a/mount/titus-mount-container-to-container.c
+++ b/mount/titus-mount-container-to-container.c
@@ -1,22 +1,22 @@
 #define _GNU_SOURCE
-#include <errno.h>        // for EBADF
-#include <stdio.h>        // for perror, fprintf, snprintf, stderr, size_t
-#include <stdlib.h>       // for exit, getenv, WEXITSTATUS, WIFEXITED
+#include <errno.h> // for EBADF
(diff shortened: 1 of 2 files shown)


mount/titus-mount-container-to-container.c
```diff
@@ -91,9 +91,14 @@ static int setup_fsfd_in_namespaces(int sk, int src_nsfd, const char *src_path)
 		return 1;
 	}
 
+	setuid(0);
+	setgid(0);
```
aaronlehmann, line 95: Should we check for an error here?

aaronlehmann (approved):

mount/titus-mount-container-to-container.c
solarkennedy: Ehhh.
I'm not 100% sure.
The whole point of the mkdir thing is in theory to be `mkdir || true`.
I don't really care about errors, *unless* it actually makes something not work.
I'll skip them on these lines for now.

PR #874, opened 2022-04-26
Add api to de-allocate static ip addresses

aloktiagi:
Add api to de-allocate static ip addresses


MyUmmaGumma (commented):
Should these be separate transactions? (I am not sure if there are transaction timeouts, these would be three AWS API calls)

aloktiagi:
> Should these be separate transactions? (I am not sure if there are transaction timeouts, these would be three AWS API calls)

@MyUmmaGumma could you point it out in the code?

vpc/service/allocate.go
```diff
@@ -436,6 +431,79 @@ func (vpcService *vpcService) ValidateAllocation(ctx context.Context, req *titus
...
+	err = rows.Scan(&region, &accountid, &v4prefix, &v6prefix)
+	if err == sql.ErrNoRows {
+		return nil, errAllocationNotFound
+	}
+	if err != nil {
+		return nil, errors.Wrap(err, "Could not deserialize row")
+	}
+	rows.Close()
+
+	session, err := vpcService.ec2.GetSessionFromAccountAndRegion(ctx, ec2wrapper.Key{
```
MyUmmaGumma, line 472: This and the following two calls to DeleteSubnetCidrReservation

MyUmmaGumma (approved):

PR #877, opened 2022-05-04
kernel version check for seccomp flags

aloktiagi:
kernel version check for seccomp flags

### DIFF
--- a/tini/src/seccomp_fd_notify.c
+++ b/tini/src/seccomp_fd_notify.c
@@ -22,11 +22,15 @@
 #include <stddef.h>
 #include <sched.h>
 #include <sys/prctl.h>
+#include <linux/version.h>
+#include <sys/utsname.h>
 #include "seccomp_fd_notify.h"
 
 #ifndef CHILD_SIG
 #define CHILD_SIG SIGUSR1
 #endif
+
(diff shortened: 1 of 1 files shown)


tini/src/seccomp_fd_notify.c
```diff
@@ -248,8 +274,16 @@ static int install_notify_filter(void) {
 	/* Only one listening file descriptor can be established. An attempt to
 	   establish a second listener yields an EBUSY error. */
 
+	if (get_kernel_version(&system_kernel_version) == -1) {
+		return -1;
+	}
+
+	if (system_kernel_version >= WAIT_KILLABLE_SECCOMP_KERNEL_VERSION) {
```
solarkennedy, line 281: Can you just >= on a tuple of integers like that? How does C do a comparison with `(5, 15, 35)` ?

aloktiagi: > Can you just >= on a tuple of integers like that? How does C do a comparison with `(5, 15, 35)` ?

So for this macro `#define WAIT_KILLABLE_SECCOMP_KERNEL_VERSION KERNEL_VERSION(5, 15, 35)` `KERNEL_VERSION` expands to `(((a) << 16) + ((b) << 8) + (c))`

PR #879, opened 2022-05-06
Upgraded to k8s client libs 1.21.12

solarkennedy:


solarkennedy:
I may defer this one. I don't think titus-executor even makes any k8s api calls, just has golang objects. 
I don't think this upgrade is critical.
And I think I'm going to have to yak-shave and get it up to golang 1.17.

PR #880, opened 2022-05-14
Move NFS mounting logic to titus-storage from docker runtime

sdab:
### Description of the Change

Migrates the logic for mounting NFS volumes from the docker runtime to titus-storage sidecar. Tested with titus-integration-tests:tests/tests_efs.py (went from passing -> failing -> passing). Change should have no functional difference.

### DIFF
--- a/cmd/titus-storage/nfs.go
+++ b/cmd/titus-storage/nfs.go
@@ -1,0 +1,84 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"time"
+
+	executorDocker "github.com/Netflix/titus-executor/executor/runtime/docker"
+	corev1 "k8s.io/api/core/v1"
+)
+
+const (
+	mountCmd     = "/apps/titus-executor/bin/titus-mount-nfs"
+	mountTimeout = 10 * time.Second
+	// MS_RDONLY indicates that mount is read-only
+	MS_RDONLY = 1 // nolint: golint
+)
+
+// setupNFSMount calls out to the titus-mount-nfs command to set up a single
+// EFS mount.
+func setupNFSMount(parentCtx context.Context, taskID, cname string, vol *corev1.Volume, vm *corev1.VolumeMount) error {
+	baseMountOptions := []string{"vers=4.1,rsize=1048576,wsize=1048576,timeo=600,retrans=2"}
+	ctx, cancel := context.WithTimeout(parentCtx, mountTimeout)
+	defer cancel()
+
+	pidDir := executorDocker.GetTitusInitsPath(taskID, cname)
+	cmd := exec.CommandContext(ctx, mountCmd)
+	flags := 0
+
+	if vm.ReadOnly {
+		flags = flags | MS_RDONLY
+	}
+	mountOptions := append(
+		baseMountOptions,
+		fmt.Sprintf("fsc=%s", taskID),
+		fmt.Sprintf("source=[%s]:%s", vol.NFS.Server, filepath.Clean(vol.NFS.Path)),
+	)
+	cmd.Env = []string{
+		fmt.Sprintf("TITUS_PID1_DIR=%s", pidDir),
+		fmt.Sprintf("MOUNT_TARGET=%s", filepath.Clean(vm.MountPath)),
+		fmt.Sprintf("MOUNT_NFS_HOSTNAME=%s", vol.NFS.Server),
+		fmt.Sprintf("MOUNT_FLAGS=%d", flags),
+		fmt.Sprintf("MOUNT_OPTIONS=%s", strings.Join(mountOptions, ",")),
+	}
+
+	stdoutStderr, err := cmd.CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("Mount failure: %+v: %s", cmd, string(stdoutStderr))
+	}
+
+	return nil
+}
+
+// setupNFSMounts sets up all of the mounts across all containers for the given pod.
+func setupNFSMounts(parentCtx context.Context, taskID string, pod *corev1.Pod) error {
+	nameToMount := map[string]corev1.Volume{}
+
+	for _, vol := range pod.Spec.Volumes {
+		if vol.VolumeSource.NFS == nil {
+			continue
+		}
+		nameToMount[vol.Name] = vol
+	}
+
+	for _, container := range pod.Spec.Containers {
+		for i := range container.VolumeMounts {
+			vm := container.VolumeMounts[i]
+			vol, ok := nameToMount[vm.Name]
+			if !ok {
+				continue
+			}
+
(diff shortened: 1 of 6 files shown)


cmd/titus-storage/nfs.go
```diff
@@ -0,0 +1,87 @@
...
+
+// setupNFSMount calls out to the titus-mount-nfs command to set up a single
+// EFS mount.
+func setupNFSMount(parentCtx context.Context, taskID, cname string, vol *corev1.Volume, vm *corev1.VolumeMount) error {
+	baseMountOptions := []string{"vers=4.1,rsize=1048576,wsize=1048576,timeo=600,retrans=2"}
+	ctx, cancel := context.WithTimeout(parentCtx, mountTimeout)
+	defer cancel()
+
+	pidDir := executorDocker.GetTitusInitsPath(taskID, cname)
+	cmd := exec.CommandContext(ctx, "/apps/titus-executor/bin/titus-mount-nfs")
```
solarkennedy, line 32: Can you put this path in as `const` ?

executor/runtime/docker/docker.go
```diff
@@ -1499,8 +1485,8 @@ func maybeConvertIntoBadEntryPointError(err error) error {
 	return err
 }
 
-// setupEFSAndLogs connects to tini, and also sets up NFS mounts
-func (r *DockerRuntime) setupEFSandLogsAndMisc(ctx context.Context, typedConn *net.UnixConn, c runtimeTypes.Container) (string, error) {
+// setupLogsAndMisc connects to tini
+func (r *DockerRuntime) setupLogsAndMisc(ctx context.Context, typedConn *net.UnixConn, c runtimeTypes.Container) (string, error) {
```
solarkennedy, line 1489: Really happy to see this further reduced.

executor/runtime/types/sidecars.go
```diff
@@ -215,9 +215,9 @@ func shouldStartLogViewer(cfg *config.Config, c Container) bool {
 }
 
 func shouldStartTitusStorage(cfg *config.Config, c Container) bool {
-	// Currently titus-storage only supports EBS and /mnt-shared storage
+	// Currently titus-storage sets up NFS, EBS, and /mnt-shared storage
 	// which is currently only available on multi-container workloads
-	return c.EBSInfo().VolumeID != "" || len(c.Pod().Spec.Containers) > 1
+	return len(c.NFSMounts()) > 0 || c.EBSInfo().VolumeID != "" || len(c.Pod().Spec.Containers) > 1
```
solarkennedy, line 220: This is great. Maybe on the next pass we can remove the whole `c.NFSMounts()` idea entirely!

solarkennedy:
Excellent.
Can you confirm *all* the tests pass with the `test_efs.py` file, not just the simple singular mount one?

solarkennedy (approved):

executor/runtime/types/sidecars.go
sdab: Do you mean removing nfs mounts or just from the pod container object?

cmd/titus-storage/nfs.go
```diff
@@ -0,0 +1,87 @@
...
+	"context"
+	"fmt"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"time"
+
+	executorDocker "github.com/Netflix/titus-executor/executor/runtime/docker"
+	corev1 "k8s.io/api/core/v1"
+	//"github.com/Netflix/titus-executor/logger"
```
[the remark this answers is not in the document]
sdab: Good catch. re:logging, the returned error contains all the info we'd want to log anyway and I figure it wasnt needed in the previous code anyway so I didn't see the need to add any here.

cmd/titus-storage/nfs.go
sdab: done

executor/runtime/types/sidecars.go
solarkennedy: Removing the `NFSMounts()` function itself. 
Don't do it in this PR though.

sdab:
Ran all tests and got them to pass after fixing a net namespace bug. I also fixed the mount timeout to be less than the sidecar startup timeout.

solarkennedy:
> Ran all tests and got them to pass after fixing a net namespace bug. I also fixed the mount timeout to be less than the sidecar startup timeout.

Excellent!

cmd/titus-storage/nfs.go
```diff
@@ -0,0 +1,85 @@
...
+		fmt.Sprintf("MOUNT_NFS_HOSTNAME=%s", vol.NFS.Server),
+		fmt.Sprintf("MOUNT_FLAGS=%d", flags),
+		fmt.Sprintf("MOUNT_OPTIONS=%s", strings.Join(mountOptions, ",")),
+	}
+
+	stdoutStderr, err := cmd.CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("Mount failure: %+v: %s", cmd, string(stdoutStderr))
+	}
+	cancel()
```
solarkennedy, line 53: I think with this new setup, we don't need to call cancel here.
It will get called automatically on the next line! (return)

mount/titus-mount-nfs.c
```diff
@@ -185,22 +185,27 @@ static void do_fsconfigs(int fsfd, char *options)
...
-static int setup_fsfd_in_namespaces(int sk, int pidfd)
+static int setup_fsfd_in_namespaces(int sk, int nsfd)
 {
 	int fsfd, ret;
-	ret = setns(pidfd, CLONE_NEWUSER);
+	int usernsfd = openat(nsfd, "user", O_RDONLY);
+	int mnt_fd = openat(nsfd, "mnt", O_RDONLY);
+	int net_fd = openat(nsfd, "net", O_RDONLY);
+	assert(usernsfd != -1);
+	assert(mnt_fd != -1);
```
solarkennedy, line 195: one more assert on the net_fd

mount/titus-mount-nfs.c
```diff
@@ -331,7 +340,7 @@ int main(int argc, char *argv[])
 	/* Now we can do the fs_config calls and actual mount */
 	do_fsconfigs(fsfd, final_options);
 	mkdir_p(target);
-	mount_and_move(fsfd, target, pidfd, flags_ul);
+	mount_and_move(fsfd, target, flags_ul);
```
solarkennedy, line 343: Haha, nice, we were not even using pidfd anyway

solarkennedy (approved):
fix + ship

mount/titus-mount-nfs.c
sdab: yup

cmd/titus-storage/nfs.go
sdab: good point

mount/titus-mount-nfs.c
sdab: good catch

PR #886, opened 2022-05-26
add support for SystemDNS local resolver

acm1:
adds SystemDNS System Service support

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -375,8 +375,12 @@ func (r *DockerRuntime) computeDNSServers() []string {
 		// the burden on TSA for ipv4 udp traffic
 		return []string{"fd00:ec2::253"}
 	case titus.NetworkConfiguration_Ipv6Only.String():
-		// True ipv6 only means we really can only have a v6
-		// resolver
+		// If we're ipv6-only and running the SystemDNS local resolver
+		// which provides DNS64, we'll point to that
+		if runtimeTypes.ShouldStartSystemDNS(&r.cfg, r.c) {
+			return []string{"127.0.0.53"}
+		}
+		// otherwise return the EC2 IPv6 resolver
 		return []string{"fd00:ec2::253"}
 	default:
 		// Any other situation means we can return the classic v4 resolver
--- a/executor/runtime/types/sidecars.go
+++ b/executor/runtime/types/sidecars.go
@@ -14,6 +14,7 @@ const (
 	SidecarServiceServiceMesh      = "servicemesh"
 	SidecarServiceSshd             = "sshd"
 	SidecarServiceSpectatord       = "spectatord"
+	SidecarServiceSystemDNS        = "systemdns"
 	SidecarServiceTracingCollector = "tracing-collector"
 	SidecarServiceAtlasTitusAgent  = "atlas-titus-agent"
 	SidecarTitusStorage            = "titus-storage"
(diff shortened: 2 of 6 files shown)


executor/runtime/types/sidecars.go
```diff
@@ -33,6 +34,15 @@ var systemServices = []ServiceOpts{
 		UnitName:     "titus-sidecar-seccomp-agent",
 		Required:     true,
 		EnabledCheck: shouldStartTitusSeccompAgent,
+		Volumes: map[string]struct{}{
+			"/titus/systemdns": {},
```
solarkennedy, line 38: whoops, wrong place

executor/runtime/types/sidecars.go
```diff
@@ -33,6 +34,15 @@ var systemServices = []ServiceOpts{
...
+		},
+	},
+	{
+		ServiceName:  SidecarServiceSystemDNS,
+		UnitName:     "titus-sidecar-systemdns",
+		Required:     false,
```
solarkennedy, line 44: This needs to be required :(

executor/runtime/types/sidecars.go
```diff
@@ -153,6 +163,18 @@ func shouldStartTitusSeccompAgent(cfg *config.Config, c Container) bool {
...
 
+func shouldStartSystemDNS(cfg *config.Config, c Container) bool {
+	enabled := cfg.ContainerSystemDNS
+	if !enabled {
+		return false
+	}
+
+	if cfg.SystemDNSServiceImage == "" {
+		return false
+	}
```
solarkennedy, line 174: Hmmmmm....

This needs to only be enabled on the v6 mode.

I don't know if that is available in the `Container` object. If not, you might have to plumb it through.

solarkennedy:
The CI/CD connection to buildkite is broken.
I'm trying to sort that out now.

executor/runtime/types/sidecars.go
acm1: 🤦

executor/runtime/types/sidecars.go
acm1: Noted. What does `Required` mean in this context actually?

solarkennedy: Check out the docstring for the struct ```/ If true, the startup of a task will fail, otherwise will just log```

acm1: Ah got it -- so if the service is enabled but startup fails, the whole task will fail to start, otherwise the task will continue on without this service, IIUC?

executor/runtime/docker/docker.go
```diff
@@ -375,8 +375,14 @@ func (r *DockerRuntime) computeDNSServers() []string {
...
 	case titus.NetworkConfiguration_Ipv6Only.String():
-		// True ipv6 only means we really can only have a v6
-		// resolver
+		// If we're ipv6-only and running the SystemDNS local resolver
+		// which provides DNS64, we'll point to that
+		if systemServices, err := r.c.SystemServices(); err == nil {
+			if runtimeTypes.GetSidecarConfig(systemServices, runtimeTypes.SidecarServiceSystemDNS).EnabledCheck(&r.cfg, r.c) {
+				return []string{"[::1]:5353"}
+			}
+		}
```
acm1, line 384: so this is a little gnarly but ties the logic here to `shouldStartSystemDNS` rather than duplicating

executor/runtime/types/sidecars.go
solarkennedy: correct.

DNS is critical though. I don't want a task to start without it.
I would rather we fail fast and report the error back to the user (and have the failure show up on sidecar dashboards, etc)

solarkennedy (approved):

executor/runtime/docker/docker.go
solarkennedy: I agree, but is fine

solarkennedy: Actually, can you do `runtimeTypes.ShouldStartSystemDNS(r.cfg, r.c)` instead?
You will have to capitlize to export, but I think this is fine

acm1: Ah, was going to do that to start but realized it was unexported in a different package. If there's no reason not to export I'll do it that way 👍

solarkennedy:
ty for rebasing

PR #887, opened 2022-05-31
High Scale by default means IPv6Only

MyUmmaGumma:
TSA is disabled by default. If a job configured with High Scale mode shows up, the effective mode should be IPv6Only.

### DIFF
--- a/executor/runtime/types/types.go
+++ b/executor/runtime/types/types.go
@@ -384,11 +384,7 @@ func computeEffectiveNetworkMode(originalNetworkMode string, assignIPv6Address b
 		return titus.NetworkConfiguration_Ipv4Only.String()
 	}
 	if originalNetworkMode == titus.NetworkConfiguration_HighScale.String() {
-		if !enableTransitionNetwork {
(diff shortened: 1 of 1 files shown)


executor/runtime/types/types.go
```diff
@@ -385,7 +385,7 @@ func computeEffectiveNetworkMode(originalNetworkMode string, assignIPv6Address b
 	}
 	if originalNetworkMode == titus.NetworkConfiguration_HighScale.String() {
 		if !enableTransitionNetwork {
-			return titus.NetworkConfiguration_Ipv6AndIpv4.String()
+			return titus.NetworkConfiguration_Ipv6Only.String()
 		}
 		// HighScale is really an alias for the Transition Mode today
 		return titus.NetworkConfiguration_Ipv6AndIpv4Fallback.String()
```
solarkennedy, line 391: Wrong place.

The corect line to edit is line 391

And the !enableTransitionNetwork is no longer relevant for this HS mode

solarkennedy (changes requested):

solarkennedy (approved):

PR #889, opened 2022-06-01
Relax the docker_fuse sec profile to enable image building

solarkennedy:
Relaxes some of the docker_fuse profiles to enable image building in
containers.

I don't know all of the consequences of allowing these, but FUSE is a small subset of containers, probably fine.

With this change, the buildkit image builder can actually build images using the runc snapshotter.


root/etc/apparmor.d/docker_fuse
```diff
@@ -8,18 +8,18 @@ profile docker_fuse flags=(attach_disconnected,mediate_deleted) {
...
   umount,
   mount,
+  pivot_root,
 
   signal (send,receive) peer=@{profile_name},
 
-  deny mount fstype=tmpfs,
   deny mount fstype=ramfs,
   deny mount fstype=sysfs,
-  deny mount fstype=proc,
```
tych0, line 17: If you allow mounting proc, people can mount their own copy and write to e.g. /myproc/sysrq-trigger, rendering the rules slightly below in this file useless, so you could drop those if you want. If these containers don't run as real root, I don't know that they add much anyway. If they do, this may not be so safe depending on how much you trust the code that code runs inside.

solarkennedy: Hmm, the buildkit snappshotter really did seem to want to mount proc.

Is it really the case that `/myproc/sysrq-trigger` would work? I can try it.

solarkennedy: These are not `--privileged` in that sense, and these are not real root (they are in a user namespace).

sysrq-trigger will not work in this case, even with CAP_SYSADMIN and this PR:

```
titus_us_east_1_agents_test titusagent-devkyleacell001-r5.2xlarge.000-v002848 us-east-1 i-0ac2f9c4f7ff1032d
(nfsuper) ~ $ docker exec -it a30beaa07d38 bash

test job-test_fuse_enables_buildkit-v000 us-east-1 81e88a89-3082-4989-9a95-13d5003fa470
(root) / # echo t >/proc/sysrq-trigger 
bash: /proc/sysrq-trigger: Read-only file system

test job-test_fuse_enables_buildkit-v000 us-east-1 81e88a89-3082-4989-9a95-13d5003fa470
(root) / # mkdir myproc

test job-test_fuse_enables_buildkit-v000 us-east-1 81e88a89-3082-4989-9a95-13d5003fa470
(root) / # mount -t proc proc /myproc/

test job-test_fuse_enables_buildkit-v000 us-east-1 81e88a89-3082-4989-9a95-13d5003fa470
(root) / # echo t >/myproc/sysrq-trigger 
bash: /myproc/sysrq-trigger: Permission denied
```

tych0: Cool, if they're not real root then it seems fine. Might still be worth deleting the rules, just to avoid any confusion in the future.

solarkennedy: You are right that all the rules in this profile that have PROC will not apply to a manually mounted proc (leaving other layers of protection), but I don't think we should delete them. Leaving the standard /proc protected is still valuable.

tych0: Sure, sounds good.

tych0 (approved):

sdab (approved):

PR #894, opened 2022-06-10
seccomp: unconditionally allow the clone3 syscall

tych0:
tl;dr: we allow all containers to call clone3; this is safe because we use
       user namespaces, but is probably safe even if we don't

Timeline:

1. docker is created with a default profile of return EPERM for blocked
   syscalls, allows some combination of clone(2)'s mask, because things are
   privileged and allowing everything is maybe dangerous
2. clone3(2) syscall is created, uses fancy new extensible argument
   syscalls, which seccomp cannot inspect because of TOCTOU vulns
3. docker allows clone3(2) only for people who have CAP_SYS_ADMIN, since
   they are concerned about things they can no longer inspect and can't
   allow it in the general case
4. glibc starts to use clone3(2), but is smart enough to fall back on
   kernels which don't have it (by virtue of ENOSYS)

The problem here is that docker's default policy returns EPERM instead of
ENOSYS for calls it chooses not to allow. there is some resistance
https://github.com/moby/moby/pull/42681#issuecomment-916325271 from docker
upstream to unconditionally allowing clone3(2), presumably because docker's
default is to not use user namespaces, and there are some (are there? i'm
unaware of them, but ready to be enlightened...) container escapes via
clone3(2).

In that thread there is also some discussion about how to change
docker/runc's default seccomp response to ENOSYS, because the current
behavior is extremely unfortunate. Doesn't seem like that's going anywhere,
though.

Let's just allow clone3 unconditionally for all containers. I am unaware of
any reason to disallow it even in the privileged case, so it seems
especially safe if we allow it in our unprivileged case.

This patch also includes an updating of the stock docker profile to include
some new syscalls (landlock, sidechannel protected memory, some more futex
stuff, the rest of the new mount API).


solarkennedy (approved):

solarkennedy:
If you have verified it actually works on a real stack, with the versions of docker/seccomp/whatever we actually have, then shipit!

tych0:
Yep, works on my dev stack at least. So it must work for everyone!

PR #897, opened 2022-06-16
VPC Service Refactor

hechaoli:
# Summary
This PR contains the following changes:
* Remove statsd-related code from VPC service and VPC tool since it never works due to bugs.
  * In VPC service, the bug is that `STATSD_ADDR` environment variable is never set.
  * In VPC tool, the bug is that the variable bound to the environment variable is wrong. Should be `statsdAddrFlagName` but is `zipkinURLFlagName`.
* Move the function of exporting DB metrics from `migrate.go` to `metrics.go`. It is confusing to put metrics collecting code in `migrate.go`, which was used for migrating DB schema.
# Test
```
$ make test-local 
```

### DIFF
--- a/cmd/titus-vpc-service/main.go
+++ b/cmd/titus-vpc-service/main.go
@@ -42,7 +42,6 @@ import (
 
 const (
 	atlasAddrFlagName             = "atlas-addr"
-	statsdAddrFlagName            = "statsd-addr"
 	zipkinURLFlagName             = "zipkin"
 	debugAddressFlagName          = "debug-address"
 	gcTimeoutFlagName             = "gc-timeout"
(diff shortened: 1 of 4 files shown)


MyUmmaGumma (approved):

solarkennedy:
Hmm, did my CODEOWNERS thing not work? It should have added more reviewers.

hechaoli:
> Hmm, did my CODEOWNERS thing not work? It should have added more reviewers.

As per [this doc](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners), seems that we should use "/vpc" rather than "vpc/*" in order to include all subdirectories. I'm trying to update it in [this PR](https://github.com/Netflix/titus-executor/pull/902). Hopefully that will work.

PR #898, opened 2022-06-16
Add a counter for scan ENI error

hechaoli:
# Summary

Add a counter for scan ENI errors.

# Test

No local test. In theory this should work as it works for other stats like `db.waitCount`. Will query Atlas after the code is deployed.

### DIFF
--- a/vpc/service/detach_unused_branch_eni.go
+++ b/vpc/service/detach_unused_branch_eni.go
@@ -9,8 +9,11 @@ import (
 
 	"github.com/Netflix/titus-executor/logger"
 	"github.com/Netflix/titus-executor/vpc/service/ec2wrapper"
+	"github.com/Netflix/titus-executor/vpc/service/metrics"
 	"github.com/Netflix/titus-executor/vpc/tracehelpers"
 	"github.com/pkg/errors"
+	"go.opencensus.io/stats"
+	"go.opencensus.io/tag"
 	"go.opencensus.io/trace"
 )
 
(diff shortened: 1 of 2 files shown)


vpc/service/metrics/metrics.go
```diff
@@ -17,10 +17,11 @@ var (
...
 	// Counter measures
-	waitCount         = stats.Int64("db.waitCount", "The total number of connections waited for", "connections")
-	waitDuration      = stats.Int64("db.waitDuration", "The total time blocked waiting for a new connection", "ns")
-	maxIdleClosed     = stats.Int64("db.maxIdleClosed", "The total number of connections closed due to SetMaxIdleConns", "connections")
-	maxLifetimeClosed = stats.Int64("db.maxLifetimeClosed", "The total number of connections closed due to SetConnMaxLifetime", "connections")
+	waitCount               = stats.Int64("db.waitCount", "The total number of connections waited for", "connections")
+	waitDuration            = stats.Int64("db.waitDuration", "The total time blocked waiting for a new connection", "ns")
+	maxIdleClosed           = stats.Int64("db.maxIdleClosed", "The total number of connections closed due to SetMaxIdleConns", "connections")
+	maxLifetimeClosed       = stats.Int64("db.maxLifetimeClosed", "The total number of connections closed due to SetConnMaxLifetime", "connections")
+	ErrorScanBranchEniCount = stats.Int64("error.scanBranchEni", "The total number of connections closed due to SetConnMaxLifetime", stats.UnitNone)
```
MyUmmaGumma, line 24: "ENI's that could not be cleaned up due to error in query"?

vpc/service/detach_unused_branch_eni.go
```diff
@@ -105,6 +107,7 @@ LIMIT 1
 	}
 	if err != nil {
 		err = errors.Wrap(err, "Cannot scan branch ENI to delete")
+		stats.Record(ctx, metrics.ErrorScanBranchEniCount.M(1))
```
MyUmmaGumma, line 110: Record with tags and add error string/type?

vpc/service/metrics/metrics.go
hechaoli: Oops copy-pasta.

MyUmmaGumma (approved):

PR #900, opened 2022-06-17
Add metrics for table count

hechaoli:
# Summary

This change adds a go routine to Titus VPC service that periodically collect and record certain DB table metrics. This change mainly creates a framework for collecting and recording metrics. Other metrics can be added easily after this change. 

# Metrics
For now, metrics added by this diff are:
* Number of rows in `subnets` table
* Number of rows in `assignments` table
* Number of rows in `branch_enis` table
* Number of rows in `branch_eni_attachments` table
* Number of unattached ENIs, i.e. ENIs with a row in `branch_enis` table but no corresponding row in `branch_eni_attachments` table with state "attached"

# Killswitch &Config

The collect interval is controlled by a flag. 0 means disable collecting table metrics. The default value is 5 minutes.

# Test
A unit test is added. The test starts a Postgres DB in a docker container and use this DB as a test DB. This method might be useful to unit-test other functionalities.

```
$ go test ./vpc/service/metrics/
ok      github.com/Netflix/titus-executor/vpc/service/metrics   9.797s 
```


vpc/service/db/test/test_util.go
```diff
@@ -0,0 +1,230 @@
...
+	case <-time.After(time.Minute):
+		err = container.Shutdown(ctx)
+		if err != nil {
+			logger.G(ctx).Errorf("failed to clean up container: %s", err)
+		}
+		return nil, errors.New("container health check failed")
+	}
+}
+
+// Connect to the test DB and set up table schema using the DB migrate tool
```
MyUmmaGumma, line 114: Was the intent to call `err = db.MigrateTo(ctx, testDb, 40, false)` here? (according to the comment)

hechaoli: Oh I tried putting it there in the first place. Somehow `make test-local` works on my machine but failed remotely when I do that. Updated the comment.

PR #901, opened 2022-06-20
Remove test.db file

hechaoli:
# Summary

Accidentally checked in a test db file which is not needed for testing. Remove it.


solarkennedy:
Can you add a .gitignore entry if this is generated?

hechaoli:
This is not generated. I tried using this file first but later abandoned the method. Forgot to remove the file.

PR #902, opened 2022-06-21
Add explicit env var bind for dburl flag

hechaoli:
# Summary

The way Titus VPC Service get the DB URL is through case-insensitive viper config "dburl" plus `AutomaticEnv`, which automatically get the value from environment variable `DBURL`. This magic is a bit confusing. This commit adds an explicit environment variable bind for "dburl" config to avoid confusion.

Also 
* Fixed a typo I made earlier.
* Updated CODEOWNERS file to support files in subdirectories.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,11 +1,11 @@
 * @solarkennedy
 * @tych0
 
-vpc/* @aloktiagi
-vpc/* @DonavanFritz
-vpc/* @MyUmmaGumma
-vpc/* @hechaoli
-vpc/* @aloktiagi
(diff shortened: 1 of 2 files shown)


CODEOWNERS
```diff
@@ -1,11 +1,11 @@
 * @solarkennedy
 * @tych0
 
-vpc/* @aloktiagi
-vpc/* @DonavanFritz
-vpc/* @MyUmmaGumma
-vpc/* @hechaoli
-vpc/* @aloktiagi
+/vpc/ @aloktiagi
```
solarkennedy, line 4: ty

PR #905, opened 2022-06-23
Refactor VPC Service

hechaoli:
# Summary

I’ve found the following issues when reading VPC service code:
* When starting VPC service, some operations are done in `main.go`  and some are done in `Run` function. It's unclear what operations are needed only for running VPC service and what are needed only by the command itself.
* `vpcService` struct members is a mix of configs and data structures. Some of the configs are duplicates of the `Config` struct.

This refactoring tries to achieve the following to make the code clearer:
* Separate configs and members for `VpcService` struct. Make sure all configs are in `Config` struct. Configs are what we passed at start-up time (e.g. DB URL). Members are the dependencies of `VpcService` (e.g. a `*sql.DB` struct). `VpcService` also depends on `Config` as one member to get configs at runtime.
* `cmd/titus-vpc-service/main.go` mainly does the following things. It knows nothing about operations needed in order to run VPC service. For example, it doesn't know how to connect to the DB needed for VPC service. That is done by `Run` function.
  1) Get the configs from user input (flags, env vars, etc).
  2) Construct the `Config` struct for `VpcService`. 
  3) Construct a `VpcService` struct using the config.
  4) Start running the `VpcService` by calling `Run` function. 
* `VpcService.Run` function is responsible for performing all the actions needed for running VPC service. For example, connecting to DB, setting up monitoring, starting grpc server, etc. It takes config as an input and knows nothing about where the configs are from.
* Put all config names into one file `vpc/service/config/constants.go`.

I also removed some configs that are not used anywhere to reduce confusion, including
* RefreshInterval
* GcTimeout
* subnetCacheExpirationTime

# Test
```
make test-local
```

### DIFF
--- a/vpc/service/service_test.go
+++ b/vpc/service/service_test.go
@@ -24,6 +24,8 @@ func TestService(t *testing.T) {
 	defer cancel()
 	listener, err := net.Listen("tcp", "localhost:0")
 	assert.NilError(t, err)
+	addr := listener.Addr().String()
+	listener.Close()
 	_, privatekey, err := ed25519.GenerateKey(nil)
 	assert.NilError(t, err)
 
@@ -38,14 +40,11 @@ func TestService(t *testing.T) {
 	}
 
 	go func() {
-		serverErr := Run(ctx, &Config{
-			Listener:              listener,
-			DB:                    nil,
-			Key:                   key, // nolint: govet
+		vpcServiceConfig := &Config{
+			DBURL:                 "",
+			Key:                   key, // nolint:govet
 			MaxConcurrentRefresh:  10,
-			GCTimeout:             2 * time.Minute,
 			ReconcileInterval:     5 * time.Minute,
-			RefreshInterval:       30 * time.Second,
 			TLSConfig:             nil,
 			EnabledTaskLoops:      []string{},
 			EnabledLongLivedTasks: []string{},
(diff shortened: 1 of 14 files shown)


vpc/service/service_test.go
```diff
@@ -55,16 +54,22 @@ func TestService(t *testing.T) {
...
+		vpcService, err := NewVpcService(ctx, vpcServiceConfig)
+		if err != nil {
+			panic(err)
+		}
+		serverErr := vpcService.Run(ctx, addr)
 		if serverErr != nil && serverErr != context.Canceled {
 			panic(serverErr)
 		}
 	}()
+	time.Sleep(time.Second * 2)
```
solarkennedy, line 67: why do we need a sleep here?

vpc/service/branch_enis_test.go
```diff
@@ -63,7 +63,7 @@ func TestListBranchToTrunkENIMapping(t *testing.T) {
 	mock.ExpectQuery("SELECT branch_eni, trunk_eni FROM branch_eni_attachments WHERE state = 'attached'").WillReturnRows(rows)
 	mock.ExpectCommit()
 
-	service := vpcService{db: db}
+	service := VpcService{db: db}
```
solarkennedy, line 66: Does `VpcService` have to be exported? (capitalized)
When you export stuff from other packages, you lose the ability for go linters to let you know if you have dead code in certain situations

solarkennedy (approved):
This change is a little to big to really review, as it is in 1 big commit.

In the future, please break up the changes into logical commits that we can review one at a time, to give it the attention it needs.

Otherwise, all I can say is "shipit"

solarkennedy:
Please get this reviewed by someone on the HN team

vpc/service/branch_enis_test.go
hechaoli: I changed it so that any users of `VpcService` needs to construct the struct first and then use the `Run` method to start it. I think it needs to be exported for `main.go` (in `cmd` package) to use it. Do you know if there is a way in Golang to keep it private but also visible to `cmd/titus-vpc-service`?

vpc/service/service_test.go
hechaoli: In the past, the caller of `Run` function starts listening and then pass the listener to `Run`. With this refactoring, the listening is started inside `Run` and the caller only passes an address. In this particular test, before the refactoring, listening is done in the main thread. After the refactoring, listening is done in the go routine where `Run` is called. So it may take a while before the health check starts. Actually, even in the old code, if the whole RPC server setup is done in a separate go routine, then it may take a while before the health check in the main thread to pass. I guess moving listening to `Run` just made this time difference larger.

Sleep is not ideal. Ideally, we should have a retry with timeout. I'll do that in later cleanup.

vpc/service/branch_enis_test.go
solarkennedy: The only way I know how to do it is to make the cmd main.go be as minimal as possible, and then put everything in vpc/service folder

hechaoli: Alright. Changed back to `vpcService` for better lints. Well, I have to say, coming from C++ and Rust world, Go's visibility controlled by naming thing bothers me a little.

MyUmmaGumma (approved):

PR #908, opened 2022-06-23
Prepended "Error while pulling Docker image" to image pull failures

solarkennedy:
We depend on this error so that we treat them as "localsystemerrors" (by
the control plane).

Also it is just lame to see docker's "500" error unadorned.

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -633,7 +633,7 @@ func (r *DockerRuntime) DockerPull(ctx context.Context, c runtimeTypes.Container
 				return nil, &runtimeTypes.RegistryImageNotFoundError{Reason: err}
 			}
 			tracehelpers.SetStatus(err, span)
-			return nil, err
+			return nil, fmt.Errorf("Error checking if image %s exists: %w", imgName, err)
 		}
 
 		if imgInfo != nil {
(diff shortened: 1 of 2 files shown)


executor/runtime/docker/docker.go
```diff
@@ -633,7 +633,7 @@ func (r *DockerRuntime) DockerPull(ctx context.Context, c runtimeTypes.Container
 				return nil, &runtimeTypes.RegistryImageNotFoundError{Reason: err}
 			}
 			tracehelpers.SetStatus(err, span)
-			return nil, err
+			return nil, fmt.Errorf("Error checing if image %s exists: %w", imgName, err)
```
aaronlehmann, line 636: Typo: `checing`

aaronlehmann (commented):
> We depend on this error so that we treat them as "localsystemerrors" (by the control plane).

Do you mean that the control plane is checking for a substring inside the error? Is there a cleaner way to do this, perhaps using error codes or a structured error?

solarkennedy:
> > We depend on this error so that we treat them as "localsystemerrors" (by the control plane).
> 
> Do you mean that the control plane is checking for a substring inside the error? Is there a cleaner way to do this, perhaps using error codes or a structured error?

There is no such thing dude, this is k8s. You get strings.
We could stick json in there? :)
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#podstatus-v1-core

aaronlehmann:
Eww :). Well, maybe a prefix like `PULLERR` would make it clear it's something intended to be machine-parsed? Just worried it's a little fragile as-is; changing the error message could break stuff.

solarkennedy:
> Eww :). Well, maybe a prefix like `PULLERR` would make it clear it's something intended to be machine-parsed? Just worried it's a little fragile as-is; changing the error message could break stuff.

That is a great idea (given the situation).
I'll make constants for each of the systemerror regexs we have.

amit-git (approved):

PR #910, opened 2022-06-24
Consolidate localSystemError strings into a single place

solarkennedy:
This makes it more obvious that these strings are special and cannot be
changed willy-nilly

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -86,6 +86,14 @@ const (
 	nonTerminalDockerEvent  = false
 )
 
+const (
+	// These Special strings are used by the Titus Control Plane as special "retryable errors" (localSystemError)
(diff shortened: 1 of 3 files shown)


executor/runtime/docker/docker.go
```diff
@@ -921,7 +929,7 @@ func (r *DockerRuntime) Prepare(ctx context.Context, pod *v1.Pod) (err error) {
 			}
 			tracehelpers.SetStatus(netErr, span)
 			if netErr != nil {
-				return fmt.Errorf("network setup error: %w", netErr)
+				return fmt.Errorf("%s: %w", networkSetupErrorString, netErr)
```
aaronlehmann, line 932: How about a function `retryableError` which we can call like this?

```
return retryableError(vpcErrorString, "%w", netErr)
```

PR #911, opened 2022-06-24
Add support for dynamic configs

hechaoli:
# Summary

Since we are going to have more changes to VPC service coming, it's safer to guard any new changes with killswitches, which can be enabled/disabled at runtime. 

This change adds a framework to add dynamic configs. Periodically, the config values are fetched from a provided dynamic configs URL. For a config, if there is no value is fetched, then the default value is used.

This change doesn't add any dynamic config yet. They'll be added as we add new features.

# Test
Added a new UT
```
$ go test -v -test.run='TestDynamicConfig' ./vpc/service/
=== RUN   TestDynamicConfig
time=2022-06-24T09:13:09-07:00 level=info msg=Dynamic configs updated to {"TEST_CONFIG":"24"}
--- PASS: TestDynamicConfig (0.01s)
PASS
ok      github.com/Netflix/titus-executor/vpc/service
```

### DIFF
--- a/vpc/service/config.go
+++ b/vpc/service/config.go
@@ -4,9 +4,14 @@ import (
 	"context"
 	"crypto/tls"
 	"crypto/x509"
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"net/http"
 	"os"
+	"reflect"
+	"strings"
+	"sync"
 	"time"
 
 	"github.com/Netflix/titus-executor/logger"
@@ -15,6 +20,7 @@ import (
 	"github.com/Netflix/titus-executor/vpc/service/config"
 	"github.com/golang/protobuf/jsonpb" // nolint: staticcheck
 	"github.com/pkg/errors"
+	"github.com/spf13/cast"
 	"github.com/spf13/viper"
 )
 
@@ -44,6 +50,9 @@ type Config struct {
 
 	// This is only used for testing.
 	disableRouteCache bool
+
+	// The URL from where to fetch dynamic configs
+	DynamicConfigURL string
 }
 
 func NewConfig(ctx context.Context, v *viper.Viper) (*Config, error) {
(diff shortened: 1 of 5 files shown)


vpc/service/config.go
```diff
@@ -86,9 +95,86 @@ func NewConfig(ctx context.Context, v *viper.Viper) (*Config, error) {
 		WorkerRole: v.GetString(config.WorkerRoleFlagName),
 
 		ZipkinURL: v.GetString(config.ZipkinURLFlagName),
+
+		DynamicConfigURL: v.GetString(config.DynamicConfigURLFlagName),
 	}, nil
 }
 
+// Dynamic configs have a default value, which can be overriden at runtime.
```
solarkennedy, line 103: I don't think we use this code actually,
But there is

https://github.com/Netflix/titus-executor/tree/master/properties

And then there is also this?

https://github.com/Netflix/quitelite-client-go/tree/master/properties

Is one or both of those useful to you?

solarkennedy (approved):
This needs to be reviewed by someone on the HN team.

In general someone on the HN team should be your primary reviewer

vpc/service/config.go
hechaoli: I looked at this library. I think you're right, it doesn't seem to be used anywhere. And looks like it was designed to use for getting configs for CLI application built with`cli/altsrc` (e.g. `fetchQuiteLiteSource` returns a `altsrc.InputSourceContext`). I suppose we could modify it to make it work with a more general use case and use it in VPC service. But I personally feel it's easier for us to add some minimal code that just works for VPC service for now.

vpc/service/service.go
```diff
@@ -330,7 +331,19 @@ func (vpcService *vpcService) run(ctx context.Context, address string) error {
 		logger.G(ctx).WithField("endpoint", endpoint).WithField("url", vpcService.config.ZipkinURL).Info("Setting up tracing")
 		trace.RegisterExporter(zipkin.NewExporter(reporter, endpoint))
 	}
+	return nil
+}
 
+func (vpcService *vpcService) run(ctx context.Context, address string) error {
+	if vpcService.config.DynamicConfigURL != "" {
+		vpcService.dynamicConfig = NewDynamicConfig()
+		vpcService.dynamicConfig.Start(ctx, 5*time.Minute, vpcService.config.DynamicConfigURL)
```
MyUmmaGumma, line 340: Move this as a constant to the top of the file

MyUmmaGumma (approved):

PR #912, opened 2022-06-25
Fix VPC service metrics test

hechaoli:
# Background

In VPC metrics test, a psql docker container is started to be used as a test DB. The container port 5432 (i.e. psql port) is bound to a random available host port for the test to connect.

# Problem

Occasionally, we see a test failure in CI with the error: 
```
failed to connect to test DB: Could not select current database: read tcp 127.0.0.1:60806->127.0.0.1:43871: read: connection reset by peer
```

But I can't reproduce the problem locally.

# Fix
The symptom seems similar to [this `lib/pq` issue](https://github.com/lib/pq/issues/835). Though the issue is still open, several PRs referencing the issue mentioned that upgrading `lib/pq` library could fix it. This change upgrades `lib/pq` library from `1.3.0` to `1.10.0`. Also added a retry just in case.

# Test
Can't reproduce this locally. So retried the CI build 10+ times and they are all green: https://buildkite.com/netflix/titus-executor/builds?branch=fix_metrics_test

### DIFF
--- a/vpc/service/db/test/test_util.go
+++ b/vpc/service/db/test/test_util.go
@@ -7,11 +7,14 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"log"
 	"math/big"
 	"net"
 	"time"
 
 	"github.com/Netflix/titus-executor/logger"
+	"github.com/Netflix/titus-executor/vpc/service/db"
+	"github.com/Netflix/titus-executor/vpc/service/db/wrapper"
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/container"
 	"github.com/docker/docker/client"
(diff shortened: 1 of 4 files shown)


vpc/service/db/test/test_util.go
```diff
@@ -114,7 +117,26 @@ func StartPostgresContainer(ctx context.Context) (*PostgresContainer, error) {
 // Connect to the test DB
 func (c *PostgresContainer) Connect(ctx context.Context) (*sql.DB, error) {
 	url := fmt.Sprintf("postgres://pgtest:%s@localhost:%s/pgtest?sslmode=disable", c.password, c.port)
-	return sql.Open("postgres", url)
+
+	var err error
+	// Connect to DB and set up tables. Retry a couple of times in case of failure.
+	for i := 0; i < 10; i++ {
```
tych0, line 123: Is it worth merging the version bump without this retry loop since the bug is supposedly fixed in the library?

hechaoli: I don't think it was fixed completely. After bumping the version, the rate of failure decreased but I still saw it once in 5 runs. That issue in `lib/pq` is still open. Someone asked if it was fixed last year but no reply. I asked again. Before we have a definite answer, it's safer to just retry on our side.

tych0: Ok, thanks.

MyUmmaGumma (approved):

solarkennedy (approved):

hechaoli:
Unfortunately, I've seen the same error even after 10 retries. Given that the `lib/pq` issue is still open, I think the problem is not solved fully. It could also have something to do with the CI environment since I can't reproduce locally and we've never seen it in production. For now, I'll just skip the tests if failing to connect to DB and will debug the problem later.

PR #913, opened 2022-06-27
runner: track tasks stuck in uninterruptible sleep

tych0:
Log something about tasks that have been stuck for too long. Maybe we
should surface something to users about this, but for starters let's just
surface something to ourselves.

Note that currently on our agents one needs to set
kernel.hung_task_warnings to -1 in order to get unlimited warnings (or pair
this with my commit to titus-agent to add the permanent sysctl).

I've run this on my devstack faking kmsg logs and it works, but I haven't
written an integration test for it yet since it requires ssh-ing to the
host, which I don't see any code for doing in the integration tests. Maybe
soon...

### DIFF
--- a/executor/runner/kernel_logs_linux.go
+++ b/executor/runner/kernel_logs_linux.go
@@ -1,0 +1,84 @@
+//go:build linux
+// +build linux
+
+package runner
+
+import (
+	"fmt"
+	"os"
+	"regexp"
+	"strconv"
+
+	"github.com/opencontainers/runc/libcontainer/cgroups"
+	"github.com/sirupsen/logrus"
+)
+
+var errNotTitusTask = fmt.Errorf("Not a titus task")
+
+var cgroupTaskIDRegex = regexp.MustCompile("titus-executor@default__([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}).service")
+
+func findTitusTaskIDFromPid(pid int) (string, error) {
+	cgroups, err := cgroups.ParseCgroupFile(fmt.Sprintf("/proc/%d/cgroup", pid))
+	if err != nil {
+		return "", err
+	}
+
+	// pick a representative cgroup, say cpuset
+	root, ok := cgroups["cpuset"]
+	if !ok {
+		// "" is unified, which is all that's present in full v2
+		root = cgroups[""]
+	}
+
+	matches := cgroupTaskIDRegex.FindStringSubmatch(root)
+	if len(matches) != 2 {
+		return "", errNotTitusTask
+	}
+
+	return matches[1], nil
+}
+
+var stuckTaskKernelLogLineRegex = regexp.MustCompile(`INFO: task ([\w-]+):(\d+) blocked for more than \d+ seconds`)
+
+func parseBlockedTaskKernelLogLine(log logrus.FieldLogger, line string, myTaskID string) {
+	matches := stuckTaskKernelLogLineRegex.FindStringSubmatch(line)
+	if len(matches) == 0 {
+		return
+	}
+
+	if len(matches) != 3 {
+		log.Errorf("didn't understand regex match %v for %v", matches, line)
+		return
+	}
+
+	name := matches[1]
+	pid, err := strconv.Atoi(matches[2])
+	if err != nil {
+		log.Errorf("strange pid value in %s: %s", line, matches[2])
+		return
+	}
+
+	pidsTaskID, err := findTitusTaskIDFromPid(pid)
+	if err != nil {
+		// maybe the task died between when we got this message and
+		// when we tried to access the cgroup file.
+		if os.IsNotExist(err) {
+			return
+		}
+
+		// maybe this wasn't a titus task, and something on the host is locked up? punt for now.
+		if err == errNotTitusTask {
+			return
+		}
+
+		log.Errorf("error getting titus task from pid: %v", err)
+		return
+	}
+
+	// this is somebody else's task
+	if myTaskID != pidsTaskID {
+		return
+	}
+
+	log.Infof("Process %d (%s) in %s hung for the task timeout length", pid, name, myTaskID)
+}
--- a/executor/runner/runner.go
+++ b/executor/runner/runner.go
@@ -1,8 +1,11 @@
 package runner
 
 import (
+	"bufio"
 	"context"
 	"fmt"
+	"io"
+	"os"
 	"sync"
 	"time"
 
@@ -211,6 +214,7 @@ func (r *Runner) runMainContainerAndStartSidecars(ctx context.Context, startTime
 		return
 	}
 	logger.G(ctx).Infof("started %d container(s): %s, now monitoring for container status", len(containerNames), containerNames)
+	r.monitorKernelLogsForStuckTasks(ctx)
 
 	err = r.maybeSetupExternalLogger(ctx, logDir)
 	if err != nil {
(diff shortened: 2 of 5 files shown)


tych0:
Ah, of course. Tests are run in a container which can't talk to systemd...

```Failed to connect to bus: No such file or directory```

Maybe those can just make cgroups directly then?

**Comment by tych0** [comment]
I think the approach I'll take here is to use the same trick with `docker run --cgroup-parent` and run these in the standalone DIND tests, unless there are any objections.

File: executor/runner/kernel_logs.go [hunk 1]
```diff
@@ -0,0 +1,118 @@
...
+	"bufio"
+	"fmt"
+	"os"
+	"regexp"
+	"strconv"
+	"strings"
+)
+
+func readCgroupFile(pid int) ([]string, error) {
+	rawCgroups, err := os.Open(fmt.Sprintf("/proc/%d/cgroup", pid))
```
solarkennedy, line 13: We already have code like this, that parses cgroup files,
can you look at `setCgroupOwnership` to see if you can reuse?

executor/runner/kernel_logs.go
```diff
@@ -0,0 +1,118 @@
...
+				return id, nil
+			}
+		}
+
+		return "", fmt.Errorf("couldn't find cpuset controller for pid %d\n", pid)
+	}
+}
+
+func handleKernelLogLine(line string, task Task) (string, error) {
+	re := regexp.MustCompile(`INFO: task ([\w-]+):(\d+) blocked for more than \d+ seconds`)
```
solarkennedy, line 80: You don't have to compile this regex every log line
Just put it at the top https://stackoverflow.com/questions/25096688/go-global-variables

executor/runner/kernel_logs.go
```diff
@@ -0,0 +1,118 @@
...
+
+var errNotTitusTask = fmt.Errorf("Not a titus task")
+
+func getControllerAndTaskid(line string) (string, string, error) {
+	fields := strings.Split(line, ":")
+	if len(fields) != 3 {
+		return "", "", fmt.Errorf("strange /proc/pid/cgroup line %s\n", line)
+	}
+
+	re := regexp.MustCompile("titus-executor@default__([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}).service")
```
solarkennedy, line 41: same here, global

executor/runner/kernel_logs.go
```diff
@@ -0,0 +1,118 @@
...
+
+func handleKernelLogLine(line string, task Task) (string, error) {
+	re := regexp.MustCompile(`INFO: task ([\w-]+):(\d+) blocked for more than \d+ seconds`)
+	matches := re.FindStringSubmatch(line)
+	if len(matches) == 0 {
+		return "", nil
+	}
+
+	if len(matches) != 3 {
+		return "", fmt.Errorf("didn't understand regex match %v for %v", matches, line)
```
solarkennedy, line 87: Isn't this going to happen all the time?
Tons of kernel messages are not going to match the `INFO: task pid` format in dmesg.
I think this should silently return in that case, as I dare say it is the "normal" situation.

executor/runner/kernel_logs.go
```diff
@@ -0,0 +1,118 @@
...
+			if controller == "cpuset" {
+				return id, nil
+			}
+		}
+
+		return "", fmt.Errorf("couldn't find cpuset controller for pid %d\n", pid)
+	}
+}
+
+func handleKernelLogLine(line string, task Task) (string, error) {
```
solarkennedy, line 79: This function only handles `blocked for more than` log lines.
I think it would be better named `parseBlockedTaskKernelLogLine`, until it handles more than that

executor/runner/runner.go
```diff
@@ -211,6 +214,7 @@ func (r *Runner) runMainContainerAndStartSidecars(ctx context.Context, startTime
 		return
 	}
 	logger.G(ctx).Infof("started %d container(s): %s, now monitoring for container status", len(containerNames), containerNames)
+	r.monitorKernelLogsForStuckTasks(ctx, task)
```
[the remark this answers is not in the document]
tych0: Ah, I looked and couldn't find it. Seems like `r.container.TaskID()` might work, thanks.

executor/runner/runner.go
```diff
@@ -470,3 +474,61 @@ func (r *Runner) getPlatformSidecarStatuses() []corev1.ContainerStatus {
...
+		}
+	}()
+
+	go func() {
+		for {
+			select {
+			case <-r.killChan:
+				return
+			case kernelLogLine, more := <-ch:
+				msg, err := handleKernelLogLine(kernelLogLine, task)
```
solarkennedy, line 519: I think this might read better like

if isStuckTaskLogLine(kernelLogLine) {
  parseStuckKernelLogLine(kernelLogLine)
}
And can give us more room in the future for other types of lines, and makes it more clear what that function is doing exactly. And it doesn't have to worry about so many error cases as much

executor/runner/kernel_logs.go
tych0: Not re-use directly, but I can commonize it. Will do that.

executor/runner/kernel_logs.go
tych0: Yes, it should be handled by the len(matches) == 0 test above this one. I actually don't know that this len(matches) != 3 can happen, the regex API just returns matches and groups as an array of strings, and there are two groups here. The API docs aren't super obvious, though...

tych0:
Ugh, getting closer,

```
+ mkdir -p /sys/fs/cgroup/cpuset/docker/7f4fdc366617785ef73a42e99524419703dc27ecff8b62ba13d35f2b9e8d0a7d/7e938658b685ff284983a640a9f67326639f52f8b90deb2171ec29ef840c812a/titus-executor@default__9c97e413-77bb-4953-8232-2ea03f43d36c.service
mkdir: cannot create directory '/sys/fs/cgroup/cpuset/docker': Permission denied
```

tych0:
Ok, finally! I think this is ready for a re-review.

executor/runner/kernel_logs_linux.go
```diff
@@ -0,0 +1,84 @@
...
+
+	if len(matches) != 3 {
+		log.Errorf("didn't understand regex match %v for %v", matches, line)
+		return
+	}
+
+	name := matches[1]
+	pid, err := strconv.Atoi(matches[2])
+	if err != nil {
+		log.Errorf("didn't understand regex match %v for %v", matches, line)
```
solarkennedy, line 57: wrong error

hack/tests-with-dind.sh
```diff
@@ -69,6 +69,20 @@ if [[ $(uname) == "Darwin" ]]; then
...
 
+docker run $TTYFLAG --privileged \
+  --cgroupns private \
+  -v ${PWD}:${PWD} \
+  -w ${PWD} \
+  --label "$run_id" \
+  --rm --name "$titus_agent_name" \
+  -d titusoss/titus-agent
+
+docker exec $TTYFLAG --privileged -e DIND=true -e DEBUG=${debug} -e SHORT_CIRCUIT_QUITELITE=true -e GOPATH=${GOPATH} "$titus_agent_name" \
```
solarkennedy, line 80: I don't understand, are we not already running these tests, but in a different context? Different github runner.
We probably don't need to run them twice, right?

solarkennedy (approved):

hack/tests-with-dind.sh
tych0: I don't claim to really understand the CI environment these are run in, but it looks like it doesn't have perms to create cgroups by default:

https://app.circleci.com/pipelines/github/Netflix/titus-executor/2135/workflows/bb67c865-5f66-439f-af0a-d2193f655898/jobs/20525#step-102-156

I skip the test in that case, but I added this bit so that they would at least run somewhere.

executor/runner/kernel_logs_linux.go
tych0: whoops, thanks.

PR #914, opened 2022-06-29
VPC Service Local E2E Tests

hechaoli:
# Background
Currently, VPC service has 2 types of tests: 
* Unit tests that mostly don't test prod behaviors.
* Integration tests that don't run in CI. And they require both a real production DB and AWS permissions.

It is better to have some E2E tests that don't require additional permissions. In order for such tests to run locally and on CI, we only need to mock 2 external dependencies: DB and AWS.

# Summary

This PR contains 2 commits:

1) Add dependency injection to `EC2SessionManager`. Specifically, pass two functions `NewSts` and `NewEC2` which create a AWS STS and EC2 client respectively. In prod code, we continue use AWS API to create real clients. In test code, pass functions that return mocked STS and EC2 clients.
2) Add a local E2E test for `ProvisionInstanceV3` API. The test uses a local DB inside a docker container and mocked AWS services. 

# Test
```
$ go test -covermode=count ./vpc/service/
ok      github.com/Netflix/titus-executor/vpc/service   10.318s coverage: 6.6% of statements
```

# Note

* The two mocked AWS API files `mock_aws_ec2.go` and `mock_aws_sts.go` are auto-generated by `mockgen`. No need to review. 
* This diff only adds a most basic test for `ProvisionInstanceV3` API. It is also for me to better understand the service. With the local E2E test framework added by this this, other test cases shouldn't be too hard to add.


vpc/service/service_test.go
```diff
@@ -60,7 +60,24 @@ func TestService(t *testing.T) {
 			panic(err)
 		}
 	}()
-	time.Sleep(time.Second * 2)
+	done := make(chan bool)
```
solarkennedy, line 63: I think this would make a nice function called `waitForPortToBeOpen(port)`

vpc/service/e2e_test.go
```diff
@@ -0,0 +1,354 @@
...
+	}
+	vpcService.ec2.NewEC2 = func(p client.ConfigProvider, cfgs ...*aws.Config) ec2iface.EC2API {
+		return mockEC2
+	}
+
+	err = vpcService.run(ctx, addr)
+	require.NoError(t, err)
+}
+
+func TestProvisionInstanceV3(t *testing.T) {
```
solarkennedy, line 155: Confirmed this test really did run under buildkite and took 52 seconds.
Great job!

vpc/service/e2e_test.go
```diff
@@ -0,0 +1,354 @@
+package service
```
solarkennedy, line 1: If this test gets too big, it might be worth splitting out like we do with the `standalone` tests and out of test-local.

I think this would be good because the tests only "happen" to run on buildkite under the standalone test, because docker happens to be present. Otherwise they are silently skipped.

I think having a dedicated test target can help separate failures (makes it obvious if the failure is from vpc e2e vs standalone stuff) and makes it so we don't accidentally skip the vpc e2e tests from being run in the future via other refactors.

solarkennedy (approved):

vpc/service/e2e_test.go
hechaoli: That's a good idea. Let me figure out how to do it. The test will surely become big as I add more tests.

vpc/service/e2e_test.go
hechaoli: It's a bit strange that both this test and `TableMetricsTest` took 50+ seconds. I thought the most time consuming step is to pull docker image. And both tests pull the same `postgres` docker image. I expected to see the test that runs later takes shorter time. Anyway, I will try figuring it out after I separate these tests to its own target.

MyUmmaGumma: +1 to this!

MyUmmaGumma (approved):

PR #916, opened 2022-07-07
Vpc service db queries

hechaoli:
# Summary

The PR contains the following changes:

* Add a `data` package for structures used to represent DB data model.
* Add a `db/query.go` file for all DB queries. 
    * Functions in this file should run a DB query and return a data model from data package
* Move error-related files to `vpcerrors` package
  * This is needed because `db` package now depends on `errors` so it can't be in `service` package which depends on `db` package.
* Migrate all DB queries in `assign_addresses_v3` to `query.go`

The goal is to achieve:

* Establish “DB <-> Data Model <-> VPC Service” abstraction layers. 
    * Currently, there is no clear data model. The data and code are tightly coupled, making the logic a bit hard to understand. 
    * Ideally, the service code should only deal with data model and should know nothing about the actual DB schema.
* Easy to unit-test DB queries
    * With each query in one function with a meaningful name, it’s easier to unit-test them. Though this PR doesn’t add any. 
* Put all DB queries in one place so that it’s easier to add metrics such as execution time for them.
    * Will add metrics for some important queries in a later change.

NOTE:

This PR simply moves (i.e. copy-paste) queries to `query.go`. There are more refactoring opportunities to make the data model clearer. I’ll do it in future changes.


vpc/service/assign_addresses_v3.go
```diff
@@ -194,7 +169,7 @@ WHERE assignment_id = $1
 	if !completed {
 		if deleteUnfinishedAssignment {
 			// Delete the assignment
-			_, err = tx.ExecContext(ctx, "DELETE FROM assignments WHERE assignment_id = $1", assignmentID)
+			_, err = tx.ExecContext(ctx, "DELETE FROM assignments WHERE assignment_id = $1", taskID)
```
solarkennedy, line 172: Is this right here? Is assignment_id really a taskId in this context?

vpc/service/assign_addresses_v3.go
```diff
@@ -618,26 +601,26 @@ func (vpcService *vpcService) assignIPsToENI(ctx context.Context, req *vpcapi.As
 		}
 	case *vpcapi.AssignIPRequestV3_TransitionRequested:
 		resp.TransitionAssignment = &vpcapi.AssignIPResponseV3_TransitionAssignment{
-			AssignmentId: transitionAss.assignmentID,
+			AssignmentId: transitionAss.TaskID,
```
solarkennedy, line 602: I'm really worried that assignment id and taskID are getting mixed here

vpc/service/assign_addresses_v3.go
```diff
@@ -803,33 +803,19 @@ func assignElasticAddressesBasedOnGroupName(ctx context.Context, tx *sql.Tx, ses
...
-		err = fmt.Errorf("No EIP in group %s free in network border group %s", groupName, borderGroup)
-		span.SetStatus(traceStatusFromError(err))
-		return nil, err
-	} else if err != nil {
-		err = errors.Wrap(err, "Cannot query for free elastic IPs")
+	elasticAddress, err := db.GetAvailableElasticAddressByGroupAndLock(ctx, tx, aws.StringValue(branchENI.OwnerId), borderGroup, groupName)
+	if err != nil {
+		if err == sql.ErrNoRows {
+			err = fmt.Errorf("no EIP in group %s free in network border group %s", groupName, borderGroup)
+		} else if err != nil {
```
solarkennedy, line 810: This error handling here is a bit of a departure from the original code.
At this line, err is already != nil, we'll always hit it.
I guess it is not technically incorrect.

vpc/service/assign_addresses_v3.go
```diff
@@ -1016,7 +1016,7 @@ func assignArbitraryIPv4AddressV3(ctx context.Context, tx *sql.Tx, branchENI *ec
 	}, nil
 }
 
-func (vpcService *vpcService) unassignStaticAddress(ctx context.Context, assignmentID string) (bool, error) {
+func (vpcService *vpcService) unassignStaticAddress(ctx context.Context, taskID string) (bool, error) {
```
solarkennedy, line 1019: This is another assignmentID -> taskID

solarkennedy (approved):
Thanks for doing individual commits to make this at least possible to review.

My biggest concern is around the assignmentid->taskid changes.
I realize that is "just" a variable name change, but if it isn't actually true that there is a 1:1 correspondence, then that change will make things more confusing and introduce more bugs down the line.

My other concern is around transactions.
Previously the `tx` object was just a local var, so it wasn't too crazy to lose track of, but it did result in larger functions.
With the refactor, we just need to be super careful to pass `tx` correctly so that multiple query functions will be rolled back correctly and stay consistent.


In theory this refactor doesn't have any behavior changes, but it is too big for me to be able to actually say that confidently.

vpc/service/assign_addresses_v3.go
hechaoli: There are two "ids" in an assignment table:
* id: The PK, which is auto-generated.
* assignment_id: Use the task ID passed by caller (titus-vpc-tool/titus-agent) in the API calls such as `AssignIPV3`.

In the code, some places we have "assignment.id" (referring to `id`) and some places we have `assignment_id`, which are kind of confusing. I'd like to refer to the second using `taskId` to avoid confusing.

Eventually, this query will also be moved to its own function like `DeleteAssignmentByTaskID`. Because the caller knows nothing about the two IDs, what I want to achieve is that the service code also knows nothing about it. Eventually, only DB-related code should know that when deleting an assignment for a given task ID, it should use the task ID as the assignment ID in the DB query.

vpc/service/assign_addresses_v3.go
hechaoli: You're right. It is indeed a bit special when it comes to transition assignment. The assignment ID is not exactly the task ID but a generated ID in the form of "t-<UUID>", calling it `taskID` is confusing. Will revert this.

vpc/service/assign_addresses_v3.go
hechaoli: I moved `err == sql.ErrNoRows` check to under `err != nil` check so that it's clear that when `err != nil`, we always return from here. You're right about the redundant `else if`. Will change it to `else` only.

vpc/service/assign_addresses_v3.go
hechaoli: To answer the question, yes in this context, `assignment_id` is really the task ID. If you trace the callers, you'll find they all pass `req.TaskId` as the `assignment_id`. I'll make sure things are clearer after some other refactors to separate code and data.

hechaoli:
> Previously the tx object was just a local var, so it wasn't too crazy to lose track of, but it did result in larger functions.
With the refactor, we just need to be super careful to pass tx correctly so that multiple query functions will be rolled back correctly and stay consistent.

Given that 
* Most functions involve only one locally created transaction.
* The DB layer is one-level deep, meaning they just take one `tx`, run a query in it and return without passing the transaction further to other functions.

it's unlikely to lose track of a transaction. There is an exception that involves two transactions, fast and slow. Those two are already passed around like crazy. I have a plan to refactor and make that clearer later.

PR #919, opened 2022-07-08
vpctool: don't run bpfloader test if it will fail

tych0:
Most linux distros disallow unprivlieged bpf by default, so
`make test-local` in this configuration fails as an unprivleged user:

=== RUN   TestLoad

    bpfloader_linux_test.go:20:
        	Error Trace:	bpfloader_linux_test.go:20
        	Error:      	Received unexpected error:
        	            	operation not permitted
        	Test:       	TestLoad
--- FAIL: TestLoad (0.01s)

Let's detect this case and skip it.

### DIFF
--- a/vpc/tool/bpfloader/bpfloader_linux_test.go
+++ b/vpc/tool/bpfloader/bpfloader_linux_test.go
@@ -4,7 +4,11 @@
 package bpfloader
 
 import (
+	"io/ioutil"
 	"os"
+	"os/user"
+	"strconv"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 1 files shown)


MyUmmaGumma (approved):

solarkennedy (approved):

solarkennedy:
Is the test running under CI though?

tych0:
No, it skips because of the CIRCLECI env var: https://github.com/Netflix/titus-executor/blob/master/vpc/tool/bpfloader/bpfloader_linux_test.go#L14

presumably because they don't want people loading random blobs into their kernels. (maybe more reason to spend some time setting up a hardware virtualization based CI setup...)

vpc/tool/bpfloader/bpfloader_linux_test.go
```diff
@@ -14,6 +16,15 @@ func TestLoad(t *testing.T) {
 	if os.Getenv("CIRCLECI") == "true" {
 		t.Skip("Test does not work on Travis")
 	}
+
+	content, err := ioutil.ReadFile("/proc/sys/kernel/unprivileged_bpf_disabled")
+	assert.NoError(t, err)
+	u, err := user.Current()
+	assert.NoError(t, err)
+	if string(content) != "0\n" && u.Uid != "0" {
```
sargun, line 24: strconv.ParseInt?

tych0: Sure, although then we have to check another error. I can make the change on Monday.

PR #921, opened 2022-07-12
Delete reconcileBranchENIAttachment code

hechaoli:
# Background

VPC Service has a long-lived task that periodically "reconcile" branch ENI attachments. For each <region, account>, it fetches all associations from AWS and then do the reconciliation. 

# Why Delete the Function

If you look at `reconcileBranchENIAttachment` function call, which is the first step after fetching all associations from AWS, it actually does nothing other than selecting the association from DB. It does NOT insert associations in AWS but not in DB into the DB. From the comment, I think it's because this is a valid case. The association might have been deleted right before we fetch all associations from AWS. However, though the code does nothing when there is a "missing" association, it logs an error, which is more confusing than useful. 

From the git log, I found there used to be a `reconcileBranchENIAttachmentMissingFromDatabase` which was later deleted. My guess is, we used to insert "missing" attachments but later found this to be a valid case so that function was removed. Removal of that function also left a piece of code like
```
if err == sql.ErrNoRows {
    if err != nil {
         ...
         return err
    }
    // the rest of the code
}
```

in `reconcileBranchENIAttachment` which made "the rest of the code" to be dead code. This is also confusing.

This PR deletes `reconcileBranchENIAttachment` function completely to avoid confusion in code and in log.

# Other Thoughts
Despite of the valid case, it makes sense that the reconciliation ignore any attachments not in AWS . Because for branch ENI attachment, we should use DB as the source of truth since all branch ENI attachments are done through VPC Service.

### DIFF
--- a/vpc/service/reconcile_branch_eni_attachments.go
+++ b/vpc/service/reconcile_branch_eni_attachments.go
@@ -88,11 +88,6 @@ func (vpcService *vpcService) reconcileBranchAttachmentsENIsForRegionAccount(ctx
 	associationIDs := sets.NewString()
 	// 1. Check that there aren't any stray associations that we do not know about
 	for _, assoc := range associations {
-		ctx2 := logger.WithField(ctx, "associationID", aws.StringValue(assoc.AssociationId))
-		err = vpcService.reconcileBranchENIAttachment(ctx2, session, assoc)
-		if err != nil {
-			logger.G(ctx2).WithError(err).Error("Could not reconcile association")
-		}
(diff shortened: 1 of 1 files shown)


solarkennedy:
I agree there is a small wtf here.
Maybe it is just in place so we log the inconsistency?

hechaoli:
I don’t find this log useful. Other than knowing there are attachments deleted right before querying them on AWS. Such logs could easily be confused with real reconcile issues.

MyUmmaGumma (approved):

PR #924, opened 2022-07-20
runner: always delegate all cgroups

tych0:
This patch makes two changes:

1. delegate all cgroups. this is necessary for systemd to do resource
   accounting for its services, since it wants to be able to create subdirs
   and put tasks in there in order to do this accounting

2. always delegate cgroups instead of just in the systemd case. there's
   nothing really special about systemd here, it is conceivable that some
   other application (or more realistically, application runtime) may want
   to do its own sort of accounting using cgroups. let's not force people
   to run systemd in this case.

### DIFF
--- a/executor/runtime/docker/docker_linux.go
+++ b/executor/runtime/docker/docker_linux.go
@@ -282,10 +282,6 @@ func cleanupCgroups(cgroupPath string) error {
 }
 
 func setCgroupOwnership(parentCtx context.Context, c runtimeTypes.Container, cred ucred) error {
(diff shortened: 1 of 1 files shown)


solarkennedy (approved):

solarkennedy:
A task couldn't like, give itself more ram, could it?

tych0:
Only if we chown the `memory.*` files at the top level themselves, which we don't (and need to be careful) not to do. I've written a test for that in our unit tests.

PR #931, opened 2022-07-27
Add error counters and latency measures for long-running tasks

hechaoli:
# Problem

For APIs, we have gRPC stats provided by [opencensus](https://github.com/danielkhan/openconsensus/tree/master/schemas/opencensus). However, no stats/metrics exist for long-running tasks in VPC Service. 

# Fix
This PR adds an error counter and a latency measure for all long-running tasks. With this change, we ensure that for each failed execution, we count the error whereas for each successful execution, we measure its latency. Alerts will be set up based for these two metrics so that we will catch it early when certain tasks keep failing or have an increased execution time.

### DIFF
--- a/vpc/service/action_worker.go
+++ b/vpc/service/action_worker.go
@@ -39,7 +39,8 @@ type actionWorker struct {
 	finishedChanel  string
 	name            string
 	table           string
-	errorCounter    *stats.Int64Measure
+	errorMeasure    *stats.Int64Measure
+	latencyMeasure  *stats.Int64Measure
 
 	maxWorkTime time.Duration
 
@@ -221,6 +222,7 @@ func (actionWorker *actionWorker) worker(ctx context.Context, wq workqueue.RateL
 		 * worker loop (for the most part). We do not want to cause these things failing to cause the entire action
 		 * worker to bail.
 		 */
+		start := time.Now()
 		err = actionWorker.cb(ctx, tx, id)
 		// TODO: Consider updating the table state here
 		if vpcerrors.IsPersistentError(err) {
@@ -230,10 +232,12 @@ func (actionWorker *actionWorker) worker(ctx context.Context, wq workqueue.RateL
 		} else if err != nil {
 			tracehelpers.SetStatus(err, span)
 			logger.G(ctx).WithError(err).Error("Failed to process item")
-			stats.Record(ctx, actionWorker.errorCounter.M(1))
+			stats.Record(ctx, actionWorker.errorMeasure.M(1))
 			wq.AddRateLimited(item)
 			return nil
 		}
+		stats.Record(ctx, actionWorker.latencyMeasure.M(time.Since(start).Milliseconds()))
+
 		err = tx.Commit()
 		if err != nil {
 			err = errors.Wrap(err, "Could not commit database transaction")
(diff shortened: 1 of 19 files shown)


MyUmmaGumma (approved):
Super!

aloktiagi (commented):
changes look good. Trying to understand the motivation, are we looking to get better insights into AWS API call latencies or will these be useful in debugging issues.

Additionally do you see if these can be actionable in some sense? Like open an AWS ticket?

hechaoli:
> changes look good. Trying to understand the motivation, are we looking to get better insights into AWS API call latencies or will these be useful in debugging issues.

Not only AWS API latency but the overall monitoring of long-running tasks is missing. We don't know which ones fail when or how long does each task takes in general. Had we had the monitoring in place, we should have caught the state-size incident earlier before it was manifested as a state-size increase.

> Additionally do you see if these can be actionable in some sense? Like open an AWS ticket?
I will create alerts for long-running task errors and latencies. Debugging will be needed when they are fired. As of the next step action, it could be AWS ticket if it's proven to be AWS issue or fixing bugs in VPC service itself.

aloktiagi:
> Not only AWS API latency but the overall monitoring of long-running tasks is missing. We don't know which ones fail when or how long does each task takes in general. Had we had the monitoring in place, we should have caught the state-size incident earlier before it was manifested as a state-size increase.

I remember most of the tasks setting a context deadline. If a task's time to finish is equal to the context deadline, maybe we wont see an uptick in metrics? Or are you also measuring the number of times we hit the context deadline expiry?

hechaoli:
> I remember most of the tasks setting a context deadline. If a task's time to finish is equal to the context deadline, maybe we wont see an uptick in metrics? Or are you also measuring the number of times we hit the context deadline expiry?

There are 3 cases:
* The task finishes within the deadline. Then each successful execution will record a latency. We'll have an alert on the latency with a reasonable threshold based on historical data. Say, if a task usually takes 100ms to finish but suddenly it takes 1 seconds, then we should debug why.
* The task doesn't finish within the deadline, then an error counter will be increased. If it reaches a threshold, then an alert will be triggered and we should debug why it fails to finish within the deadline.
* The task doesn't time out but also doesn't finish successfully. Again, an error counter will be increased and we should debug why the failure rate increases.

aloktiagi:
> > I remember most of the tasks setting a context deadline. If a task's time to finish is equal to the context deadline, maybe we wont see an uptick in metrics? Or are you also measuring the number of times we hit the context deadline expiry?
> 
> There are 3 cases:
> 
> * The task finishes within the deadline. Then each successful execution will record a latency. We'll have an alert on the latency with a reasonable threshold based on historical data. Say, if a task usually takes 100ms to finish but suddenly it takes 1 seconds, then we should debug why.
> * The task doesn't finish within the deadline, then an error counter will be increased. If it reaches a threshold, then an alert will be triggered and we should debug why it fails to finish within the deadline.
> * The task doesn't time out but also doesn't finish successfully. Again, an error counter will be increased and we should debug why the failure rate increases.

Cool. Determining the threshold based on historical data has been a challenge in the past for me in other areas. But maybe the data might not be too noisy here. We can tune as we move along. I am sure you are planning to document these.

aloktiagi (approved):

hechaoli:
> Cool. Determining the threshold based on historical data has been a challenge in the past for me in other areas. But maybe the data might not be too noisy here. We can tune as we move along. I am sure you are planning to document these.

I'll update the runbook once the change is rolled out.

PR #932, opened 2022-07-28
runtime: don't start optional service if image pull fails

tych0:
If a docker image pull fails:

Jul 28 00:58:52 titusagent-main01cell002-m5metal009-i-0e2b0dca2ac46325d titus-executor-[370813]: Pull error message: map[error:received unexpected HTTP status: 500 Internal Server Error errorDetail:map[message:received unexpected HTTP status: 500 Internal Server Error]]
Jul 28 00:59:03 titusagent-main01cell002-m5metal009-i-0e2b0dca2ac46325d titus-executor-[370813]: Error pulling image 'registry.us-west-2.streamingprod.titus.netflix.net:7002/titusoss/titus-sshd@sha256:f50b6cac87b820b08c7783f7db7b4fda88937386d621cfa9ab690f148d2a1d38', due to reason: Error response from daemon: received unexpected HTTP status: 500 Internal Server Error
Jul 28 00:59:16 titusagent-main01cell002-m5metal009-i-0e2b0dca2ac46325d titus-executor-[370813]: Error pulling image 'registry.us-west-2.streamingprod.titus.netflix.net:7002/titusoss/titus-sshd@sha256:f50b6cac87b820b08c7783f7db7b4fda88937386d621cfa9ab690f148d2a1d38', due to reason: Error response from daemon: received unexpected HTTP status: 500 Internal Server Error
Jul 28 00:59:40 titusagent-main01cell002-m5metal009-i-0e2b0dca2ac46325d titus-executor-[370813]: Error pulling image 'registry.us-west-2.streamingprod.titus.netflix.net:7002/titusoss/titus-sshd@sha256:f50b6cac87b820b08c7783f7db7b4fda88937386d621cfa9ab690f148d2a1d38', due to reason: Error response from daemon: received unexpected HTTP status: 500 Internal Server Error
Jul 28 00:59:40 titusagent-main01cell002-m5metal009-i-0e2b0dca2ac46325d titus-executor-[370813]: Unable to setup optional sshd container 'sshd': Error response from daemon: received unexpected HTTP status: 500 Internal Server Error
Jul 28 00:59:40 titusagent-main01cell002-m5metal009-i-0e2b0dca2ac46325d titus-executor-[370813]: SystemD image label set
Jul 28 00:59:40 titusagent-main01cell002-m5metal009-i-0e2b0dca2ac46325d titus-executor-[370813]: Skipping volume container of optional sidecar sshd

then the system service will fail to start, because it's not present, which
is confusing:

Jul 28 00:59:54 titusagent-main01cell002-m5metal009-i-0e2b0dca2ac46325d titus-nsenter[380041]: BUG:do_nsenter:268:No such file or directory Could not execute child

let's just not start it at all.

The calls happen in two different phases of creation, Prepare() and
Start(), so we need to keep a copy of the system services pointers around
because SystemServices() explicitly makes a copy of them. Looks like this
was a TODO, so probably not a big deal.

### DIFF
--- a/executor/runtime/docker/docker_linux.go
+++ b/executor/runtime/docker/docker_linux.go
@@ -125,7 +125,7 @@ func stopSystemServices(ctx context.Context, c runtimeTypes.Container) error {
 	return nil
 }
 
-func setupSystemServices(parentCtx context.Context, c runtimeTypes.Container, cfg config.Config) error { // nolint: gocyclo
+func setupSystemServices(parentCtx context.Context, systemServices []*runtimeTypes.ServiceOpts, c runtimeTypes.Container, cfg config.Config) error { // nolint: gocyclo
 	ctx, cancel := context.WithTimeout(parentCtx, systemServiceStartTimeout)
 	defer cancel()
 
@@ -135,11 +135,6 @@ func setupSystemServices(parentCtx context.Context, c runtimeTypes.Container, cf
 	}
 	defer conn.Close()
 
-	// TODO: it would be nice not to fetch this twice
-	systemServices, err := c.SystemServices()
-	if err != nil {
-		return err
-	}
 	// TODO: Can we somehow make sure titus-container always starts first?
 	for _, svc := range systemServices {
 		if svc.EnabledCheck != nil && !svc.EnabledCheck(&cfg, c) {
@@ -147,6 +142,11 @@ func setupSystemServices(parentCtx context.Context, c runtimeTypes.Container, cf
 			continue
 		}
 		if svc.UnitName == "" {
+			continue
+		}
+
+		if svc.Image != "" && svc.ContainerName == "" {
+			logrus.Infof("Not starting %s, its container creation failed", svc.ServiceName)
 			continue
 		}
 
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -142,6 +142,7 @@ type DockerRuntime struct { // nolint: golint
 	startTime        time.Time
 	gpuManager       runtimeTypes.GPUManager
 	volumeContainers []string
+	systemServices   []*runtimeTypes.ServiceOpts
 }
 
 type Opt func(ctx context.Context, runtime *DockerRuntime) error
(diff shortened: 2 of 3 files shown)


executor/runtime/docker/docker_linux.go
```diff
@@ -135,11 +135,6 @@ func setupSystemServices(parentCtx context.Context, c runtimeTypes.Container, cf
 	}
 	defer conn.Close()
 
-	// TODO: it would be nice not to fetch this twice
```
solarkennedy, line 138: Thanks!

executor/runtime/docker/docker_linux.go
```diff
@@ -150,6 +145,11 @@ func setupSystemServices(parentCtx context.Context, c runtimeTypes.Container, cf
 			continue
 		}
 
+		if svc.PullFailed {
```
solarkennedy, line 148: Do we really need a new field here?

Can we not use `svc.ContainerName` ? If it is unset (==""), then the volume container didn't get created for whatever reason, so is unsuitable to start.

tych0: Right now the name is set in the middle of the function, and the actual `ContainerCreate` call could fail and we'd still wind up in the same situation. I'll move setting it to the end and drop the field.

executor/runtime/docker/docker_linux.go
```diff
@@ -150,6 +145,11 @@ func setupSystemServices(parentCtx context.Context, c runtimeTypes.Container, cf
 			continue
 		}
 
+		if svc.Volumes != nil && svc.ContainerName == "" {
```
solarkennedy, line 148: Would it be more natural to say:

   if svc.Image != "" && svc.ContainerName == ""

Who cares if it has volumes, really.
Sure in the cases recently they happened to have volumes.
But the more important thing is if it had an *image*.
Cause if it had an image, but no container, it means we can't continue.

tych0: Yes, yes it would, thank you.

solarkennedy (approved):

PR #940, opened 2022-08-04
reaper: check for wedged fuse tasks too

tych0:
We see this a fair amount, so let's try and unwedge these things. This
works for me on some of our hosts in test, which have wedged tasks:

time="2022-08-04T21:08:31Z" level=debug msg="found wedged pidns 2249373 2913cd68-e39e-487c-847a-8698a96f868d"
time="2022-08-04T21:08:31Z" level=debug msg="checking 2250865 for request_wait_answer"
time="2022-08-04T21:08:31Z" level=debug msg="checking 2250917 for request_wait_answer"
time="2022-08-04T21:08:31Z" level=info msg="reaping task 2913cd68-e39e-487c-847a-8698a96f868d because fuse is wedged. fuse minor is: 47"
time="2022-08-04T21:08:31Z" level=info msg="reaping task 2913cd68-e39e-487c-847a-8698a96f868d because fuse is wedged. fuse minor is: 47"
time="2022-08-04T21:08:31Z" level=info msg="reaping task 2913cd68-e39e-487c-847a-8698a96f868d because fuse is wedged. fuse minor is: 47"

### DIFF
--- a/reaper/reaper.go
+++ b/reaper/reaper.go
@@ -139,6 +139,8 @@ func processContainerJSON(ctx context.Context, container types.ContainerJSON, do
 		return fmt.Errorf("Could not determine whether or not process with exe %q / and stat %v was a titus executor", link, stat)
 	}
 
+	checkIfFuseWedgedPidNs(container.State.Pid, taskID)
+
 	l.WithFields(map[string]interface{}{
 		"link": link,
 		"exe":  exe,
--- a/reaper/reaper_linux.go
+++ b/reaper/reaper_linux.go
@@ -1,0 +1,140 @@
+//go:build linux
+// +build linux
+
+package reaper
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"strconv"
+	"strings"
+
+	"github.com/moby/sys/mountinfo"
+	"github.com/sirupsen/logrus"
+)
+
+// there is a kernel bug described in:
+// https://lore.kernel.org/all/YrShFXRLtRt6T%2Fj+@risky/ where fuse can cause a
+// pidns exit to deadlock. They symptoms are:
+//
+//    1. pidns pid 1 in S (sleeping) state
+//    2. that pid has zap_pid_ns_processes() in its /proc/pid/stack
+//    3. some fuse mount exists, and one of the threads from that fuse mount is
+//       stuck in fuse's request_wait_answer()
+//
+// if those conditions are true, we need to manually tear down the fuse
+// connection so the pidns can exit and docker doesn't get stuck. we can do
+// this by writing something to
+//
+// 	/sys/fs/fuse/connections/$dev_minor/abort
+//
+// where $dev_minor is the minor number from the fuse superblock mount.
+func checkIfFuseWedgedPidNs(pid int, taskID string) {
+	if !kernelStackHas(pid, "zap_pid_ns_processes") {
+		return
+	}
+	logrus.Debugf("found wedged pidns %d %s", pid, taskID)
+
+	// the kernel has already destroyed the mountinfo for the pidns' init,
+	// since it is pretty far along in do_exit(). we need to keep the tid
+	// of the fuse process around so we can inspect its mountinfo instead.
+	//
+	// ideally we'd use the pids cgroup (i.e. the docker API) here to
+	// figure out what tasks we should look at, but that *also* has been
+	// invalidated and is incorrect at this point. luckily for us the fuse
+	// daemon that's causing this hang in our production case is a child of
+	// init, so we can just look at that.
+	targetTid := 0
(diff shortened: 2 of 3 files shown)


reaper/reaper_linux.go
```diff
@@ -0,0 +1,133 @@
...
+		logrus.Errorf("couldn't read task dir for %d: %v", pid, err)
+		return nil
+	}
+
+	threads := []int{}
+	for _, e := range strings.Fields(string(threadsRaw)) {
+		child, err := strconv.Atoi(e)
+		if err != nil {
+			logrus.Errorf("bad child for %d: %s %v", pid, e, err)
+			return nil
```
fabiokung, line 114: what about a `continue` and just skipping this particular "bad" pid?

reaper/reaper_linux.go
```diff
@@ -0,0 +1,133 @@
...
+		}
+
+		ents, err := ioutil.ReadDir(fmt.Sprintf("/proc/%d/task", child))
+		if err != nil {
+			logrus.Errorf("couldn't read threads for %d: %v", child, err)
+			return nil
```
fabiokung, line 120: same here, `continue`?

reaper/reaper_linux.go
```diff
@@ -0,0 +1,133 @@
...
+		return false
+	}
+
+	// the format of this file has changed somewhat over time; here's a
+	// reasonable guess at e.g.
+	//      [<0>] vfs_read+0x9c/0x1a0
+	return strings.Contains(string(content), function+"+0x")
+}
+
+func childThreadsOfPid(pid int) []int {
```
fabiokung, line 99: minor: comment that this is a shallow search (only direct children), or rename to `directChildThreadsOfPid(pid)`

reaper/reaper_linux.go
```diff
@@ -0,0 +1,133 @@
...
+		ents, err := ioutil.ReadDir(fmt.Sprintf("/proc/%d/task", child))
+		if err != nil {
+			logrus.Errorf("couldn't read threads for %d: %v", child, err)
+			return nil
+		}
+		for _, ent := range ents {
+			tid, err := strconv.Atoi(ent.Name())
+			if err != nil {
+				logrus.Errorf("bad thread for %d: %s %v", child, ent.Name(), err)
+				return nil
```
fabiokung, line 126: same, `continue`?

reaper/reaper_linux.go
```diff
@@ -0,0 +1,133 @@
...
+	// figure out what tasks we should look at, but that *also* has been
+	// invalidated and is incorrect at this point. luckily for us the fuse
+	// daemon that's causing this hang in our production case is a child of
+	// init, so we can just look at that.
+	targetTid := 0
+	for _, tid := range childThreadsOfPid(pid) {
+		logrus.Debugf("checking %d for request_wait_answer", tid)
+		if kernelStackHas(tid, "request_wait_answer") {
+			targetTid = tid
+			break
```
fabiokung, line 53: what's the chance of anyone doing something weird with multiple fuse daemons?

fabiokung (approved):

reaper/reaper_linux.go
tych0: I think we'll pick it up below if everything is in one mount namespace, since we look at all the mount infos (this is just figuring out which thread we can actually use, since the init pid's mountinfo has been freed already).

Containers could create their own mount namespace with multiple fuses, but I haven't seen that in practice with the lockups we're experiencing.

reaper/reaper_linux.go
tych0: yes, all good points. it's a kernel API so it really shouldn't break, but if it does we should try to cleanup whatever we can.

tych0:
One not-so-minor update to this based on additional testing: it now skips aborting the lxcfs mount's connection, which is a fuse filesystem that's bind mounted in from the host.

reaper/reaper.go
```diff
@@ -139,6 +139,8 @@ func processContainerJSON(ctx context.Context, container types.ContainerJSON, do
 		return fmt.Errorf("Could not determine whether or not process with exe %q / and stat %v was a titus executor", link, stat)
 	}
 
+	checkIfFuseWedgedPidNs(container.State.Pid, taskID)
```
solarkennedy, line 142: shouldn't this go as a `else` on line 124?
(we only want to run this if the pid exists)

solarkennedy (approved):

reaper/reaper.go
tych0: `err` could be other things there (EINVAL maybe if things are really screwed up), we check for these non-ENOENT errors on line 126 and early return, so I think this is only called when the pid exists (and when the executor pid exists).

PR #944, opened 2022-08-06
Vpc service gc transition assignment

hechaoli:
# Background

When an `AssignIPV3` API is called with `transitionRequested=true`, it will create a transition assignment if one doesn't already exist. Then VPC tool receives the transition assignment in the response. VPC tool will create a transition namespace and mount it to `/run/transition/<transition-assignment-id>` if one doesn't exist. On one instance, multiple containers using the same branch ENI will share one transition assignment. 

# Problem

Currently, the transition assignments are not recycled. That means, even if a transition assignment is no longer used by any containers, the transition namespace will live on the instance forever and the IP allocated to this transition assignment will not be recycled. This will waste some IPv4 addresses and thus increase the state size.

# Solution

This PR reuses the GC assignment mechanism to also GC transition assignment. Specifically, 
1) In `GCV3` API, set tombstone for all unused transition assignments on the instance in addition to regular assignments. 
2) When VPC tool receives assignments to remove from `GCV3` API response, for regular assignments, tear down the container network as usual; for transition assignments, tear down the transition network namespace.
3) After tearing down the transition network namespace, VPC tool calls `UnassignIPV3` API to eventually delete the transition assignment from VPC service and thus release its IPv4 address.

# Race condition
One race condition to be aware is that, it's possible that when `UnassignIPV3` is called to delete the transition assignment, another concurrent `AssignIPV3` API may decide to reuse it. Because 1) both transition assignment reuse and deletion are in a transaction and 2) the reuse transaction is a serializable transaction, we can cover the case by a) set tombstone to NULL when reusing and b) delete only if tombstone is set.

Then, there are only 2 cases:

## Case 1: Delete then update
|  TX A                                                                                       | TX B                                                                                     |
| -------------------------------------------------------- | ------------------------------------------------------ |
| `UnassignIPV3` deletes transition assignments "t-123"   |                                                                                              |
|                                                                                                  |  `AssignIPV3` updates "t-123" tombstone to be NULL |
|  COMMIT                                                                                 |                                                                                              |
|                                                                                                  |   Rollback since the row is deleted, will retry                   |

## Case 2: Update then delete
|  TX A                                                                                       | TX B                                                                                     |
| -------------------------------------------------------- | ------------------------------------------------------ |
|                                                                                                  | `AssignIPV3` updates "t-123" tombstone to be NULL  |
| `UnassignIPV3` deletes transition assignments "t-123"   |                                                                                              |
|                                                                                                  |  COMMIT                                                                             |
|   Rollback since the tombstone is NULL, will skip               |                                                                                              |

### DIFF
--- a/vpc/service/gc_v3.go
+++ b/vpc/service/gc_v3.go
@@ -109,6 +109,76 @@ func (vpcService *vpcService) GCV3(ctx context.Context, req *vpcapi.GCRequestV3)
 	return resp, nil
 }
 
+// Set gc_tombstone time to now() on any assginments that match the following condition:
+// 1) The assignment uses the given trunk ENI.
+// 2) The assignment_id is not in the given running task list.
+// 3) The assignment has been created for more than 5 minutes.
+// 4) Tha assignment is not a transition assignment.
+// 5) The gc_tombstone is not already set .
+func tombstoneUnusedAssignments(ctx context.Context, tx *sql.Tx, trunkENI string, runningTasks []string) (int64, error) {
+	rowsAffected, err := tx.ExecContext(ctx, `
+WITH unused_assignments AS
+  (SELECT assignments.id
+   FROM branch_eni_attachments
+   JOIN assignments ON branch_eni_attachments.association_id = assignments.branch_eni_association
+   WHERE trunk_eni = $1
+     AND assignments.assignment_id NOT IN
+       (SELECT unnest($2::text[]))
+     AND assignments.created_at < now() - INTERVAL '5 minutes'
+     AND NOT assignments.is_transition_assignment
+     AND assignments.gc_tombstone IS NULL)
+UPDATE assignments
+SET gc_tombstone = now()
+FROM unused_assignments
+WHERE assignments.id = unused_assignments.id
+`, trunkENI, pq.Array(runningTasks))
+	if err != nil {
+		err = fmt.Errorf("Could not tombstone unused assignments: %w", err)
+		return 0, err
+	}
+
+	n, err := rowsAffected.RowsAffected()
+	if err != nil {
+		err = fmt.Errorf("Could not find rows affected: %w", err)
+		return 0, err
+	}
+	return n, nil
+}
+
+// Get all assignments that can be removed. An assignment can be removed if it meets all following conditions:
+// 1) The assignment uses the given trunk ENI.
+// 2) The assignment_id is not in the given running task list.
+// 3) The gc_tombstone on the assignment has been set for more than 30 minutes
+func getAssignmentsToRemove(ctx context.Context, tx *sql.Tx, trunkENI string, runningTasks []string) ([]string, error) {
+	rows, err := tx.QueryContext(ctx, `
+	SELECT assignment_id
+	FROM branch_eni_attachments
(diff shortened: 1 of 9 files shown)


vpc/tool/container2/setup_container_linux.go
```diff
@@ -762,6 +762,48 @@ func TeardownNetwork(ctx context.Context, assignment *vpcapi.AssignIPResponseV3)
...
+func TeardownTransitionNetwork(ctx context.Context, transitionNamespaceDir string, transitionNamespaceID string) error {
+	logger.G(ctx).Infof("Tearing down transition network %s", transitionNamespaceID)
+	_, cleanup, err := transition.LockTransitionNamespaces(ctx, transitionNamespaceDir)
+	if err != nil {
+		return fmt.Errorf("Could not lock transition namespace dir: %w", err)
+	}
+	defer cleanup()
+
+	filePath := filepath.Join(transitionNamespaceDir, transitionNamespaceID)
+	err = unix.Unmount(filePath, unix.MNT_DETACH)
```
tych0, line 775: Is a MNT_DETACH required here? If we don't actually unmount here, will that cause any problems with interface destruction?

vpc/service/gc_v3.go
```diff
@@ -106,9 +106,123 @@ func (vpcService *vpcService) GCV3(ctx context.Context, req *vpcapi.GCRequestV3)
...
+		unused_transition_assignments AS (
+			SELECT assignments.id 
+			FROM assignments
+			JOIN branch_eni_attachments ON branch_eni_attachments.association_id = assignments.branch_eni_association
+			WHERE trunk_eni = $1
+			AND assignments.is_transition_assignment
+			AND NOT EXISTS (
+				SELECT id from used_transition_assignments where used_transition_assignments.id = assignments.id
+			)
+			AND assignments.transition_last_used < now() - INTERVAL '5 minutes'
```
aloktiagi, line 172: Until we have a way to have a unique transition assignment per instance/trunk-eni/sg, do you think it might be better for this to be some value greater than 5 mins? Could we run into a situation where we end up having multiple transition assignments created over and over again?

vpc/tool/container2/setup_container_linux.go
hechaoli: Good question. I think it's not required. Because when this is called, it shouldn't be used by any container.

vpc/service/gc_v3.go
hechaoli: Good question. Note that there are 2 time durations to be aware of in a GC process. This applies to both regular and transition assignment GC. 
* 5 minutes (here): When an assignment is unused, mark `gc_timestone` only after it hasn't been used for 5 minutes.
* 30 minutes (see the query in `getAssignmentsToRemove` function): Once an assignment is marked for GC, return it in `GCV3` response so that VPC tool will tear down the network only after 30 minutes. 

As a result, an assignment will only be deleted 35 minutes after it's unused. I think that's enough time to wait before GC.

aloktiagi: Sounds good!

aloktiagi (approved):

tych0 (approved):

PR #945, opened 2022-08-11
Fix a VPC Service crash

hechaoli:
### Description of the Change

In [this PR](https://github.com/Netflix/titus-executor/pull/935), I changed `branch` field in `assignment` to be a pointer but forgot to initialize it. That caused a nil pointer reference error during `AssignIPV3` call.

The problem was caught in staging. Ideally, there should be an E2E CI test to catch it. The test case is to be added.

### DIFF
--- a/vpc/service/generate_assignment_id.go
+++ b/vpc/service/generate_assignment_id.go
@@ -148,6 +148,7 @@ func (vpcService *vpcService) generateAssignmentID(ctx context.Context, req getE
 
 	ass := &assignment{
 		trunk:            req.trunkENI,
+		branch:           &data.BranchENI{}, // Will be populated in populateAssignment
 		assignmentName:   req.assignmentID,
 		securityGroups:   req.securityGroups,
 		subnet:           req.subnet,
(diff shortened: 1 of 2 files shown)


vpc/service/generate_assignment_id.go
```diff
@@ -148,6 +148,7 @@ func (vpcService *vpcService) generateAssignmentID(ctx context.Context, req getE
 
 	ass := &assignment{
 		trunk:            req.trunkENI,
+		branch:           &data.BranchENI{}, // Will be populated in populateAssignment
```
solarkennedy, line 151: Is it so bad that it is nil at first?

Why not add this as part of the populateAssignmentUsingAlreadyAttachedENI() function? 

```
populateAssignmentUsingAlreadyAttachedENI(ass) {
 
ass.branch = &data.BranchENI{}
pg.bla....
```

hechaoli: I added it here just in case there are things in between this initialization and `populateAssignmentUsingAlreadyAttachedENI` that use `ass.branch`. Though there should be none for now.

Added another check in `populateAssignmentUsingAlreadyAttachedENI` before the usage, too.

MyUmmaGumma (approved):

aloktiagi (approved):

PR #947, opened 2022-08-12
Removed tini sched bump feature

solarkennedy:
This feature has been disabled for about two years.
I think we don't need it.


nickbp:
don't have full context but LGTM

nickbp (approved):

PR #948, opened 2022-08-15
seccomp: drop keyctl from fuse profile

tych0:
this was added as a quick hack in 063a7ead6cc2 ("Enabled the keyctl syscall
for fuse workloads").

still not really clear to me why we need to allow this at all. seems weird
that image builders would be managing the kernel keyring, but whatever...


tych0:
oh, this should drop pivot_root too. Let me do that.

solarkennedy (approved):

tych0:
```
E: Could not get lock /var/lib/dpkg/lock-frontend - open (11: Resource temporarily unavailable)
--
E: Unable to acquire the dpkg frontend lock (/var/lib/dpkg/lock-frontend), is another process using it?
```

strange.

PR #949, opened 2022-08-15
CODEOWNERS: fix wildcard match

tych0:
The docs https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#example-of-a-codeowners-file
say:

    # Order is important; the last matching pattern takes the most
    # precedence.

so I think Kyle would never get a review request; let's align this with the
rest of the file and just put everything on the same line.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,5 +1,4 @@
-* @solarkennedy
-* @tych0
+* @solarkennedy @tych0 @nickbp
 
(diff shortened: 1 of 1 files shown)


solarkennedy (approved):
@nickbp  would you also like to help review code for this codebase?

nickbp:
Yes that sounds good. Can either add me here or I can create a separate PR once this is merged

nickbp (approved):

tych0:
I went ahead and added you; will let this pass CI and then merge.

PR #950, opened 2022-08-19
Enable privileged containers in standalone mode

solarkennedy:
As we enable titus-standalone to run pods on ec2 agents, we know we want
the ultimate escape hatch of allowing privileged containers.

This enables that, but read via the pod spec (which we have now), but
only in standalone mode.

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -364,7 +364,7 @@ func maybeAddOptimisticDad(sysctl map[string]string) {
 	}
 }
 
-func (r *DockerRuntime) mainContainerDockerConfig(c runtimeTypes.Container, binds []string, imageSize int64, volumeContainers []string) (*container.Config, *container.HostConfig, error) { // nolint: gocyclo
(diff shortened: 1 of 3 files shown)


executor/runtime/docker/docker.go
```diff
@@ -2815,3 +2802,16 @@ func shouldClose(c io.Closer) {
...
+	if pod == nil {
+		return false
+	}
+	if pod.Spec.SecurityContext == nil {
+		return false
+	}
+	if len(pod.Spec.Containers) < 1 {
+		return false
+	}
+	return *pod.Spec.Containers[0].SecurityContext.Privileged && r.cfg.InStandaloneMode
```
nickbp, line 2816: Could there be a situation where there's >1 containers, and where one of the OTHER ones is marked priveleged? (In other words does it make sense to iterate over the slice)

nickbp (approved):
Don't have full context on the usage but the changes themselves LGTM

tych0 (approved):

PR #951, opened 2022-08-22
docker: set fd limit for containers

tych0:
There was a limit before, but it was docker's default at a bit over 1M:

open files                          (-n) 1048576

let's drop it to a more reasonable 100k.

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -477,7 +477,12 @@ func (r *DockerRuntime) mainContainerDockerConfig(c runtimeTypes.Container, bind
 		Soft: ((c.Resources().Disk * MiB) + 1*GiB),
 		Hard: ((c.Resources().Disk * MiB) + 1*GiB),
 	}
-	hostCfg.Ulimits = []*units.Ulimit{coreLimit}
+	fdLimit := &units.Ulimit{
+		Name: "nofile",
+		Soft: 100000,
+		Hard: 100000,
+	}
+	hostCfg.Ulimits = []*units.Ulimit{coreLimit, fdLimit}
 
 	// This is just factored out mutation of these objects to make the code cleaner.
 	r.setupLogs(c, hostCfg)


tych0:
Looks like this is what Kyle is trying to fix:

```
ERROR:  Error installing fpm:
	The last version of public_suffix (< 6.0, >= 2.0.2) to support your Ruby & RubyGems was 4.0.7. Try installing it with `gem install public_suffix -v 4.0.7` and then running the current command again
	public_suffix requires Ruby version >= 2.6. The current ruby version is 2.5.0.
```

solarkennedy:
please rebase

executor/runtime/docker/docker.go
```diff
@@ -477,7 +477,12 @@ func (r *DockerRuntime) mainContainerDockerConfig(c runtimeTypes.Container, bind
 		Soft: ((c.Resources().Disk * MiB) + 1*GiB),
 		Hard: ((c.Resources().Disk * MiB) + 1*GiB),
 	}
-	hostCfg.Ulimits = []*units.Ulimit{coreLimit}
+	fdLimit := &units.Ulimit{
+		Name: "nofile",
+		Soft: 100000,
```
solarkennedy, line 482: This seems low?

I swear I've changed this in the past, but I guess it didn't work because it must have been in some other codebase.

idk, of course the most important part is that there is *some* bound.
How high can we get away with, can we do 1mil ?

tych0: The docker default is ~1M, so that's where things are currently at. In terms of what we can get away with, I don't know. The bug that triggered this PR was tasks leaking 400-700k fds, and nothing fell over. However, that is so far above what we usually use that the graphs looked scary. Maybe that's fine, though.

nickbp: 100k is pretty high, it's about what some high-fd services would recommend for production:
- [kafka](https://docs.confluent.io/platform/current/kafka/deployment.html): "You should increase your file descriptor counttoat least 100,000."
- [cassandra](https://docs.datastax.com/en/cassandra-oss/3.x/cassandra/install/installRecommendSettings.html): "nofile 100000"

Is that 1M docker limit across all containers? If so then 100k/container would probably make sense to reduce the likelihood of multiple high-fd containers conflicting

tych0: > Is that 1M docker limit across all containers? If so then 100k/container would probably make sense to reduce the likelihood of multiple high-fd containers conflicting

No, it is per container.

solarkennedy (approved):

PR #952, opened 2022-08-22
Use nfpm instead of fpm for packaging

solarkennedy:
Ruby kinda sucks. nfpm is more self-contained.

Here is the info of a deb created this way:

```
 new Debian package, version 2.0.
 size 141167754 bytes: control archive=2153 bytes.
       1 bytes,     0 lines      conffiles
     439 bytes,    11 lines      control
    4382 bytes,    55 lines      md5sums
     179 bytes,     4 lines   *  postinst             #!/bin/bash
 Package: titus-executor
 Version: 20210319.0.0~922-g295e3e86
 Section: default
 Priority: optional
 Architecture: amd64
 Maintainer: titus-developers@netflix.com
 Installed-Size: 314583
 Depends: libc6, apparmor >= 2.12
 Recommends: docker-ce >= 5:18.09.1~3-0~ubuntu-xenial, lxcfs, atlas-titus-agent, nvidia-container-runtime-hook
 Homepage: https://github.com/Netflix/titus-executor
 Description: titus-executor is spawned by titus-virtual-kubelet
```

There is some loss here, but nothing major.

### DIFF
--- a/hack/builder/Dockerfile
+++ b/hack/builder/Dockerfile
@@ -9,11 +9,11 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get update && apt-get -y install curl && 
 
 RUN export DEBIAN_FRONTEND=noninteractive && apt-get update && \
     apt-get install -y build-essential make cmake g++ gcc libc6-dev pkg-config \
-        libattr1-dev git curl wget jq ruby ruby-dev rubygems lintian unzip bison flex clang llvm musl-tools \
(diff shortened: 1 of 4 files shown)


tych0:
Debian packaging.... bleh. Anyway, lgtm whenever CI is happy. Thanks for doing this.

tych0 (approved):

solarkennedy:
Ok green checkmark. We need to be a little bit on the lookout for any deb-related breakage if the conversion didn't translate for some reason.

PR #955, opened 2022-08-23
Added support for host bind mounts in standalone mode

solarkennedy:
This was a thing I added in my titus-hybrid-demo.

This allows us to configure host bind mounts on a pod.

For security reasons, this is only supported in standalone mode.

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -991,7 +991,12 @@ func (r *DockerRuntime) Prepare(ctx context.Context) (err error) { // nolint: go
 		bindMounts = append(bindMounts, getKernelBindMounts()...)
 	}
 	pod, podLock := r.c.Pod()
+	mainContainerBindMounts, err := r.getBindMountsForContainer(pod.Spec.Containers[0], *pod)
 	podLock.Unlock()
+	if err != nil {
+		return err
+	}
+	bindMounts = append(bindMounts, mainContainerBindMounts...)
 
 	dockerCfg, hostCfg, err := r.mainContainerDockerConfig(r.c, bindMounts, imageSize, volumeContainers, r.isContainerPrivileged(pod.Spec.Containers[0]))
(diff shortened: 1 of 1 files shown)


executor/runtime/docker/docker.go
```diff
@@ -981,12 +981,17 @@ func (r *DockerRuntime) Prepare(ctx context.Context) (err error) { // nolint: go
 	}
 	r.volumeContainers = volumeContainers
 
+	pod, podLock := r.c.Pod()
+	podLock.Unlock()
```
tych0, line 985: I realize you just copy/pasted this, but I think there's a bug and this should have a `defer`.

solarkennedy: Heh, yea. this broke the build.

solarkennedy: What if we had a new function called `PodRO()` that returned a readonly copy of the pod, so we didn't have to worry about a nil or something, and then we could get the pod spec safely without having to worry about locks? (only if we don't need to mutate anything)

tych0: Yeah, as long as it doesn't matter that the view is maybe inconsistent. But IIRC it doesn't since there's only the one writer. That seems much better.

executor/runtime/docker/docker.go
```diff
@@ -991,7 +991,12 @@ func (r *DockerRuntime) Prepare(ctx context.Context) (err error) { // nolint: go
 		bindMounts = append(bindMounts, getKernelBindMounts()...)
 	}
 	pod, podLock := r.c.Pod()
+	mainContainerBindMounts, err := r.getBindMountsForContainer(pod.Spec.Containers[0], *pod)
+	if err != nil {
+		return err
```
tych0, line 996: podLock.Unlock() here, or just move it above the err check?

solarkennedy: good idea

PR #957, opened 2022-08-31
builder: drop BUILD_NUMBER code path

tych0:
I think this is probably (?) leftover from a switch from jenkins. circleci
does not export BUILD_NUMBER at all:

https://app.circleci.com/pipelines/github/Netflix/titus-executor/2268/workflows/8f7aaa41-fe61-490c-b7da-291961aac486/jobs/21319?invite=true#step-99-2

let's add -iteration to the default $version and delete this code path so
it doesn't confuse anyone else.


tych0:
`+ filename=/tmp/tmp.l5r60iTClP/titus-executor_20220830.0.0~3-ga1599380-1661968568_amd64.deb` seems better.

solarkennedy (approved):

PR #959, opened 2022-09-01
docker: log image pull time

tych0:
By default, we don't log image pull lines, buffering them and only logging
them when an image pull fails. This results in weird timestamps though: all
the image pull lines are logged *after* the failure, making it look like
the image pull was nearly instantaneous, which for very large images it is
not.

We probably do this to keep down on log noise, which is reasonable. Let's
add one more line with the actual elapsed time in the failure case so that
it's more obvious when debugging how long this took.

### DIFF
--- a/executor/runtime/docker/docker_image_pull.go
+++ b/executor/runtime/docker/docker_image_pull.go
@@ -53,6 +53,8 @@ func doDockerPull(ctx context.Context, cfg config.Config, metrics metrics.Report
 		metrics.Counter("titus.executor.dockerPullImageError", 1, nil)
 		return fmt.Errorf("Error creating docker registry credentials: %w", err)
 	}
+
+	pullStarted := time.Now()
 	resp, err := client.ImagePull(ctx, ref, types.ImagePullOptions{RegistryAuth: auth})
 	defer func() {
 		if resp != nil {
@@ -88,6 +90,7 @@ func doDockerPull(ctx context.Context, cfg config.Config, metrics metrics.Report
 			for _, pullMessage := range pullMessages {
 				log.Warning("Pull Message: ", pullMessage)
 			}
+			log.Infof("Image %s pull took: %s", ref, time.Since(pullStarted).String())
 			return err
 		}
 		pullMessages = append(pullMessages, msg)


tych0:
I'm still investigating how we can surface a reasonable error here (and elsewhere) when things fail due to stuff like this, but it seams like that'll be a larger change. At least this will help us debug such cases in the meantime.

tych0:
I dunno. If there are any metrics about this they might end up slightly weird, since we care about the error path too. The patch looks something like this:
```
diff --git a/executor/runtime/docker/docker.go b/executor/runtime/docker/docker.go
index 3143ffdf..3fd2f79b 100644
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -640,11 +640,12 @@ func (r *DockerRuntime) DockerPull(ctx context.Context, c runtimeTypes.Container
        r.metrics.Counter("titus.executor.dockerImagePulls", 1, nil)
        logger.Infof("DockerPull: pulling image")
        pullStartTime := time.Now()
-       if err := pullWithRetries(ctx, r.cfg, r.metrics, r.client, c.QualifiedImageName(), doDockerPull); err != nil {
+       err := pullWithRetries(ctx, r.cfg, r.metrics, r.client, c.QualifiedImageName(), doDockerPull)
+       r.metrics.Timer("titus.executor.imagePullTime", time.Since(pullStartTime), c.ImageTagForMetrics())
+       if err != nil {
                return nil, err
        }
-       r.metrics.Timer("titus.executor.imagePullTime", time.Since(pullStartTime), c.ImageTagForMetrics())
        return nil, nil
 }
```

which would catch everything that fails, even if a user typos something, etc. I suppose we could check to see if the context is cancelled, and log only then too, but that also feels like we're lying since the image wasn't really successfully downloaded. Maybe that would be easier to detect though? I don't know how that metric is surfaced.

solarkennedy:
I don't exactly understand what you are saying, but maybe we just need more logging in different places.

The outer log should be info/warn about the overall sucess/fail of the pull.
The inter log can be debug about the exact pull attempt?

tych0:
> I don't exactly understand what you are saying

I'm just worried if we include the error path we'll skew `titus.executor.imagePullTime` incorrectly.

> maybe we just need more logging in different places.

Yes, I think this is what the patch does.

> The inter log can be debug about the exact pull attempt?

Right now we log the docker API's output when the thing fails in this inner log, which is lots of lines (think mostly useless progress bars, since the timestamps aren't right). This adds one more line to that inner logging with an explicit duration to hopefully trigger peoples' memories that the timestamps associated with the logs are wrong because of the buffering.

executor/runtime/docker/docker_image_pull.go
```diff
@@ -88,6 +90,7 @@ func doDockerPull(ctx context.Context, cfg config.Config, metrics metrics.Report
 			for _, pullMessage := range pullMessages {
 				log.Warning("Pull Message: ", pullMessage)
 			}
+			log.Infof("Image pull took: %s", time.Since(pullStarted).String())
```
solarkennedy, line 93: go ahead and throw in the name of the image, since there will be multiple pulls per titus-executor

PR #963, opened 2022-09-08
TITUS-6224: add extra lxcfs endpoints

tych0:
in particular, we want /sys/devices/system/cpu/online, but with lxcfs 3.1.2 we get these other ones for free, so why not.

This needs to wait for merge until we have a new version of lxcfs that actually exports these files, otherwise these bind mounts will fail. But here's the patch for now so I don't lose track of it.

### DIFF
--- a/executor/runtime/docker/lxcfs.go
+++ b/executor/runtime/docker/lxcfs.go
@@ -13,11 +13,21 @@ const (
 
 func getLXCFsBindMounts() []string {
 	extraBinds := []string{}
-	for _, file := range []string{"cpuinfo", "meminfo", "uptime"} {
-		path := filepath.Join(lxcfs, "proc", file)
+	lxcfsEndpoints := []string{
+		"/proc/cpuinfo",
+		"/proc/diskstats",
+		"/proc/meminfo",
+		"/proc/stat",
+		"/proc/swaps",
+		"/proc/uptime",
+		"/proc/slabinfo",
+		"/sys/devices/system/cpu",
+		"/sys/devices/system/cpu/online",
+	}
+	for _, file := range lxcfsEndpoints {
+		path := filepath.Join(lxcfs, file)
 		if err := unix.Access(path, 0); err == nil {
-			destPath := filepath.Join("/proc", file)
-			extraBinds = append(extraBinds, fmt.Sprintf("%s:%s", path, destPath))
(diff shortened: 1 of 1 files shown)


executor/runtime/docker/lxcfs.go
```diff
@@ -13,11 +13,19 @@ const (
 
 func getLXCFsBindMounts() []string {
 	extraBinds := []string{}
-	for _, file := range []string{"cpuinfo", "meminfo", "uptime"} {
-		path := filepath.Join(lxcfs, "proc", file)
+	lxcfsEndpoints := []string{
```
nickbp, line 16: I was looking at info about lxcfs and noticed [this page](https://github.com/lxc/lxcfs#introduction) which sounds like lxcfs tries to generate container-scoped stats. Are these bind mounts in support of that, or are these instead for "punching a hole" up to the host-level stats? FWIW that page also lists `/sys/devices/system/cpu` (no idea how this interacts with `/sys/devices/system/cpu/online`) and `/proc/slabinfo`.

PS: could alphabetize the list - but the above lxcfs doc doesn't fully alphabetize theirs either

solarkennedy:
As long as these are not more misleading compared to if we didn't expose them at all.

Why do we need/want the `online` file?

executor/runtime/docker/lxcfs.go
tych0: > Are these bind mounts in support of that,

Yes.

> /sys/devices/system/cpu

We can't easily take a version that supports those files, due to the libfuse3 dependency.

tych0:
> Why do we need/want the online file?

The jira issue has more details, but roughly: older glibcs read this to make memory arena decisions in the allocator and wrong values are causing issues.

tych0:
> As long as these are not more misleading compared to if we didn't expose them at all.

They are all exposed, just with the host values. We're not adding more info here, just trying to display stuff relevant to the container.

solarkennedy (approved):

nickbp (approved):

executor/runtime/docker/lxcfs.go
tych0: Finally getting back to this:

> PS: could alphabetize the list - but the above lxcfs doc doesn't fully alphabetize theirs either

I went ahead and put them in the order that the lxcfs readme has them, so that hopefully it's easier to compare when someone in the future is curious.

PR #968, opened 2022-09-16
WIP

solarkennedy:
### Description of the Change

### DIFF
--- a/cmd/titus-executor-backend/main.go
+++ b/cmd/titus-executor-backend/main.go
@@ -30,7 +30,6 @@ import (
 
 type commandConfig struct {
 	// runtime dir will have pod.json
-	// and be populated with state.json
 	// All other file names are meant to be ignored.
 	runtimeDir string
 	journald   bool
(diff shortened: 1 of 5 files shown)


solarkennedy:
Actually, sharing the whole run dir may make other things simpler.
In fact, we could put the tini sockets there too? We kinda have a lot of directory bind mounting sprawl.

tych0:
Agreed. I'll make the change. Maybe we can do tini sockets there too, I'll look into that as well in a separate PR.

tych0:
So this works, but I don't see "Started" in the status where I would expect it when i run `sleep 10 && cat /titus/run/pod.json` as the container:
```
	"status": {
		"phase": "Pending",
		"conditions": [
			{
				"type": "PodScheduled",
				"status": "True",
				"lastProbeTime": null,
				"lastTransitionTime": "2022-11-15T19:09:01Z"
			}
		],
		"message": "starting 1 container(s) in order: [main]",
		"reason": "TASK_STARTING",
		"containerStatuses": [
			{
				"name": "main",
				"state": {
					"waiting": {}
				},
				"lastState": {},
				"ready": true,
				"restartCount": 0,
				"image": "registry.us-east-1.streamingtest.titus.netflix.net:7002/baseos/bionic@sha256:7f277b879df5cd01e77784ec372bda2d8b2dadb420c392020fd4f2a7d7007ffc",
				"imageID": "",
				"containerID": "413784cc-651e-464e-9be4-80f3a6b5335f"
			}
		],
		"qosClass": "Guaranteed"
	}
```

solarkennedy:
While cool, we decided to do this in 1 passes.

1. @tych0 is going to make a PR to mount run/ instead of run/pod.json, and verify that vk continuously updates that file.
Once done, tasks will get updated pod.json via that file, mission accomplished there

2. (a month from now) we'll come up with an alternative way for executor to send updates to vk, and come up with a deprecation plan of the state.json communication method.

PR #972, opened 2022-09-21
docker: use only one dbus connection

tych0:
I don't know that this is really as big of a win as the titus-virtual-kubelet one (in fact, it might be a loss since for the most part we don't use this connection, just once at container start and once at container stop). Here it is anyway for science. IMO, we should not merge this.

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -33,6 +33,7 @@ import (
 	vpcapi "github.com/Netflix/titus-executor/vpc/api"
 	"github.com/Netflix/titus-executor/vpc/tracehelpers"
 	"github.com/Netflix/titus-kube-common/pod"
+	"github.com/coreos/go-systemd/v22/dbus"
 	"github.com/docker/distribution/reference"
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/container"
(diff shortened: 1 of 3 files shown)


solarkennedy (approved):

solarkennedy:
Are you sure this is a loss? This makes us use 1 connection to dbus instead of N? (N=number of titus system services)

executor/runtime/docker/docker.go
```diff
@@ -2486,7 +2495,9 @@ func (r *DockerRuntime) setupPostStartNetworkingAndIsolate(parentCtx context.Con
 	r.registerRuntimeCleanup(func() error {
 		ctx, cancel := context.WithCancel(context.Background())
 		defer cancel()
-		if err := stopSystemServices(ctx, c); err != nil {
+		err := stopSystemServices(ctx, r.dbusConn, c)
+		r.dbusConn.Close()
```
nickbp, line 2499: Is the dbus connection held open for the lifetime of the container when it wasn't before?

tych0: Yes, that's what I was trying to say in the commit message, but didn't say it as clearly. That's why I don't think we should merge this -- there's really no need to have this connection open.

tych0:
> Are you sure this is a loss? This makes us use 1 connection to dbus instead of N? (N=number of titus system services)

No, we already did use one connection to start the N services:

https://github.com/Netflix/titus-executor/blob/master/executor/runtime/docker/docker_linux.go#L111

this just makes it so we persist that connection for the container's lifetime and disconnect after shutting the services down.

PR #973, opened 2022-09-23
tini check for linux extra version before setting seccomp wait killable flag

aloktiagi:


MyUmmaGumma (approved):

tini/src/seccomp_fd_notify.c
```diff
@@ -31,6 +31,7 @@
 #endif
 
 #define WAIT_KILLABLE_SECCOMP_KERNEL_VERSION KERNEL_VERSION(5, 15, 35)
+#define WAIT_KILLABLE_SECCOMP_EXTRA_KERNEL_VERSION "netflix-g54efd87a8576"
```
solarkennedy, line 34: This is not super sustainable. As soon as we do an upgrade, which will still have all the seccomp goodies, won't it break?

Don't you really need a blacklist of 5.15.35 kernels that don't support this feature?

solarkennedy (approved):

tini/src/seccomp_fd_notify.c
tych0: I agree. I don't think these kinds of version checks are really that reliable anyway, since people can backport, etc. I'd do it like the kernel test suite or libseccomp does it, and just check for EINVAL from the actual syscall: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/tools/testing/selftests/seccomp/seccomp_bpf.c#n2377

aloktiagi: Yeah I was checking if the kernel version was greater than 5.15.35 or equal and extra version was netflix-g54efd87a8576, but yeah as @tych0 said someone could backport and we can have a new version of 5.15.35 and this check will break. I'll add a check to see what flags are supported as @tych0 suggested.

aloktiagi: @tych0 @solarkennedy do the new changes look ok?

tini/src/seccomp_fd_notify.c
```diff
@@ -277,12 +256,14 @@ static int install_notify_filter(void) {
 	/* Only one listening file descriptor can be established. An attempt to
 	   establish a second listener yields an EBUSY error. */
 
-	if (get_kernel_version(&system_kernel_version) == -1) {
-		return -1;
-	}
+	for (unsigned int i = 0; i < ARRAY_SIZE(flags); i++) {
+		seccomp(SECCOMP_SET_MODE_FILTER, flags[i], NULL);
```
tych0, line 260: seems a little weird not to check the result of this, but I think it's probably ok since if this behavior changes that means a kernel ABI break, which the kernel's tests will check for.

tych0 (approved):

tini/src/seccomp_fd_notify.c
aloktiagi: I agree. I wanted to check the return value as well, but the assumption is the seccomp call will return a -1 always if the arg field is NULL.

tych0: > I wanted to check the return value as well, but the assumption is the seccomp call will return a -1 always if the arg field is NULL.

Sure, I guess I'm just saying i'd feel safer if we "trust-but-verified" :)

PR #974, opened 2022-09-23
Increase timeouts to vpc tool assign/unassign

MyUmmaGumma:
Align vpctool call timeouts from docker to what vpctool uses internally

https://github.com/Netflix/titus-executor/blob/master/vpc/tool/assign3/assign_network.go#L113
https://github.com/Netflix/titus-executor/blob/master/vpc/tool/assign3/unassign.go#L18

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ GOBIN                 ?= $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))
 PATH                  := $(PATH):$(GOBIN)
 GOBIN_TOOL            = $(shell which gobin || echo $(GOBIN)/gobin)
 GOIMPORT_TOOL		  = $(GOBIN_TOOL) -m -run golang.org/x/tools/cmd/goimports@v0.1.0 -w
-GOLANGCI_LINT_TIMEOUT := 2m
+GOLANGCI_LINT_TIMEOUT := 6m
 ifdef FAST
 	GOLANGCI_LINT_ARGS = --fast
 endif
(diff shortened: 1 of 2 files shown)


executor/runtime/docker/networking.go
```diff
@@ -27,13 +27,14 @@ import (
 
 const (
 	defaultNetworkBandwidthBps = 128 * MB
-	vpctoolTimeout             = 45 * time.Second
+	assignVpctoolTimeout       = 120 * time.Second
```
nickbp, line 30: Given the tools are also in titus-executor, would it be straightforward to export the values and reuse them here and in the tools? It'd at least make the sharing more explicit.

If that's a pain then worst case could add comments here pointing to the other files.

MyUmmaGumma: Done.

Makefile
```diff
@@ -17,7 +17,7 @@ GOBIN                 ?= $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))
 PATH                  := $(PATH):$(GOBIN)
 GOBIN_TOOL            = $(shell which gobin || echo $(GOBIN)/gobin)
 GOIMPORT_TOOL		  = $(GOBIN_TOOL) -m -run golang.org/x/tools/cmd/goimports@v0.1.0 -w
-GOLANGCI_LINT_TIMEOUT := 2m
+GOLANGCI_LINT_TIMEOUT := 6m
```
hechaoli, line 20: Why is this also changed?

hechaoli (approved):

Makefile
MyUmmaGumma: thanks, good catch - my local build was failing without increasing this timeout! - reverting.

fabiokung (approved):

PR #976, opened 2022-10-06
HighScale now defaults to using TSA

MyUmmaGumma:
HighScale network mode now uses TSA by default for better adoption


executor/runtime/types/types.go
```diff
@@ -382,7 +382,10 @@ func computeEffectiveNetworkMode(originalNetworkMode string, assignIPv6Address b
 		return titus.NetworkConfiguration_Ipv4Only.String()
 	}
 	if originalNetworkMode == titus.NetworkConfiguration_HighScale.String() {
-		return titus.NetworkConfiguration_Ipv6Only.String()
+		if !enableTransitionNetwork {
```
solarkennedy, line 385: Can you make the logic positive instead of negative?

if tsa {
  return fallback
}
return dualstack

solarkennedy (approved):

PR #977, opened 2022-10-11
Always provide a Cmd (`CMD`) for an extra container

solarkennedy:
Normal kubelet will not respect the `CMD` specified in a docker image if the entryPoint is specified in the pod (k8s `Command`).

I think that is wrong, and not consistent with us using the entrypoint of an image if provided.

Additionally this is not consistent with docker run.

This PR makes it so we use the `CMD` from an image if the pod doesn't have a cmd (k8s `Arguments`) set.

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -2144,10 +2144,10 @@ func (r *DockerRuntime) k8sContainerToDockerConfigs(c *runtimeTypes.ExtraContain
 		// Hostname must be empty here because setting the hostname is incompatible with
 		// a container:foo network mode
 		Hostname:    "",
-		Cmd:         v1Container.Args,
 		Image:       v1Container.Image,
 		WorkingDir:  v1Container.WorkingDir,
 		Entrypoint:  computeExtraContainersDockerEntrypoint(c),
+		Cmd:         computeExtraContainersDockerCommand(c),
 		Labels:      labels,
 		Env:         append(baseEnv, v1ConatinerEnvToList(v1Container.Env)...),
 		Healthcheck: healthcheck,
@@ -2216,6 +2216,29 @@ func getExtraContainerEntrypoint(c *runtimeTypes.ExtraContainer) []string {
 	}
 	// Otherwise we provide whatever the original image had, even if it is an empty array
 	return c.ImageInspect.ContainerConfig.Entrypoint
+}
+
+func computeExtraContainersDockerCommand(c *runtimeTypes.ExtraContainer) []string {
+	if len(c.V1Container.Args) > 0 {
+		// If explicitly set k8s Args (docker CMD) on the pod, we need to return that, no question.
+		// This will be used as the CMD for the ENTRYPOINT already configured (which come from either the pod or the image)
+		return c.V1Container.Args
+	}
+
+	if len(c.V1Container.Command) > 0 {
+		// If there was a k8s Command (docker ENTRYPOINT) specified on the pod, then it would be inappropriate
+		// to use the k8s args (docker CMD) that came with the container, because they would have been originally
+		// written for a different docker ENTRYPOINT!
+		// This mimics the k8s and docker behavior documented here: https://docs.docker.com/engine/reference/run/#entrypoint-default-command-to-execute-at-runtime
+		return []string{}
+	}
+
+	// And finally, we need to explicitly set the CMD to what the image had.
+	// Normally with docker we wouldn't have to explicitly do this.
+	// But with titus-executor extra containers, we *always* override the docker ENTRYPOINT to inject tini,
+	// which would normally make docker ignore the docker CMD in the image.
+	// Therefore, we need to explicitly set the CMD from whatever was in the original image, to explicitly configure docker to use it.
+	return c.ImageInspect.ContainerConfig.Cmd
 }
 
 func v1ConatinerEnvToList(v1Env []v1.EnvVar) []string {


aaronlehmann:
Ah, this is one of the misfeatures of `docker run` that the kubelet is probably trying to correct. So much confusion has been caused by `entrypoint` and `cmd` combinding in this way.

I think it's okay for us to be compatible with `docker run`.

solarkennedy:
I think I take it back, this is not what docker run does

https://docs.docker.com/engine/reference/run/#entrypoint-default-command-to-execute-at-runtime

> Passing --entrypoint will clear out any default command set on the image (i.e. any CMD instruction in the Dockerfile used to build it).

tych0:
Yeah, this is always super confusing. It does seem reasonable to use the original image config if people didn't set things up in the k8s blessed way, though. Maybe,
```
if len(c.ImageInspect.ContainerConfig.Entrypoint) > 0 {
    return c.ImageInspect.ContainerConfig.Entrypoint
}
return c.ImageInspect.ContainerConfig.Cmd
```
like what (my interpretation of) your link says docker does?

solarkennedy:
@aaronlehmann can I have another review of this one?
This feels like one of those PRs we are making a behavior decision, and it will be something we will not be able to take back. (probably implies we should add more tests for it?)

executor/runtime/docker/docker.go
```diff
@@ -2218,6 +2218,29 @@ func getExtraContainerEntrypoint(c *runtimeTypes.ExtraContainer) []string {
...
 
+func computeExtraContainersDockerCommand(c *runtimeTypes.ExtraContainer) []string {
+	if len(c.V1Container.Args) > 0 {
+		// If explicitly set k8s Args (docker CMD) on the pod, we need to return that, no question.
+		// This will be used as the CMD for the ENTRYPOINT already configured (which come from either the pod or the image)
+		return c.V1Container.Args
+	}
+
+	if len(c.V1Container.Command) > 0 {
+		// If there was a k8s Command (docker ENTRYPOINT) specified on the pod, then it would be inappropriate
```
aaronlehmann, line 2229: It's so confusing that `Command` refers to entrypoint rather than the args (`CMD`) here :(.

executor/runtime/docker/docker.go
```diff
@@ -2218,6 +2218,29 @@ func getExtraContainerEntrypoint(c *runtimeTypes.ExtraContainer) []string {
...
+		// to use the k8s args (docker CMD) that came with the container, because they would have been originally
+		// written for a different docker ENTRYPOINT!
+		// This mimics the k8s and docker behavior documented here: https://docs.docker.com/engine/reference/run/#entrypoint-default-command-to-execute-at-runtime
```
aaronlehmann, line 2232: Thanks for fixing this comment - it is much clearer and clarifies that we are bringing the behavior into alignment with the stock kubelet and `docker run`, not deviating from their behavior.

executor/runtime/docker/docker.go
```diff
@@ -2218,6 +2218,29 @@ func getExtraContainerEntrypoint(c *runtimeTypes.ExtraContainer) []string {
...
+		return []string{}
+	}
+
+	// And finally, we need to explicitly extract the image had.
```
aaronlehmann, line 2236: This sentence ends abruptly.

aaronlehmann (commented):
The `CMD`/`ENTRYPOINT` situation is always very confusing, but here's what I understand so far:

- `docker run` has a behavior where specifying `--entrypoint` clears anything specified in the image's `CMD` (because those would be interpreted as args, and they might not make sense for the new entrypoint).

- The stock kubelet has analagous behavior when you specify an entrypoint in the pod, clearing any `CMD` provided by the image.

- This PR aligns the `titus-executor` behavior with both `docker run` and the stock kubelet. Previously, args from the image were ignored. Now, they will be used in the same situations that the stock kubelet uses them.

This seems like a positive change to me because it brings our behavior in-line with other tools (and also, using the args from the container image makes sense in this context).

slushie (approved):
This implementation looks good to me, thanks! 

For cases when neither Command or Args /  Entrypoint or Cmd are set in the image or in the pod config, the current behaviour is an error response about `tini` not starting, which is pretty confusing. I think it would be helpful to provide something along the lines of: 
```
⫎  docker run --rm -ti --entrypoint="" dockerregistry.test.netflix.net:7002/baseos/bionic
docker: Error response from daemon: No command specified.
```

solarkennedy:
> This implementation looks good to me, thanks!
> 
> For cases when neither Command or Args / Entrypoint or Cmd are set in the image or in the pod config, the current behaviour is an error response about `tini` not starting, which is pretty confusing. I think it would be helpful to provide something along the lines of:
> 
> ```
> ⫎  docker run --rm -ti --entrypoint="" dockerregistry.test.netflix.net:7002/baseos/bionic
> docker: Error response from daemon: No command specified.
> ```

That sounds like a nice improvement. I'll add that on the next PR.

executor/runtime/docker/docker.go
```diff
@@ -2212,6 +2212,26 @@ func getExtraContainerEntrypoint(c *runtimeTypes.ExtraContainer) []string {
...
 
+func computeExtraContainersDockerCommand(c *runtimeTypes.ExtraContainer) []string {
+	// What docker calls "command", k8s calls "args".
+	// If explicitly set on the pod, we need to return that.
+	if len(c.V1Container.Args) > 0 {
+		return c.V1Container.Args
+	}
+	// Stock kubelet would stop there, and the Cmd in the image would be ignored if entrypoint
+	// is provided. However, we think that is incorrect and instead think we should be
+	// defaulting to whatever the image had as the original cmd, but *only* if the `Command` (entrypoint)
```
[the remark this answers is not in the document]
solarkennedy: I was looking around for others complaining about this confusion. Surprisingly I found not many.

One said (can't find anymore) that Docker calls them what they "are" where k8s calls them based on how they are "used". I thought that was an interesting insight.

PR #978, opened 2022-10-11
Put hsperfdata on a tmpfs for java. JVM-67

solarkennedy:
On EC2, you can `mount` your own tmpfs and use that for java hotspot perf data. In a container that isn't as easy.

This PR makes it so we get tmpfs for the common java unix users for all workloads.

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -449,6 +449,12 @@ func (r *DockerRuntime) mainContainerDockerConfig(c runtimeTypes.Container, bind
 		"/run/netns": "rw,size=" + defaultRunTmpFsSize,
 	}
 
+	// For performance reasons, we put the hsperfdata for JVM-based workloads on a tmpfs
+	// See https://www.evanjones.ca/jvm-mmap-pause.html for an explanation. JVM-67
+	hostCfg.Tmpfs["/tmp/hsperfdata_www-data"] = "rw,size=" + defaultRunTmpFsSize
+	hostCfg.Tmpfs["/tmp/hsperfdata_nfsuper"] = "rw,size=" + defaultRunTmpFsSize
+	hostCfg.Tmpfs["/tmp/hsperfdata_root"] = "rw,size=" + defaultRunTmpFsSize
+
 	if c.IsSystemD() {
 		// systemd requires `/run/lock` to be a separate mount from `/run`
 		hostCfg.Tmpfs["/run/lock"] = "rw,exec,size=" + defaultRunLockTmpFsSize


executor/runtime/docker/docker.go
```diff
@@ -449,6 +449,12 @@ func (r *DockerRuntime) mainContainerDockerConfig(c runtimeTypes.Container, bind
 		"/run/netns": "rw,size=" + defaultRunTmpFsSize,
 	}
 
+	// For performance reasons, we put the hsperfdata for JVM-based workloads on a tmpfs
+	// See https://www.evanjones.ca/jvm-mmap-pause.html for an explanation. JVM-67
+	hostCfg.Tmpfs["/tmp/hsperfdata_www-data"] = "rw,size=" + defaultRunTmpFsSize
```
nickbp, line 454: The blog post mentions that these are paths to files rather than directories, is that okay here in the context of creating mountpoints?

solarkennedy: Waaaat?
I was going based off of https://stash.corp.netflix.com/projects/BASEAMI/repos/nflx-ezconfig-core/browse/root/apps/ezconfig/templates/apps/ezconfig/gen/java_run.sh#67-81

cc @danielthomas

nickbp: OK it looks like the blog post is inaccurate because I checked `/tmp` and it's a directory:
```
$ ls -l /tmp
drwxr-xr-x 2 nickparker    100 Oct 12 17:55 hsperfdata_nickparker
```

So nevermind, good to go IMO unless @DanielThomas has anything

nickbp (approved):

executor/runtime/docker/docker.go
DanielThomas: Yeah, it's a directory. Each process gets a `<pid>` file in there.

tych0:
> On EC2, you can mount your own tmpfs and use that for java hotspot perf data. In a container that isn't as easy.

Why isn't it that easy? Can we make it that easy? Encoding workload specific mounts into the executor seems not ideal to me.

solarkennedy:
> > On EC2, you can mount your own tmpfs and use that for java hotspot perf data. In a container that isn't as easy.
> 
> Why isn't it that easy? Can we make it that easy? Encoding workload specific mounts into the executor seems not ideal to me.

I guess we could by relaxing the `mount` part of the seccomp policies?

tych0:
> I guess we could by relaxing the mount part of the seccomp policies?

People can already mount a tmpfs using the `docker_fuse` profile, so it's not giving them anything they couldn't already do with some extra work. tmpfs mounts are accounted for in the memory cgroup, so it Seems Safe To Me. I can re-work the profiles if we want, or we can leave this in here.

solarkennedy:
> > I guess we could by relaxing the mount part of the seccomp policies?
> 
> People can already mount a tmpfs using the `docker_fuse` profile, so it's not giving them anything they couldn't already do with some extra work. tmpfs mounts are accounted for in the memory cgroup, so it Seems Safe To Me. I can re-work the profiles if we want, or we can leave this in here.

I don't think that is quite the right way to think about it. 
Sure, in theory every java user at netflix could enable fuse, but that would make us less secure *and* require every user to change.

tych0:
>  Sure, in theory every java user at netflix could enable fuse, but that would make us less secure and require every user to change.

I'm not suggesting they enable fuse, I'm suggesting we let people do tmpfs mounts in the "normal" profile.

solarkennedy:
> > Sure, in theory every java user at netflix could enable fuse, but that would make us less secure and require every user to change.
> 
> I'm not suggesting they enable fuse, I'm suggesting we let people do tmpfs mounts in the "normal" profile.

Ahhh. Yea if you think that can be safely done, shipit!

PR #981, opened 2022-10-18
docker: relax default capabilities to allow tmpfs mounts

tych0:
We have to do three things here:

1. give everyone CAP_SYS_ADMIN (this is probably the scariest)
2. allow `mount` in our seccomp profile
3. drop the `deny mount` from apparmor. since apparmor denies-by-default, we still can't mount other things that you generally can mount in a namespace, given that we only specify tmpfs:

     + mkdir -p /tmp/foo /tmp/bar
     + mount -t tmpfs -o size=128M tmpfs /tmp/foo
     + mount -t sysfs sysfs /tmp/bar
     mount: /tmp/bar: cannot mount sysfs read-only.

   so this is less scary than it looks.

This also updates the default docker seccomp profile with a few new things like CAP_BPF -> bpf syscall and CAP_PERFMON -> perf_event_open.

### Description of the Change


solarkennedy:
Do we really need 1?
I thought we only needed 1 for non-user-namespaced filesystemd (and tmpfs is user-namespaced)

tych0:
>  I thought we only needed 1 for non-user-namespaced filesystemd (and tmpfs is user-namespaced)

The kernel's check looks like this: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/namespace.c?h=v5.15#n1711 (callsite here: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/namespace.c?h=v5.15#n3263)

You need to have CAP_SYS_ADMIN in the userns that owns your mountns. So we could allow users to unshare() and then they could set it all up as well, but if they don't have it in the userns docker creates for them, they can't do it themselves.

You're probably thinking of the FS_USERNS_MOUNT check, which happens later on in the code path, and is set for tmpfs and a few others:

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/namespace.c?h=v5.15#n2985
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/super.c?h=v5.15#n486

That was part of the reason I wrote this patch, because I had forgotten about CAP_SYS_ADMIN as well and thought it was just apparmor+seccomp :(

solarkennedy:
In that case, no, I think this is too much and that we should find some other way.

tych0:
It sounds like maybe they're just happy to wait for a location param for the JVM here, so we can just punt on this entirely for now... in any case I'll close this.

PR #982, opened 2022-10-19
Titus 6247

tych0:
output of version information and a couple of other fixes

### DIFF
--- a/executor/runtime/types/types.go
+++ b/executor/runtime/types/types.go
@@ -3,12 +3,14 @@ package types
 import (
 	"context"
 	"fmt"
+	"os"
 	"regexp"
 	"strings"
 	"sync"
 	"time"
 
 	"github.com/Netflix/titus-executor/config"
+	"github.com/Netflix/titus-executor/executor"
 	"github.com/Netflix/titus-executor/uploader"
 	vpcapi "github.com/Netflix/titus-executor/vpc/api"
 	podCommon "github.com/Netflix/titus-kube-common/pod"
@@ -498,6 +500,20 @@ func (n *NetworkConfigurationDetails) PickPrimaryIP() string {
 // not returned by normal container start calls.
 type Details struct {
 	NetworkConfiguration *NetworkConfigurationDetails
+}
+
+func getKernelVersion() string {
+	kernelVersionRaw, err := os.ReadFile("/proc/sys/kernel/osrelease")
+	if err != nil {
+		return fmt.Sprintf("problem reading kernel version: %v", err)
+	}
+
+	return strings.TrimSpace(string(kernelVersionRaw))
+
+}
+
+func (d Details) RenderVersionDetails() string {
+	return fmt.Sprintf("kernel-version=%s,titus-executor=%s", getKernelVersion(), executor.TitusExecutorVersion)
 }
 
 type ContainerRuntimeProvider func(ctx context.Context, c Container, startTime time.Time) (Runtime, error)
--- a/executor/version.go
+++ b/executor/version.go
@@ -1,0 +1,5 @@
+package executor
+
+var (
+	TitusExecutorVersion = "ManualBuild"
+)
--- a/vk/backend/backend.go
+++ b/vk/backend/backend.go
@@ -16,6 +16,7 @@ import (
 	"github.com/Netflix/titus-executor/executor/runner"
 	runtimeTypes "github.com/Netflix/titus-executor/executor/runtime/types"
 	"github.com/Netflix/titus-executor/logger"
+	podCommon "github.com/Netflix/titus-kube-common/pod"
 	resourceCommon "github.com/Netflix/titus-kube-common/resource"
 	units "github.com/docker/go-units" // nolint: staticcheck
 	"github.com/google/renameio"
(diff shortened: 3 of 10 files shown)


vk/backend/backend.go
```diff
@@ -146,6 +149,27 @@ func NewBackend(ctx context.Context, rp runtimeTypes.ContainerRuntimeProvider, p
...
+
+	kernelVersion := strings.TrimSpace(string(kernelVersionRaw))
+
+	// this is actually kind of racy... if mutator runs and upgrades the
+	// package during the window between when this binary starts and this
+	// command runs, we'll render the wrong version. we could plumb through
+	// the version via golang's -ldflags -X, but we'll still have this
+	// problem if we start querying other package versions. best to do this
+	// as early as possible, I think.
+	executorVersionRaw, err := exec.Command("dpkg-query", "--show", "--showformat=${Version}", "titus-executor").CombinedOutput()
```
solarkennedy, line 165: Yea don't do this.

Instead, can you do this?

https://levelup.gitconnected.com/a-better-way-than-ldflags-to-add-a-build-version-to-your-go-binaries-2258ce419d2d

way faster, no exec, much simpler at runtime

vk/backend/backend.go
```diff
@@ -146,6 +149,27 @@ func NewBackend(ctx context.Context, rp runtimeTypes.ContainerRuntimeProvider, p
...
+	if err != nil {
+		executorVersionRaw = []byte(fmt.Sprintf("problem reading executor version: %v", err))
+	}
+	executorVersion := strings.TrimSpace(string(executorVersionRaw))
+
+	pod.Annotations[podCommon.AnnotationKeyRuntimeVersions] = fmt.Sprintf("kernel-version=%s,titus-executor=%s", kernelVersion, executorVersion)
```
solarkennedy, line 171: I think this is kinda an awkward place to put this.

Instead can you extend the *runtimeTypes.Details struct and put this csv in there? Say `VersionDetailsCSV` ?

And then find the line `update.Details.NetworkConfiguration.ToAnnotationMap()` and add an analogous line to add in the annotation there in the update

vk/backend/backend.go
```diff
@@ -146,6 +149,27 @@ func NewBackend(ctx context.Context, rp runtimeTypes.ContainerRuntimeProvider, p
 		return nil, errors.New("error converting network resource units")
 	}
 
+	kernelVersionRaw, err := os.ReadFile("/proc/sys/kernel/osrelease")
```
solarkennedy, line 152: Can you make all the logic for computing the value (CSV) for the runtime versions into a function?

vk/backend/backend.go
solarkennedy: It is another YAGNI. Start with ldflags and only get fancier if we know we need it.

tych0:
Ok, I think I've got this refactored to put everything in the right place. IIUC, we'll need a build of titus-executor-builder to be uploaded before the new titus-executor-builder.sh gets put there to pass the right -ldflags -X incantation for real builds to work. Maybe that'll happen automagically when this lands? I'll double check when it merges in any case.

vk/backend/backend.go
```diff
@@ -322,13 +325,18 @@ func (b *Backend) handleUpdate(ctx context.Context, update runner.Update) {
...
 		defer b.podLock.Unlock()
+
+		// These annotations allow us to get data "back out" from the executor up to the control-plane, without having
+		// to depend on the limitations of the PodStatus structure.
 		for k, v := range update.Details.NetworkConfiguration.ToAnnotationMap() {
 			b.pod.Annotations[k] = v
 		}
+
+		if _, ok := b.pod.Annotations[podCommon.AnnotationKeyRuntimeVersions]; !ok {
+			b.pod.Annotations[podCommon.AnnotationKeyRuntimeVersions] = update.Details.RenderVersionDetails()
```
solarkennedy, line 338: You can just force set it

executor/runtime/types/types.go
```diff
@@ -500,6 +503,17 @@ type Details struct {
 	NetworkConfiguration *NetworkConfigurationDetails
 }
 
+func (d Details) RenderVersionDetails() string {
+	kernelVersionRaw, err := os.ReadFile("/proc/sys/kernel/osrelease")
+	if err != nil {
+		kernelVersionRaw = []byte(fmt.Sprintf("problem reading kernel version: %v", err))
+	}
+
+	kernelVersion := strings.TrimSpace(string(kernelVersionRaw))
```
solarkennedy, line 512: Can you make this a function? Is there really no golang function for this today?

executor/runtime/types/types.go
```diff
@@ -51,6 +52,8 @@ const (
 	testImageTag     = "latest"
 )
 
+var Version = ""
```
solarkennedy, line 55: can you make this `titusExecutorVersion` ?

Is there any other place we can put this where it can be more accessible? Where it would be easy to put it in logs or help text or other places? (types.go seems like a strange place to put it. Maybe /executor/version.go ?)

executor/runtime/types/types.go
tych0: I guess there is this library, but it just does the same thing: https://github.com/zcalusic/sysinfo/blob/v0.9.5/kernel.go#L21

executor/runtime/types/types.go
tych0: Sure, I was trying to preemptively YAGNI this, but I can move it there.

solarkennedy: I know, it is a balancing act. I swear we will use this :)

executor/version.go
```diff
@@ -0,0 +1,5 @@
+package executor
+
+var (
+	TitusExecutorVersion = ""
```
solarkennedy, line 4: How about `ManualBuild` so that if it is unset we have that sentinel value? (we know that it didn't go through CI)

solarkennedy (approved):

PR #984, opened 2022-10-21
TITUS-5376: add support for container+sidecar ordering

tych0:
Note that with this patch if any container's health check fails during startup, the container will fail to start, which was not necessarily true before. It will give people a nice error message like:

    Failed to start a user-defined container: container 5217f571-f07e-4813-bd8c-dc4ad74c0096 is unhealthy (last failure 127: /bin/sh: 1: curl: not found )

but it is a change in semantics.

Maybe we *only* want to check the healthchecks if ordering was specified? but it does feel like the health checks should be sane... idk.

Also, we use a stable sort here and take some pains to preserve the ordering from before this patch, just in case someone depended on that ordering but wasn't explicit about it.

This Works For Me, but it is Friday afternoon, so I'll re-test everything Monday just to be sure :)

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -1873,14 +1873,143 @@ func (r *DockerRuntime) startNonMainContainers(ctx context.Context) error {
 	return group.Wait()
 }
 
-// launchAllContainers starts all existing (pre-created) containers, even the 'main' one (via tini)
-func (r *DockerRuntime) launchAllContainers(ctx context.Context, tiniConns map[string]*net.UnixConn) error {
-	l := log.WithField("taskID", r.c.TaskID())
+func inOrderingAnnotation(thePod *v1.Pod, cName string, suffix string, target string) bool {
+	ann, ok := thePod.Annotations[pod.ContainerAnnotation(cName, suffix)]
+	if ok {
+		for _, c := range strings.Split(ann, ",") {
+			if c == target {
+				return true
+			}
+		}
+	}
 
+	return false
+}
+
+func (r *DockerRuntime) orderSortedContainerNames() []string {
 	// First is platform sidecars, then is user-sidecars, then main
 	cNames := r.getPlaformContainerNames()
 	cNames = append(cNames, r.getUserContainerNames()...)
 	cNames = append(cNames, runtimeTypes.MainContainerName)
+
+	origOrder := map[string]int{}
+	for i, cName := range cNames {
+		origOrder[cName] = i
+	}
+
+	thePod, podLock := r.c.Pod()
+	defer podLock.Unlock()
+
+	sort.SliceStable(cNames, func(i, j int) bool {
+		// does i come before j?
+		if inOrderingAnnotation(thePod, cNames[i], pod.AnnotationKeySuffixContainersStartBefore, cNames[j]) {
+			return true
+		}
+
+		// does i come after j?
+		if inOrderingAnnotation(thePod, cNames[i], pod.AnnotationKeySuffixContainersStartAfter, cNames[j]) {
+			return false
+		}
+
+		// does j come before i?
+		if inOrderingAnnotation(thePod, cNames[j], pod.AnnotationKeySuffixContainersStartBefore, cNames[i]) {
+			return false
+		}
+
+		// does j come after i?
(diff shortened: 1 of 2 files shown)


tych0:
Ok, I am reasonably confident this works as intended. Thoughts welcome.

executor/runtime/docker/docker.go
```diff
@@ -1873,15 +1873,128 @@ func (r *DockerRuntime) startNonMainContainers(ctx context.Context) error {
...
+			if count > 0 {
+				// oldest first, so let's pick the last one, i.e. the newest one
+				last := inspectOutput.State.Health.Log[count-1]
+				msg = last.Output
+				exitCode = last.ExitCode
+			}
+			return fmt.Errorf("container %s is unhealthy (last failure %d: %v)", dockerName, exitCode, msg)
+		default:
+			return fmt.Errorf("unknown container health status for %s: %s", dockerName, inspectOutput.State.Health.Status)
+		}
```
solarkennedy, line 1982: This is a little weird, because we *already* have code to parse docker events and put them in the pod status?
Can this be reused? Or is it that these goroutines have not started yet I guess?

executor/runtime/docker/docker.go
```diff
@@ -1873,15 +1873,128 @@ func (r *DockerRuntime) startNonMainContainers(ctx context.Context) error {
...
+				if c.Name == cName {
+					dockerName = c.Status.ContainerID
+				}
+			}
+			for _, c := range r.c.ExtraUserContainers() {
+				if c.Name == cName {
+					dockerName = c.Status.ContainerID
+				}
+			}
+		}
```
solarkennedy, line 1949: Can you make this a containerNameToContainerID function?

executor/runtime/docker/docker.go
```diff
@@ -1873,15 +1873,128 @@ func (r *DockerRuntime) startNonMainContainers(ctx context.Context) error {
...
+
+	return cNames
+}
+
+func (r *DockerRuntime) waitForHealthy(ctx context.Context, cName string) error {
+	// XXX: should max tries here? seems like relying on context is enough...
+	for {
+		dockerName := cName
+		if dockerName == runtimeTypes.MainContainerName {
+			dockerName = r.c.TaskID()
```
solarkennedy, line 1937: just use container ID consistently here. r.c.ID() is it

See `getContainerNameFromID`

executor/runtime/docker/docker.go
```diff
@@ -1873,15 +1873,128 @@ func (r *DockerRuntime) startNonMainContainers(ctx context.Context) error {
...
+			count := len(inspectOutput.State.Health.Log)
+			if count > 0 {
+				// oldest first, so let's pick the last one, i.e. the newest one
+				last := inspectOutput.State.Health.Log[count-1]
+				msg = last.Output
+				exitCode = last.ExitCode
+			}
+			return fmt.Errorf("container %s is unhealthy (last failure %d: %v)", dockerName, exitCode, msg)
+		default:
+			return fmt.Errorf("unknown container health status for %s: %s", dockerName, inspectOutput.State.Health.Status)
```
solarkennedy, line 1981: Can you make this a function that takes an inspect and returns a Error?

executor/runtime/docker/docker.go
```diff
@@ -1873,15 +1873,128 @@ func (r *DockerRuntime) startNonMainContainers(ctx context.Context) error {
...
+			if c == target {
+				return true
+			}
+		}
+	}
+
+	return false
+}
 
+func (r *DockerRuntime) determineContainerOrdering() []string {
```
solarkennedy, line 1889: how about `orderSortedContainerNames` ?

solarkennedy:
I think this all looks pretty sane

executor/runtime/docker/docker.go
tych0: You're talking about the stuff in `statusMonitor`? It looks like we have started those at this point. Not a ton of docs on those events: https://pkg.go.dev/github.com/docker/docker/api/types/events but it might be possible based on https://docs.docker.com/engine/reference/commandline/events/ let me see.

tych0: Doesn't look like it, we only get `start`, `die`, and `destroy`; I messed around with the [filters](https://pkg.go.dev/github.com/docker/docker/api/types/filters#Args.Add) and couldn't get anything to pop out either :(

PR #986, opened 2022-11-02
Don't panic when injecting imds on errors

solarkennedy:
I don't like to see panics when the error is "normal". Granted, I don't like the bug, but it doesn't indicate an actual bug in the code (probably).

This also adds a timer so we can actually see how long we ended up waiting. Maybe we are not waiting long enough!

### DIFF
--- a/cmd/titus-inject-metadataproxy/main.go
+++ b/cmd/titus-inject-metadataproxy/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"fmt"
+	"os"
 
 	"github.com/Netflix/titus-executor/cmd/common"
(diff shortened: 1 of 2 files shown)


cmd/titus-inject-metadataproxy/main.go
```diff
@@ -50,14 +51,17 @@ func main() {
 	rootCmd.PersistentFlags().String("log-level", "info", "Log Level")
 
 	if err := v.BindPFlags(rootCmd.PersistentFlags()); err != nil {
-		panic(err)
+		fmt.Fprintln(os.Stderr, err)
```
tych0, line 54: This all looks good, looks like there's one more call to `BindPFlags` above this that could just `return err` too?

PR #987, opened 2022-11-03
Validate the liveness probe

solarkennedy:


executor/runtime/docker/docker.go
```diff
@@ -2217,13 +2220,25 @@ func v1ConatinerEnvToList(v1Env []v1.EnvVar) []string {
...
+		return nil, fmt.Errorf("HTTPGet probes are not supported")
+	}
+	if probe.Exec == nil {
+		return nil, fmt.Errorf("Exec are currently the only kind supported")
+	}
+	if len(probe.Exec.Command) == 0 {
+		return nil, fmt.Errorf("Exec probes must have a command defined")
+	}
+	if !strings.HasPrefix(probe.Exec.Command[0], "/") {
+		return nil, fmt.Errorf("Exec command %v must start with a '/'", probe.Exec.Command)
```
aaronlehmann, line 2240: Huh, why? The example at https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-liveness-command just uses `cat`.

tych0: I don't think there's a shell in the middle to do $PATH resolution.

tych0 (approved):

aaronlehmann (approved):

PR #991, opened 2022-11-05
Don't start any Titus system services in standalone mode

solarkennedy:
Eventually we will run them all, but for now we are not ready to run any!

This PR turns them all off in standalone mode until we are ready to turn them on, one by one.

### DIFF
--- a/executor/runtime/types/sidecars.go
+++ b/executor/runtime/types/sidecars.go
@@ -25,6 +25,7 @@ const (
 )
 
 var systemServices = []ServiceOpts{
+
 	/*
 	 * Titus sidecar seccomp agent must be in the beginning of the list because
 	 * other sidecars may depend on it for IPv6 connectivity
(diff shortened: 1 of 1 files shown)


tych0:
I guess the plan is to enable them one by one? If not, might be simpler to do it here: https://github.com/Netflix/titus-executor/blob/master/executor/runtime/docker/docker.go#L872 ?

tych0 (approved):

solarkennedy:
> I guess the plan is to enable them one by one? If not, might be simpler to do it here: https://github.com/Netflix/titus-executor/blob/master/executor/runtime/docker/docker.go#L872 ?

Yea, my plan is to turn them on one-by-one (read the PR description).
This is because some system services will provide value to EC2, where others won't make any sense at all.

tych0:
> (read the PR description)

heh, derp.

> some system services will provide value to EC2, where others won't make any sense at all

Makes sense, thanks.

PR #992, opened 2022-11-07
End-to-end tests for AssignIPV3

jromero-nflx:
Adding end-to-end tests for the AssignIPv3 API call


hechaoli:
This is awesome! Thank you for adding the E2E test for the most critical `AssignIPv3` API.

Going forward, I'd suggest breaking down the big commit into multiple smaller commits for better reviewing. For instance, I think the "End-to-end tests for AssignIPV3” commits can be broken down to the following commits
* Add `request.Option` param to `CreateNetworkInterface`.
* Rename `db_test` to `dbutil` and refactor `provision_instance_test`.
* DB seeds.
* The actual AssignIPv3 test.

hechaoli (approved):

PR #993, opened 2022-11-08
Use a script to generate the vpc limits information

solarkennedy:
This change needs to be looked at carefully, because it determines the network spec of a Titus agent at startup.

Previously, we were adding in the vpc limits for an instance type by hand, because some of the information (branch enis) were not public.

Now, the data is public, and ec2 has the describeInstances api.

This PR adds a python script to programmatically generate the vpc limits `.go` file based on generated data.

However! Even with this data, there are undocumented outliers, which worry me.

In the end, hopefully the instance_limits.go file is more accurate than it was before.
It certainly supports more (all) instance types!

----

Each commit can be looked at individually for review (recommended).


solarkennedy:
Here is probably the most important diff: The difference between configs with the limits we had in place before:
https://www.diffchecker.com/r5lTkOBc

Looking at just this diff, I think we are fine. If anything this change is more accurate given the current state of branch enis.

vpc/generate.py
```diff
@@ -0,0 +1,458 @@
...
+        "p4d.24xlarge": 120,
+        "p4de.24xlarge": 120,
+        "vt1.3xlarge": 40,
+        "vt1.6xlarge": 60,
+        "vt1.24xlarge": 120,
+    }
+
+    # The page says these instance types are not supported for branch ENIs,
+    # yet we use them in practice?
+    undocumented = {
```
fabiokung, line 371: 😬

vpc/Makefile
```diff
@@ -0,0 +1,17 @@
+refresh: creds fetch generate
+
+creds:
+	newt --app-type awscreds refresh -r arn:aws:iam::686589254418:role/titus_us_east_1_agents_test_titus titus_us_east_1_agents_test
```
fabiokung, line 4: minor: change to `weep file arn:aws:iam::686589254418:role/titus_us_east_1_agents_test_titus -f -p titus_us_east_1_agents_test`

vpc/generate.py
```diff
@@ -0,0 +1,458 @@
...
+        "m5.metal": 198,
+        "r5d.metal": 198,
+    }
+
+    # Merging of both dictionaries of data
+    branch_limits = {**documented, **undocumented}
+    return branch_limits.get(i, 0)
+
+
+def parse_network_info(i):
```
fabiokung, line 382: link to where `Low -> High` values came from?

fabiokung (approved):
👏

PR #995, opened 2022-11-09
Print a friendly error message if entry/cmd is unset

solarkennedy:
For extra containers, it can be easy to not set a command or entrypoint. When this happens, we end up with a weird error because we set the entrypoint to `tini --`.

This PR makes us fail faster if either entrypoint or command is unset.

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -2140,14 +2140,20 @@ func (r *DockerRuntime) k8sContainerToDockerConfigs(c *runtimeTypes.ExtraContain
 	if err != nil {
 		return nil, nil, nil, fmt.Errorf("Parsing the liveness probe for the %s container: %w", v1Container.Name, err)
 	}
+	entrypoint := computeExtraContainersDockerEntrypoint(c)
+	command := computeExtraContainersDockerCommand(c)
+	if len(entrypoint) == 0 && len(command) == 0 {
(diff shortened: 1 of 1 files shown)


executor/runtime/docker/docker.go
```diff
@@ -2140,14 +2140,20 @@ func (r *DockerRuntime) k8sContainerToDockerConfigs(c *runtimeTypes.ExtraContain
 	if err != nil {
 		return nil, nil, nil, fmt.Errorf("Parsing the liveness probe for the %s container: %w", v1Container.Name, err)
 	}
+	entrypoint := computeExtraContainersDockerEntrypoint(c)
+	command := computeExtraContainersDockerCommand(c)
+	if len(entrypoint) == 0 && len(command) == 0 {
+		err := fmt.Errorf("An Entrypoint or Command must be specified for the container")
```
slushie, line 2146: Nice, thank you for this! Would this error show up in the UI next to the specific container which failed, or should it container the name of the container with missing data?

aaronlehmann, line 2146: nit: why not just return `fmt.Errorf(...)` directly?

aaronlehmann (approved):

executor/runtime/docker/docker.go
solarkennedy: The caller will wrap this and name the container that failed to create

solarkennedy: eh, just shorter lines

PR #998, opened 2022-11-17
RFC: Added emptyDir tmpfs support

solarkennedy:
I don't thinks this is "right".

It is simple though: we initialize ram disks at container launch time.

But it doesn't really match what k8s does: where each volume is mounted by the individual containers.

I think to do it "right", I'll need to make a ram disk for each volume and inject the mount (via titus-storage).


solarkennedy:
It looks like we already have decent prior art (that I wrote and forgot!) with shared metatron tmpdirs for pods.
That is more akin to what I'm trying to accomplish here with emptydir, so I'll try that approach.

I also need to verify that if you make a ramdisk on the host and then inject it... that the ram usage comes from container?

PR #1002, opened 2022-11-17
docker: bind mount entire runtime dir vs. just pod

tych0:
This will save us from having to do fancy tricks to keep the pod.json visible in the container, and might be useful for shipping other things to the container (e.g. tini sockets)

We have to manually add a chown because the file is root owned, and we want users to be able to read it.

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -175,7 +175,7 @@ func NewDockerRuntime(ctx context.Context, m metrics.Reporter, dockerCfg Config,
 	}
 
 	defaultBindMounts := []string{}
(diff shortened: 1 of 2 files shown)


solarkennedy (approved):

solarkennedy:
The only thing I could think of is the fact that systemd will try to delete this run dir, but it might be still mounted by containers if they haven't shut down yet? *could* be a side effect of this change?

tych0:
The mount should only exist inside the mountns of the container. If the containers haven't shut down, the systemd unit won't be exiting in theory. But even if it does, I think it'll just show up with the ` (deleted)` suffix in `/proc/self/mountinfo`.

Now, if systemd manages to successfully do an `rm -rf`, that will remove the pod.json so the (zombie) container won't be able to see it any more. But things are already in a pretty weird state at that point.

PR #1003, opened 2022-11-29
Added first pass at titus-mount-lustre

solarkennedy:
This code does work!
It borrows heavily from titus-mount-nfs, which is a different network filesystem.

However, the nuances about mount `src` vs `device` are weird and different.

Another thing I was unable to solve was how to mount this with `MS_NOATIME`. No matter what I did I got EINVAL when trying to mount with that flag.

### DIFF
--- a/cmd/titus-storage/lustre.go
+++ b/cmd/titus-storage/lustre.go
@@ -64,9 +64,9 @@ func setupLustreMount(parentCtx context.Context, taskID, cname string, vol *core
 	fsxFileSystemID := vol.FlexVolume.Options["fsxFileSystemId"]
 	lustreHost := fmt.Sprintf("%s.fsx.%s.amazonaws.com", fsxFileSystemID, region)
 	mountOptions := "recovery_time_soft=300,recovery_time_hard=900"
-	flags := 0
+	flags := MOUNT_ATTR_NOATIME
 	if vm.ReadOnly {
-		flags = flags | MS_RDONLY
+		flags = flags | MOUNT_ATTR_RDONLY
 	}
 
 	cmd := exec.CommandContext(ctx, lustreMountCmd)
--- a/cmd/titus-storage/lustre.go
+++ b/cmd/titus-storage/lustre.go
@@ -6,9 +6,9 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
-	"strings"
 
 	executorDocker "github.com/Netflix/titus-executor/executor/runtime/docker"
+	"github.com/Netflix/titus-executor/logger"
 	corev1 "k8s.io/api/core/v1"
 )
 
@@ -31,7 +31,7 @@ func lustreStart(ctx context.Context, config MountConfig) error {
 	pod := config.pod
 	nameToMount := map[string]corev1.Volume{}
 	for _, v := range pod.Spec.Volumes {
-		if v.FlexVolume != nil && v.FlexVolume.Driver == "LustreVolumeSource" && v.FlexVolume.Options != nil {
+		if v.FlexVolume != nil && v.FlexVolume.Driver == "FSxLustreVolumeSource" && v.FlexVolume.Options != nil {
 			nameToMount[v.Name] = v
 		}
 	}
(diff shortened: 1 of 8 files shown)


cmd/titus-storage/lustre.go
```diff
@@ -70,21 +71,19 @@ func setupLustreMount(parentCtx context.Context, taskID, cname string, vol *core
...
 		fmt.Sprintf("MOUNT_FLAGS=%d", flags),
-		fmt.Sprintf("MOUNT_OPTIONS=%s", strings.Join(mountOptions, ",")),
+		fmt.Sprintf("MOUNT_OPTIONS=%s", mountOptions),
 	}
 
+	l.Infof("Running lustre mount command %+v %+v", cmd.Env, cmd)
 	stdoutStderr, err := cmd.CombinedOutput()
 	if err != nil {
-		return fmt.Errorf("Lustre Mount failure: %+v: %s", cmd, string(stdoutStderr))
+		return fmt.Errorf("Lustre Mount failure: %+v: %s (%w)", cmd, string(stdoutStderr), err)
```
nickbp, line 86: Any benefit to including the taskID or similar context here?

cmd/titus-storage/lustre.go
```diff
@@ -54,7 +54,8 @@ func lustreStart(ctx context.Context, config MountConfig) error {
 }
 
 func setupLustreMount(parentCtx context.Context, taskID, cname string, vol *corev1.Volume, vm *corev1.VolumeMount) error {
-	baseMountOptions := []string{"noatime,flock"}
+	l := logger.GetLogger(parentCtx)
+	mountOptions := "recovery_time_soft=300,recovery_time_hard=900"
```
nickbp, line 58: It looks like this is constant now, could just pass this directly as `MOUNT_OPTIONS=recov[...]` at L80?

mount/titus-mount-lustre.c
```diff
@@ -0,0 +1,359 @@
...
+
+/*
+ *Construct lustre device mount options in the format expected by the syscall
+ */
+static void add_device_options(const char *ipaddress, const char *mount_name, char *final_options)
+{
+	strcat(final_options, ",device=");
+	strcat(final_options, ipaddress);
+	strcat(final_options, "@tcp:/");
+    strcat(final_options, mount_name);
```
nickbp, line 41: m-m-m-meganitpick: indentation (here and at L302)

mount/titus-mount-lustre.c
```diff
@@ -0,0 +1,359 @@
...
+	const char *pidDir = getenv("TITUS_PID1_DIR");
+	const char *lustre_mount_hostname = getenv("MOUNT_LUSTRE_HOSTNAME");
+    const char *lustre_mount_mountname = getenv("MOUNT_LUSTRE_MOUNTNAME");
+	const char *target = getenv("MOUNT_TARGET");
+	const char *flags = getenv("MOUNT_FLAGS");
+	const char *options = getenv("MOUNT_OPTIONS");
+
+	int buf_size = sysconf(_SC_PAGESIZE);
+	char final_options[buf_size];
+	strcpy(final_options, options);
```
nickbp, line 309: `strncpy(final_options, options, buf_size)`?

Could also check length of `options` to see if it would exceed `_SC_PAGESIZE`. For context it looks like `_SC_PAGESIZE` would typically be 4096 so that'd be weird but not impossible

mount/titus-mount-lustre.c
```diff
@@ -0,0 +1,359 @@
...
+
+	if (!(target && flags && options && pidDir)) {
```
nickbp, line 311: Should put this check above the `options` access at L309

mount/titus-mount-lustre.c
```diff
@@ -0,0 +1,359 @@
...
+
+	if (!(lustre_mount_hostname && lustre_mount_mountname)) {
+		fprintf(stderr,
+			"Usage: must provide MOUNT_LUSTRE_HOSTNAME, MOUNT_LUSTRE_MOUNTNAME env vars");
+		return 1;
+	}
+
+	errno = 0;
+	flags_ul = strtoul(flags, NULL, 10);
+	if (errno) {
```
tych0, line 325: The man page says "The  implementation  may also set errno to EINVAL in case no conversion was performed (no digits seen, and 0 returned)." so maybe we need a check for flags_ul != 0 here too?

mount/titus-mount-lustre.c
```diff
@@ -0,0 +1,359 @@
...
+	E_fsconfig(fsfd, FSCONFIG_CMD_CREATE, NULL, NULL, 0);
+}
+
+static int setup_fsfd_in_namespaces(int sk, int nsfd)
+{
+	int fsfd, ret;
+	int usernsfd = openat(nsfd, "user", O_RDONLY);
+	int mnt_fd = openat(nsfd, "mnt", O_RDONLY);
+	int net_fd = openat(nsfd, "net", O_RDONLY);
+	assert(usernsfd != -1);
```
tych0, line 198: Here and elsewhere it would probably be useful to know why usernsfd failed to open (i.e. what the value of errno is). `assert()` for error checking scares me a bit, since it can get inadvertently compiled out; can you change this to a regular perror() like you have later in this function?

tych0:
> Another thing I was unable to solve was how to mount this with MS_NOATIME. No matter what I did I got EINVAL when trying to mount with that flag.

Too many time specifier flags maybe? https://github.com/Netflix-Skunkworks/linux/blob/793d8378b74ac283a4dd7cef1b304553c8a42260/fs/namespace.c#L3613-L3624

It looks like the lustre module code may also parse some mntopts and add time flags on its own. If you have a mount you want to try and do I can try to figure out why it's failing.

mount/titus-mount-lustre.c
solarkennedy: That is weird. `0` is very normal (no extra flags).
But if literally *no* digits were seen, then yea that would be an err and should be reported already on that line

mount/titus-mount-lustre.c
solarkennedy: Haha. We have a `make fmt`, let me run that...

cmd/titus-storage/lustre.go
solarkennedy: It shouldn't, because this *should* get bubbled up to the titus-storage output, which is logging using the taskid already to the journal, which means this error will (is) visible one click away in the titus UI if you click on "titus-storage"

mount/titus-mount-lustre.c
solarkennedy: good idea. Let me do that on a second pass to all the helpers.

solarkennedy:
> > Another thing I was unable to solve was how to mount this with MS_NOATIME. No matter what I did I got EINVAL when trying to mount with that flag.
> 
> Too many time specifier flags maybe? https://github.com/Netflix-Skunkworks/linux/blob/793d8378b74ac283a4dd7cef1b304553c8a42260/fs/namespace.c#L3613-L3624
> 
> It looks like the lustre module code may also parse some mntopts and add time flags on its own. If you have a mount you want to try and do I can try to figure out why it's failing.

The lustre module code can't really be parsing mount options. The `mount` userspace command isn't coming into play at all.
The FLAGS we are passing are `0` here.
However, I can confirm that when one uses the official mount.lustre command with `noatime`, it really does pass the `MS_NOTIME` flag.

But yea I can repro with you watching, I just don't know how to debug further.

mount/titus-mount-lustre.c
tych0: Yeah, I guess it's really strlen(flags) == 0 too; the implementation is free to *not* report an error in the zero-length case, or it might. But actually 0 seems like a fine default for no mount flags specified, so I don't think there's any change necessary here.

mount/titus-mount-lustre.c
solarkennedy: We are controlling all the inputs here, we don't have to be too defensive

tych0:
> But yea I can repro with you watching, I just don't know how to debug further.

I don't know of a great way, in the past I have used:
https://github.com/tych0/kernel-utils/blob/master/why-did-mount-fail.bt

solarkennedy:
thanks to @tych0 for helping me understand the difference between MOUNT_ATTR flags and MS_FLAGs.
I was using MS when I should have used ATTR, which explains why my `noatime` didn't work.
This is now fixed and now everyone will get noatime by default.

mount/titus-mount-lustre.c
tych0: seems like these are still here?

tych0 (approved):
modulo the still present assert()s, lgtm, thanks!

mount/titus-mount-lustre.c
solarkennedy: Yea yea, let me do that on the second pass and I'll cover all the mount helpers, not this one.

PR #1008, opened 2022-12-14
Added panic/recover for main path

solarkennedy:
I appreciate the last minute `state.json` curl, but I think ideally we could catch crashes a little earliar and handle some of the cleanup.

This PR makes it so we are a tiny bit more graceful and let the user know what the panic was and to report it to #titus.

### DIFF
--- a/executor/runner/runner.go
+++ b/executor/runner/runner.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"runtime/debug"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


executor/runner/runner.go
```diff
@@ -163,6 +163,12 @@ func (r *Runner) prepareContainer(ctx context.Context) update {
 // This is just splitting the "run" part of the of the runner
 func (r *Runner) runMainContainerAndStartSidecars(ctx context.Context, startTime time.Time, updateChan chan update) {
 	defer close(updateChan)
+	defer func() {
+		if r := recover(); r != nil {
+			msg := fmt.Sprintf("titus-executor panicked, please report this bug to #titus: %v", r)
+			updateChan <- update{status: titusdriver.Failed, msg: msg, details: nil}
```
tych0, line 169: Does this print the stack trace? I think we should at least log it, if not render it here.

tych0 (approved):

executor/runner/runner.go
solarkennedy: The actual string that is return to the user is whatever is inside the `panic("bla")`. The trace goes stdout (kibana).
I don't think we should render it here (this update is supposed to be 1 line, gets into the pod + cassandra)

tych0: Yeah, as long as it goes to kibana that's fine with me.

tych0: Though, I don't understand who would log it to stdout:
```
~ cat tester.go
package main

import (
	"fmt"
)

func foo() {
	panic("foo")
}

func main() {
	defer func() {
		r := recover()
		fmt.Printf("recover: %v\n", r)
	}()
	foo()
}```

solarkennedy: Hmmm. I saw the panics in kibana!
But uh, now you have me second guessing.

solarkennedy: Well, added explicit logging, which is nice because it gets structured tags and stuff.

PR #1009, opened 2022-12-16
Added debug for full command printing when we get exit code 127

solarkennedy:
It is pretty hard to debug exit code 127.

This PR adds some debug for the user to print *Exactly* what was (attempted to) run in the container.

This is a little weird though, because the command could still be a shell script which *also* can return 127.

I'm not super sure how to capture that nuance.

Example error on the titus ui:

> main container couldn't find the command 'bad_entrypoint' to run in the path (exit code 127). Full command: ["bad_entrypoint"]

### DIFF
--- a/executor/runtime/docker/docker.go
+++ b/executor/runtime/docker/docker.go
@@ -58,6 +58,7 @@ import (
 var (
 	_                   runtimeTypes.Runtime = (*DockerRuntime)(nil)
 	errMissingResources                      = errors.New("Missing container resources")
+	fullCommandForDebug                      = map[string][]string{}
 )
 
 // units
@@ -378,6 +379,7 @@ func (r *DockerRuntime) mainContainerDockerConfig(c runtimeTypes.Container, bind
 		return nil, nil, err
 	}
 
+	r.SetFullCommandForDebug(runtimeTypes.MainContainerName, append(entrypoint, cmd...))
 	containerCfg := &container.Config{
 		Image:      c.QualifiedImageName(),
 		Entrypoint: entrypoint,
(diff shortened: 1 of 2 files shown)


jamesmulcahy:
Thanks @solarkennedy

executor/runtime/docker/docker.go
```diff
@@ -2279,7 +2282,19 @@ func (r *DockerRuntime) handleDockerEvent(message events.Message, statusMessageC
...
+				Msg:    cName + " container successfully finished with exit code 0",
+			}
+			return isTerminalDockerEvent
+		} else if exitCode == "127" {
+			// An exit code of 127 means command not found.
+			// This isn't a very useful exit code for users, because it doesn't say exactly "What" the
+			// command was that was actually at play!
+			// We can help the user out a little bit by simply letting them know the command was.
+			firstCommand, fullCommand := r.GetFullCommandForDebug(cName)
+			msg := fmt.Sprintf("%s container couldn't find the command '%s' to run in the path (exit code %s) or a shell script returned the same thing. Full command: %s", cName, firstCommand, exitCode, fullCommand)
```
nickbp, line 2294: For the nested error case, it looks like there'd be something like this in stderr. Could also mention something about checking stderr:
```
$ ./test.sh
./test.sh: line 3: missingcommand: command not found
```

Maybe something like this? I dunno...:
> [foo] container had a command not found error (exit code 127), either from the container command '[bar]' or from a shell script executed by that command. stderr may have more detail. Full command: [bar baz]

nickbp (approved):

executor/runtime/docker/docker.go
tych0: If we wanted to go really overboard, we have the filesystem, we could start another container and inspect it. If the original entrypoint is there and executable, we can render the error about a shell script, if it's not present, then we can just say something about "this entrypoint doesn't exist in this image". IMO this as-is is fine for now.

solarkennedy: Yea, no need to go too crazy on this first pass.
However I would like to enhance the error messages on failed tasks with the last line of stdout/err, but in a different PR (more involved)

PR #1010, opened 2022-12-16
untested: add tsa version to versions string

tych0:
tsa needs a --version argument first.

### DIFF
--- a/executor/runtime/types/types.go
+++ b/executor/runtime/types/types.go
@@ -4,7 +4,9 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"os/exec"
 	"regexp"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


executor/runtime/types/types.go
```diff
@@ -514,8 +516,48 @@ func getKernelVersion() string {
...
+	}
+
+	outInt, err := strconv.Atoi(strings.TrimSpace(string(out)))
+	if err != nil {
+		return fmt.Sprintf("bad TSA pid: %s", string(out))
+	}
+
+	// the service wasn't running. possible if the user didn't request TSA in the first place.
+	if outInt == 0 {
+		return ""
```
tych0, line 540: This means that this gets re-evaluated every time there's a version query and the service wasn't present. We should probably switch to `sync.Once{}` if we decide to actually go this route.

PR #1011, opened 2022-12-16
HN-30: render TSA version in version string

tych0:


solarkennedy (approved):
Way safer. No way to time out. No additional load on systemd. lgtm!

nickbp (approved):

PR #1014, opened 2023-01-06
Added preliminary support for emptydir volume types

solarkennedy:
This adds support for emptydir (ramdisk) volume types.

Key features:

* Ram disk is real, and does show up in the mount (and host) namespace of the containers
* Supports 1->N container mappings like k8s
* Ram is accounted for in the container (will oom if you put too much!)
* Gets cleaned up after shutdown.

### DIFF
--- a/cmd/titus-storage/emptydir.go
+++ b/cmd/titus-storage/emptydir.go
@@ -1,0 +1,112 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"os/exec"
+	"path"
+	"strings"
+
+	executorDocker "github.com/Netflix/titus-executor/executor/runtime/docker"
+	"github.com/Netflix/titus-executor/logger"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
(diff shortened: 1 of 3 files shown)


tych0 (approved):

cmd/titus-storage/emptydir.go
```diff
@@ -0,0 +1,112 @@
...
+		for i := range container.VolumeMounts {
+			vm := container.VolumeMounts[i]
+			vol, ok := nameToVolume[vm.Name]
+			if !ok {
+				continue
+			}
+
+			err := setupEmptyDirMount(ctx, config.taskID, container.Name, &vol, &vm)
+			if err != nil {
+				return err
```
nickbp, line 47: Don't think it's really an issue in practice but I wonder if the task runtime dir is cleaned up automatically if task init is canceled due to an error here

nickbp (approved):

cmd/titus-storage/emptydir.go
solarkennedy: Yes, it is. That is covered by a failsafe via the systemd unit

PR #1015, opened 2023-01-06
Component version refactor (With Titus System Services)

solarkennedy:
This adds support for adding in Titus System Services into the componentVersions view in the UI.

A couple of thoughts when making this:
* I thought prefixing `tss-` would be a good idea to make it more clear these are system services, and not deb packages on an agent or container
* I tried to do the best I could to get the semantically meaningful version string, without being too verbose
* But also I didn't want to make up a fake docker url that doesn't exist (too far?)
* I butchered the image urls. I couldn't find the functions I wanted in the docker `reference` package

Screenshot from 2023-01-06 11-23-09

### DIFF
--- a/executor/runtime/docker/docker_image.go
+++ b/executor/runtime/docker/docker_image.go
@@ -1,0 +1,77 @@
+package docker
+
+import (
+	"strings"
+
+	"github.com/docker/distribution/reference"
+	"github.com/docker/docker/api/types"
+)
+
+// cleanContainerVersion tries to provide the most human-friendly version string we can get for
+// a particular image, given a version inspect.
+// It goes from most human-friendly to least (sha)
+func cleanContainerVersion(i *types.ImageInspect) string {
+	if i == nil {
+		return ""
+	}
+	tag := getImageTagFromInspect(i)
+	if tag != "" {
+		return tag
+	}
+	buildLabel := getBuildLabelFromInspect(i)
+	if buildLabel != "" {
+		return buildLabel
+	}
+	return getImageShaFromInspect(i)
+}
+
+func getImageTagFromInspect(i *types.ImageInspect) string {
+	if i.RepoTags == nil {
+		return ""
+	}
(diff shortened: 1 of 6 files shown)


executor/runtime/docker/docker_image.go
```diff
@@ -0,0 +1,70 @@
...
+	return ""
+}
+
+func getImageShaFromInspect(i *types.ImageInspect) string {
+	if len(i.RepoDigests) < 1 {
+		return ""
+	}
+	name := fullImageNameToShortName(i.RepoDigests[0])
+	splitDigest := strings.Split(i.RepoDigests[0], "@")
+	if len(splitDigest) == 0 {
```
aaronlehmann, line 55: `strings.Split` won't return a zero-length slice:

> If s does not contain sep and sep is not empty, Split returns a slice of length 1 whose only element is s.

aaronlehmann: I think you want to check `if len(splitDigest) < 2` instead to make the use of `splitDigest[1]` below safe.

executor/runtime/docker/docker_image.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+func fullImageNameToShortName(i string) string {
+	ref, err := reference.ParseNamed(i)
+	if err != nil {
+		return i
+	}
+	fullURL := ref.Name()
+	splitURL := strings.Split(fullURL, "/")
+	return strings.Join(splitURL[1:], "/")
```
aaronlehmann, line 69: Not sure this is safe unless there is some guarantee that image names in these contexts will always include a registry (i.e. `busybox` is a vaild reference).

aaronlehmann: This can be simplified to:

```go
return reference.Path(ref)
```

solarkennedy:
ty @aaronlehmann for the docker reference tips.
Tests + builds are passing now

aaronlehmann (approved):

